PR #21, opened 2022-05-17
Print test summaries and follow-up commands

fmeum:


tools/replayer/src/replayer.c
```diff
@@ -265,10 +320,102 @@ void run_file_or_dir(const char *path) {
...
+      fprintf(stderr, "To debug this failure, execute:\n\n"
+                      "    gdb -ex 'break LLVMFuzzerTestOneInput' -ex run --args '%s'\n", argv0);
+#endif
+    }
+  }
+}
+
+static void explicit_exit_handler(void) {
+  failure_reason = "Fuzz target exited";
+  print_summary();
```
adombeck, line 359: Since we always set `failure_reason` right before calling `print_summary`, why not make it an argument instead of a global variable?

adombeck (approved):

PR #23, opened 2022-05-20
Fix tests not exiting with exit code returned by m.Run()

adombeck:


jochil (approved):

fmeum (changes requested):
Please don't merge this yet.

adombeck:
> Edit with the rationale: Could you explain why you want to make this change? An explicit os.Exit should no longer be required in TestMains since https://github.com/golang/go/issues/34129, which made it into Go 1.16. Given that the pattern doesn't interact nicely with defer and makes it more difficult to build tooling around tests (I got bitten by this in https://github.com/bazelbuild/rules_go/pull/3117#issuecomment-1105201956), I specifically left out the os.Exit. I would be interested in hearing why you would want to introduce it - are any tests failing open without it?

Ah, I wasn't aware of https://github.com/golang/go/issues/34129. I made this change because I thought that's how you're supposed to use `M.Run()`, as suggested by [the documentation](https://pkg.go.dev/testing#M.Run): 
> Run runs the tests. It returns an exit code to pass to os.Exit. 

... and its usages all over the Go SDK.

fmeum:
> Ah, I wasn't aware of [golang/go#34129](https://github.com/golang/go/issues/34129). I made this change because I thought that's how you're supposed to use `M.Run()`, as suggested by [the documentation](https://pkg.go.dev/testing#M.Run):
> 
> > Run runs the tests. It returns an exit code to pass to os.Exit.
> 
> ... and its usages all over the Go SDK.

I only learned about this really useful change when the `os.Exit` pattern broke me and I started reading `testing.go`. Documentation should definitely be improved.

PR #28, opened 2022-05-24
cmake: Set compiler and linker options

fmeum:
Add the same compiler and linker options as are set for make
invocations when using clang or gcc and adapt the set of flags for
MSVC-style compilers.

### DIFF
--- a/tools/cmake/CIFuzz/share/CIFuzz/CIFuzzFunctions.cmake
+++ b/tools/cmake/CIFuzz/share/CIFuzz/CIFuzzFunctions.cmake
@@ -1,4 +1,31 @@
+# Note: Keep the clang flags used below in sync with internal/cmd/run/run.go
+#       Explanations of these flags are provided in that file.
 function(enable_fuzz_testing)
+  if(CIFUZZ_TESTING)
+    add_compile_definitions(FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION)
+    if(MSVC)
+      add_compile_options(
+          # Allow the compiler to inline more aggressively. This overrides the (questionable?) default of /Ob1 set by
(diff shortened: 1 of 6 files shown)


tools/cmake/CIFuzz/share/CIFuzz/CIFuzzFunctions.cmake
```diff
@@ -1,4 +1,28 @@
+# Note: Keep the clang flags used below in sync with internal/cmd/run/run.go
+#       Explanations of these flags are provided in that file.
 function(enable_fuzz_testing)
+  # Implicitly enable general compiler flags if any CIFUZZ_* cache variable is set to a non-default value.
+  if (CIFUZZ_TESTING OR CIFUZZ_SANITIZERS OR (NOT CIFUZZ_ENGINE STREQUAL replayer))
```
adombeck, line 5: Why do we not enable the general compiler flags when `CIFUZZ_ENGINE` is `replayer`?

tools/cmake/CIFuzz/share/CIFuzz/CIFuzzFunctions.cmake
```diff
@@ -1,4 +1,28 @@
...
+          /Ox
+          # MSVC's equivalent of -fno-omit-frame-pointer.
+          /Oy-
+      )
+    else()
+      add_compile_options(
+          -g
+          -Og
+          -fno-omit-frame-pointer
+      )
```
adombeck, line 23: What about `-DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION`?

adombeck (approved):

fmeum:
I will merge this for now and try out how well it interacts with presets. If build types would be more ergonomic with it, I would switch in a follow-up PR.

PR #30, opened 2022-05-24
Remove glog

adombeck:
Move glogutil back to core repo
    
We decided that we don't want to use glog in cifuzz.


pkg/runner/libfuzzer/libfuzzer_runner.go
```diff
@@ -194,7 +194,8 @@ func (r *Runner) RunLibfuzzerAndReport(ctx context.Context, args []string, env [
 		return err
 	}
 
-	glog.Infof("Running command %s", strings.Join(stringutil.QuotedStrings(cmd.Args), " "))
+	cmd.String()
```
fmeum, line 197: Good bot :-)

adombeck:
> There is a downside of using our dialog package... the Debug functions are using the verbose flag provided by viper.
Depending on when you are using dialog.Debug the flag might not be passed to viper
As long as it is used in the run functions of a command there should be no problem. 

In the corresponding MR of the core repo (https://gitlab.com/code-intelligence/core/-/merge_requests/4908), I'm setting the `verbose` viper value if the glog verbosity level is at least 1.

> But using it inside init will result in output not shown even if the --verbose is set

It's the same with glog if I'm not mistaken.

jochil (approved):

adombeck:
> ok. I just wanted to make sure we are aware of this behavior... as it might be confusing if you add debug output and it does not show up

Yes, it's good to be aware of that. glog actually handles this differently: If any messages are printed via glog before the glog flags were parsed, it prints the message along with a warning that the flags were not parsed. We could do something similar by setting `verbose` to `false` when the verbose flag is not set and make the `dialog` functions print a warning if `viper.Get("verbose") == ""`.

PR #32, opened 2022-05-24
cmake: Emit CMake target information to .cifuzz/fuzz_tests

fmeum:
Already after the CMake configure step, the
$CMAKE_BUILD_DIR/.cifuzz/fuzz_tests directory will contain one file per
target containing the absolute path to the corresponding binary.

This can be used to provide command completion for cifuzz run and to
determine the path of the executable to run.

### DIFF
--- a/tools/cmake/CIFuzz/share/CIFuzz/CIFuzzFunctions.cmake
+++ b/tools/cmake/CIFuzz/share/CIFuzz/CIFuzzFunctions.cmake
@@ -1,6 +1,10 @@
 # Note: Keep the clang flags used below in sync with internal/cmd/run/run.go
 #       Explanations of these flags are provided in that file.
 function(enable_fuzz_testing)
+  # Remove the metadata directory we populate with fuzz test target information at configuration time so that e.g.
+  # metadata for renamed or removed targets doesn't linger around.
+  file(REMOVE_RECURSE "${CMAKE_BINARY_DIR}/$<CONFIG>/.cifuzz")
+
   if(CIFUZZ_TESTING)
     add_compile_definitions(FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION)
     if(MSVC)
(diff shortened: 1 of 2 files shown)


fmeum:
@adombeck This is blocked on #28, but you could already check whether the provided information and its location would fit your needs.

tools/cmake/CIFuzz/share/CIFuzz/CIFuzzFunctions.cmake
```diff
@@ -75,5 +74,15 @@ function(add_fuzz_test name)
   # them and are not easy to deal with using just POSIX functions).
   target_compile_definitions("${name}" PRIVATE CIFUZZ_SEED_CORPUS="${_source_seed_corpus}")
 
+  # Collect a mapping from CMake target names to binary paths for cifuzz.
+  # We can't use add_custom_command here as we want the mapping to exist already after the configure step, not only
+  # after the build step.
+  # Note: Removed and renamed targets leave behind their entry in this mapping. Since these files are cheap to
+  #       regenerate, cifuzz can just delete the entire .cifuzz directory before each build.
+  set(_info_file "${CMAKE_BINARY_DIR}/$<CONFIG>/.cifuzz/fuzz_tests/${name}")
```
adombeck, line 82: What's the `$<CONFIG>` part here?

tools/cmake/CIFuzz/share/CIFuzz/CIFuzzFunctions.cmake
```diff
@@ -75,5 +74,15 @@ function(add_fuzz_test name)
   # them and are not easy to deal with using just POSIX functions).
   target_compile_definitions("${name}" PRIVATE CIFUZZ_SEED_CORPUS="${_source_seed_corpus}")
 
+  # Collect a mapping from CMake target names to binary paths for cifuzz.
+  # We can't use add_custom_command here as we want the mapping to exist already after the configure step, not only
+  # after the build step.
```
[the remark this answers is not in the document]
fmeum: With presets, users would naturally refresh the mapping if they use our preset in the IDE. For that, having the mapping exist already after the configure step would also be an advantage. I will look for other solutions to this first, but `cifuzz reload` sounds like a reasonable backup plan.

tools/cmake/CIFuzz/share/CIFuzz/CIFuzzFunctions.cmake
fmeum: On Windows with MSBuild, every CMake build directory contains multiple configurations (`Debug`, `Release`, ...), whereas with Makefiles and ninja every build directory contains only a single configuration. In the multi-config case, we need to include the `$<CONFIG>` in the path here or CMake produces an error about conflicting files being generated. With #28, this segment will be either `RelWithDebInfo` (MSBuild) or `.` (everything else).

adombeck (approved):

PR #34, opened 2022-05-25
Move logging functions to log package

adombeck:
... and print all log messages to stderr.
    
We only keep using stdout for dialog output (questions etc. which require user input).


internal/cmd/create/create.go
```diff
@@ -59,36 +60,36 @@ func run(cmd *cobra.Command, args []string, opts *cmdOpts) (err error) {
...
+			log.Errorf(err, "Unable to created fuzz test, file already exists %s\n", stubPath)
 			return cmdutils.WrapSilentError(err)
 		}
 	}
 
 	// show success message
-	dialog.Successf("Fuzz test stub created at %s\n", stubPath)
-	dialog.Info(`
+	log.Successf("Fuzz test stub created at %s\n", stubPath)
+	log.Info(`
```
jochil, line 92: I don't see why something like this should go to `stdErr`... it is not diagnostic/debug output. Same for the `log.Success`

internal/cmd/create/create.go
```diff
@@ -59,36 +60,36 @@ func run(cmd *cobra.Command, args []string, opts *cmdOpts) (err error) {
 	if err != nil {
 		return err
 	}
-	dialog.Debugf("Selected fuzz test type: %s\n", opts.testType)
+	log.Debugf("Selected fuzz test type: %s\n", opts.testType)
```
jochil, line 63: `\n` can be removed due to the previous commit

adombeck: Done

adombeck:
@jochil I think you meant to approve this, no?

jochil (approved):

PR #40, opened 2022-05-27
Remove afero

adombeck:
We found that it's more work to use than we benefit from it.


internal/cmd/create/create.go
```diff
@@ -7,23 +7,21 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/pkg/errors"
```
fmeum, line 10: The commit groups are no longer ordered. We don't have to care though - or we could set up go-imports.

pkg/storage/utils.go
```diff
@@ -20,10 +19,7 @@ func GetOutDir(requestedDir string, fs *afero.Afero) (string, error) {
 		return cwd, nil
 	}
 
-	if _, err := fs.Stat(requestedDir); err != nil && !os.IsNotExist(err) {
-		return requestedDir, errors.WithStack(err)
-	}
-	if err := fs.MkdirAll(requestedDir, 0744); err != nil {
+	if err := os.MkdirAll(requestedDir, 0755); err != nil {
```
fmeum, line 22: The permissions were changed from `0744` to `0755` in this commit. Now that I am thinking about it, wouldn't `0700` be more appropriate?

pkg/storage/utils_test.go
```diff
@@ -1,27 +1,46 @@
...
 	"os"
+	"path/filepath"
 	"testing"
 
-	"code-intelligence.com/cifuzz/pkg/workarounds"
 	"github.com/pkg/errors"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"code-intelligence.com/cifuzz/util/fileutil"
```
fmeum, line 14: See above, import group ordering.

pkg/stubs/stubs.go
```diff
@@ -53,16 +58,13 @@ func SuggestFilename(dir string, testType config.FuzzTestType, fs *afero.Afero)
 	counter := 1
 	for {
```
fmeum, line 59: Could this be `for counter := 1;; counter++ {`?

fmeum (approved):

internal/cmd/create/create.go
adombeck: The imports are ordered by the format.py script from our core repo (which calls goimports-reviser and goimports)

pkg/storage/utils.go
adombeck: I changed them because `0744` are incorrect permissions for directories (directories need the executable bit to be traversable). `0755` are the permissions directories are usually created with, except if the directory contains secrets. Here we're creating a directory which should contain fuzz targets, which don't qualify as secret IMO. Note that umask is applied here (so if the system umask is set to `077`, the directory will be created with permissions `0700`) and that the directory can only be accessed by group and others if the parent directories are also accessible by group and others.

pkg/stubs/stubs.go
adombeck: Yes, done

pkg/storage/utils_test.go
adombeck: Same as above, that's how goimports and goimports-reviser group the imports

PR #44, opened 2022-05-31
runner: Colorize sanitizer options if running in a tty

fmeum:
...but also filter out the color codes before parsing and storing the
logs.


fmeum:
@adombeck I may have implemented this prematurely depending on how your work will end up rendering findings, but at least the part that enables colored sanitizer reports should remain relevant.

pkg/runner/runnerutils.go
```diff
@@ -48,6 +52,15 @@ const LibFuzzerOOMExitCode = 71
 //
 const LibFuzzerTimeoutExitCode = 70
 
+var defaultSanitizerOptions = map[string]string{
+	// color defaults to "auto", which lets sanitizers detects whether colorized
+	// output is supported by determining whether their stdout is a tty. Since
+	// we run the sanitizers in a subprocess, this will never detect a tty.
```
adombeck, line 58: I don't think the reason is that we run it in a subprocess. As long as the subprocesses' stdout and stderr are connected to the TTY's stdout and stderr, the sanitizers should still be able to detect whether it's a TTY. Maybe the reason this doesn't work in cifuzz is that we're connecting the subprocesses' stderr to an `StderrTeePipe`. Note that we connect stdout to `os.Stderr`, i.e. the stderr of the TTY.

pkg/runner/runnerutils.go
```diff
@@ -177,3 +192,13 @@ func SetLDLibraryPath(env []string, libraryDirs []string) ([]string, error) {
 	libDirsList = envutil.AddToColonSeparatedList(libDirsList, libraryDirs...)
 	return envutil.Setenv(env, "LD_LIBRARY_PATH", libDirsList)
 }
+
+func sanitizerOptionsColorValue() string {
+	// Colorize sanitizer reports if cifuzz itself is running in an interactive
+	// terminal.
+	if terminal.IsTerminal(int(os.Stdout.Fd())) {
```
adombeck, line 199: We are printing fuzzer output to stderr, so we should check here if `os.Stderr` is connected to a terminal.

adombeck (approved):

pkg/parser/libfuzzer/libfuzzer_parser.go
```diff
@@ -110,7 +111,11 @@ func (p *parser) sendReport(ctx context.Context, report *report.Report) error {
 	}
 }
 
-func (p *parser) parseLine(ctx context.Context, line string) error {
+func (p *parser) parseLine(ctx context.Context, possiblyColorizedLine string) error {
+	// Sanitizer reports can be colorized, but the ANSI escapes used for colors
+	// should not be included in logs.
+	line := executil.StripANSIColors(possiblyColorizedLine)
```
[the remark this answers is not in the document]
fmeum: I'm using pterm now.

pkg/runner/runnerutils.go
fmeum: Good point, fixed the comment to be actually correct.

pkg/runner/runnerutils.go
fmeum: Fixed and commented.

PR #46, opened 2022-05-31
Print basic fuzzing metrics

adombeck:
This doesn't pretty-print findings yet, those are still printed as JSON.


internal/cmd/run/metrics/metrics.go
```diff
@@ -0,0 +1,91 @@
...
+	NumSeedsAtInit int
+
+	averageExecs int32
+	numMetrics   int32
+}
+
+func (p *Printer) PrintLiveMetrics(metrics *report.FuzzingMetric) {
+	p.updateAverageExecutionsPerSecond(metrics)
+
+	lastNewCoverageDuration := time.Second * time.Duration(metrics.SecondsSinceLastFeature)
```
fmeum, line 46: I'm not a fan of calling the features count "coverage" since that usually means something else. We could brainstorm other terms. I start to like a simple "progress" in this place.

internal/cmd/run/metrics/metrics.go
```diff
@@ -0,0 +1,91 @@
...
+		numberString("%s", lastNewCoverageDuration.String()),
+	)
+	p.UpdateText(s)
+}
+
+func (p *Printer) updateAverageExecutionsPerSecond(metrics *report.FuzzingMetric) {
+	if p.numMetrics == 0 {
+		p.averageExecs = metrics.ExecutionsPerSecond
+	}
+	p.averageExecs = (p.numMetrics*p.averageExecs + metrics.ExecutionsPerSecond) / (p.numMetrics + 1)
```
fmeum, line 63: `p.numMetrics * p.averageExecs` could potentially overflow on long & fast fuzzing runs. You could probably use [this trick](https://nullbuffer.com/articles/welford_algorithm.html) to prevent that.

internal/cmd/run/run.go
```diff
@@ -96,13 +106,17 @@ func New() *cobra.Command {
...
 	cmd.Flags().BoolVar(&opts.useSandbox, "sandbox", useMinijailDefault, "By default, fuzz tests are executed in a sandbox to prevent accidental damage to the system.\nUse --sandbox=false to run the fuzz test unsandboxed.\nOnly supported on Linux.")
+	cmd.Flags().BoolVar(&opts.printJSON, "json", false, "Print output as JSON")
 	cmdutils.MarkFlagsRequired(cmd, "seeds-dir", "fuzz-test")
 
 	return cmd
 }
 
 func (c *runCmd) run() error {
 	var err error
+	// TODO: Should we include the build time in the "Ran for X" time
```
fmeum, line 117: I would say only the fuzzing time. Printing and indicating build times would be the job of the build system, whose output we still print (right?)

internal/cmd/run/run.go
```diff
@@ -379,13 +411,89 @@ func setBuildFlagsEnvVars(env []string) ([]string, error) {
...
+
+func countSeeds(seedDirs []string) (numSeeds int, err error) {
+	for _, dir := range seedDirs {
+		var seedsInDir int
+		err = filepath.WalkDir(dir, func(path string, d os.DirEntry, err error) error {
+			if err != nil {
+				return err
+			}
+			if d.IsDir() {
+				return nil
```
fmeum, line 481: Is the libfuzzer behavior to not recurse into directories? I wasn't aware of this and implemented recursive traversals in the replayer.

internal/cmd/run/metrics/metrics.go
fmeum: That said, I think we want to make this a float/double even if we don't print it as such.

internal/cmd/run/metrics/metrics.go
adombeck: I called it that because that's the name of the report field in our core repo (see the JSON name of the field: https://github.com/CodeIntelligenceTesting/cifuzz/blob/CLI-171-print-metrics/pkg/report/report.go#L49-L49).

I'm also unhappy with this naming. When the user runs cifuzz with `--verbose`, the libFuzzer output is printed as well, which continuously prints its stats including `cov:` and `ft:`, and right below those lines we print our own metrics where `coverage:` doesn't match the `cov:` from libfuzzer but the `ft:`, which is obviously strange.

"progress" is better but to me it sounds a bit as if there was a certain value which must be reached to finish the fuzzing.

internal/cmd/run/metrics/metrics.go
adombeck: Good catch! I'm now calculating the average similar to the algorithm you linked and am using a float64 for the calculation. Please take a look.

internal/cmd/run/run.go
adombeck: > Printing and indicating build times would be the job of the build system, whose output we still print (right?)

Agreed, and yes, we still print the output of the build system

adombeck: Done

internal/cmd/run/run.go
adombeck: `filepath.WalkDir` does recurse into directories (same as libFuzzer)

internal/cmd/run/metrics/metrics.go
adombeck: Updated as per our discussion

internal/cmd/run/metrics/metrics.go
```diff
@@ -0,0 +1,88 @@
...
+		numberString("%s", lastNewFeatureDuration.String()),
+		delimString(" - "),
+		descString("exec/s: "),
+		numberString("%d", metrics.ExecutionsPerSecond),
+	)
+	p.UpdateText(s)
+}
+
+func (p *Printer) updateAverageExecutionsPerSecond(metrics *report.FuzzingMetric) {
+	p.averageExecs += (float64(metrics.ExecutionsPerSecond) - p.averageExecs) / float64(p.numMetrics+1)
```
fmeum, line 60: What I missed on the first pass: This method of calculating the overall average executions/s so far seems flawed as the duration and number of executions between two metrics reports varies. As a result, a simple average doesn't assign the correct weights to the individual data points and, more concretely, would probably always overestimate the true avg exec/s as there tend to be both more executions/s and more metric reports at the beginning of a fuzzing run.

How about we instead track the total number of executions in an uint64 and then compute the average exec/s by dividing by `duration` below?

fmeum: Thinking about this again: Couldn't we just record the time of the first and the last metric report and use that as the denominator here rather than the total duration of the fuzzing run? That would give us the correct overall average exec/s as of the last update, which is the best we could possibly do here.

adombeck: Done

adombeck:
@fmeum: I pushed the changes I made since your last review as fixup commits to hopefully ease the next round of review. I'll squash those before merging.

fmeum (approved):

adombeck:
@fmeum: Sorry, have to ask for another review. I pushed 5 more commits which improve formatting and finally (or so I hope) fix the spinner printer when used in verbose mode.

internal/cmd/run/report_handler/report_handler.go
```diff
@@ -37,27 +37,34 @@ type ReportHandler struct {
...
 func NewReportHandler(printJSON, verbose bool) (*ReportHandler, error) {
 	var err error
-	var printer metrics.Printer
+	h := &ReportHandler{
+		printJSON: printJSON,
+		verbose:   verbose,
+		startedAt: time.Now(),
+	}
+
+	// When --json was used, we don't want anything but JSON output on
```
fmeum, line 48: Is there any downside to *always* printing to stderr? If I remember correctly, we decided that "structured output to stdout" was indeed the only hard rule we wanted to apply.

pkg/log/log.go
```diff
@@ -10,12 +10,14 @@ import (
...
 func init() {
-	// Disable color if stderr is not a terminal.
-	if !term.IsTerminal(int(os.Stderr.Fd())) {
-		pterm.DisableColor()
-	}
+	// Disable color if stderr is not a terminal. We don't use
+	// color.Disable() here because that would disable color for all
+	// pterm and color methods, but we might want to use color in output
+	// printed to stdout (if stdout is a terminal).
+	disableColor = !term.IsTerminal(int(os.Stderr.Fd()))
```
fmeum, line 21: Should the FD used here depend on what we provided to `color.SetOutput`?

internal/cmd/run/report_handler/report_handler.go
```diff
@@ -68,16 +69,28 @@ func NewReportHandler(printJSON, verbose bool) (*ReportHandler, error) {
...
 
 	// Print report as JSON if the --json flag was specified
 	if h.printJSON {
-		jsonString, err := stringutil.ToJsonString(r)
-		if err != nil {
-			return err
+		var jsonString string
+		// Print with color if stdout is a TTY
+		if term.IsTerminal(int(os.Stdout.Fd())) {
+			bytes, err := prettyjson.Marshal(r)
```
fmeum, line 85: Looked great in the demo!

pkg/parser/libfuzzer/libfuzzer_parser.go
```diff
@@ -407,13 +407,19 @@ func (p *parser) parseAsFuzzingMetric(log string) *report.FuzzingMetric {
 			return nil
 		}
 		now := time.Now()
-		secondsSinceLastFeature := now.Unix() - p.lastNewFeatureTime.Unix()
+		var secondsSinceLastFeature uint64
+		if p.lastNewFeatureTime.IsZero() {
+			secondsSinceLastFeature = 0
```
fmeum, line 412: Why is this branch needed?

internal/cmd/run/metrics/updating_printer.go
```diff
@@ -68,7 +68,7 @@ func (p *UpdatingPrinter) printMetrics(metrics *report.FuzzingMetric) {
 	s := fmt.Sprint(
 		metricsToString(metrics),
 		DelimString(" ("),
-		pterm.LightYellow(time.Since(p.startedAt).Round(time.Second).String()),
+		pterm.LightYellow(time.Since(p.startedAt).Truncate(time.Second).String()),
```
fmeum, line 74: I noticed this during the demo: Does this ensure that "last new path" will now always be <= "execution time"?

util/ptermutil/ptermutil.go
```diff
@@ -1,31 +1,52 @@
...
+	// To ensure that after the write, the spinner printer has a new
+	// line for itself, we only write the output if it ends with a
+	// newline. Else, we store it in a buffer which we write the next
+	// time Write() is called with something that ends with a newline.
+	if len(p) == 0 || p[len(p)-1] != '\n' {
+		w.buf = append(w.buf, p...)
+		return len(p), nil
+	}
+	oldBuf := w.buf
+	buf := append(oldBuf, p...)
```
fmeum, line 39: I find it slightly confusing to have three different `buf`-like variables here. Couldn't we keep `len(w.buf)` in `oldBufLen`, set `w.buf = append(w.buf, p...)` and then `w.buf = []byte{}` after the `fmt.Fprint` call? That would fit my mental model of "clearing the buffer after the write" better.

util/ptermutil/ptermutil.go
```diff
@@ -1,31 +1,52 @@
...
-		// Return the first new error. Note that this is not thread-safe,
-		// the error might actually be from other concurrent calls to
-		// functions of the color package, but it's the best we can do.
-		return 0, newErrs[len(oldErrs)]
+func (w *ptermWriter) Write(p []byte) (n int, err error) {
+	// To avoid races, only one write is executed at a time
+	writeLock.Lock()
+	defer writeLock.Unlock()
+
+	// To ensure that after the write, the spinner printer has a new
```
fmeum, line 30: Great fix, thanks for digging this deep!

fmeum (approved):

pkg/parser/libfuzzer/libfuzzer_parser.go
```diff
@@ -34,22 +41,31 @@ var (
 	javaAssertionErrorPattern = regexp.MustCompile(
 		`== Java Assertion Error`)
 
-	coveragePattern = regexp.MustCompile(
-		`#(?P<total_execs>\d+).*cov:\s(?P<edges>\d+)\sft:\s(?P<features>\d+)\scorp:\s(?P<corpus_size>\d+)/.*exec/s:\s(?P<executions_per_second>\d+)\s`)
+	statsPattern = regexp.MustCompile(
+		`#(?P<total_execs>\d+)\s(?P<status>.*)\scov:\s(?P<edges>\d+)\sft:\s(?P<features>\d+)\scorp:\s(?P<corpus_size>\d+)/.*exec/s:\s(?P<executions_per_second>\d+)\s`)
```
jochil, line 44: Only a personal preference: For more complicated regex I like to have examples of targeted strings as a comment.
Helps to understand the regex way faster

jochil (changes requested):
Basically all new files are without any unit test (`line_printer.go`, `metrics.go`, `updating_printer.go`, `report_handler.go`, `ptermutil.go`

pkg/log/log.go
adombeck: No, the log package doesn't depend on `color.SetOutput` in the current implementation, it always prints to stderr, see https://github.com/CodeIntelligenceTesting/cifuzz/blob/CLI-171-print-metrics/pkg/log/log.go#L48-L48.

internal/cmd/run/report_handler/report_handler.go
adombeck: Thanks :)

pkg/parser/libfuzzer/libfuzzer_parser.go
adombeck: IIUC, when `p.lastNewFeatureTime` was not set yet, `now.Sub(p.lastNewFeatureTime)` equals `now`, so we would set `secondsSinceLastFeature` to the number of seconds since epoch. But the previous code, `now.Unix() - p.lastNewFeatureTime.Unix()`, should have the same issue and I'm a bit confused why we didn't notice it before.

internal/cmd/run/metrics/updating_printer.go
adombeck: That was the goal of this change and in my tests it worked

util/ptermutil/ptermutil.go
adombeck: Ah yes, forgot to clean up that mess. Thanks, I implemented your suggestion

internal/cmd/run/report_handler/report_handler.go
fmeum: Preserving that property makes sense. We should keep the added complexity in mind if we plan future changes in this area though.

pkg/parser/libfuzzer/libfuzzer_parser.go
adombeck: I changed this to
```go
var secondsSinceLastFeature uint64
if !p.lastNewFeatureTime.IsZero() {
    secondsSinceLastFeature = uint64(now.Sub(p.lastNewFeatureTime).Truncate(time.Second).Seconds())
}
```
and am now doing the same below for `secondsSinceLastEdge`.

adombeck:
> Basically all new files are without any unit test (line_printer.go, metrics.go, updating_printer.go, report_handler.go, ptermutil.go)
Also internal/cmd/run/run.go has a lot of changes but no updates to the test

I added tests for the report handler. There I'm also testing the line printer - the updating printer is hard to test and I don't think it's worth the effort.

jochil (approved):

PR #47, opened 2022-06-01
Make cifuzz run work with CMake integration

fmeum:


internal/cmd/run/run.go
```diff
@@ -19,12 +19,16 @@ import (
...
 	"code-intelligence.com/cifuzz/pkg/report"
+	"code-intelligence.com/cifuzz/pkg/runfiles"
 	"code-intelligence.com/cifuzz/pkg/runner/libfuzzer"
 	"code-intelligence.com/cifuzz/util/envutil"
 	"code-intelligence.com/cifuzz/util/fileutil"
 	"code-intelligence.com/cifuzz/util/stringutil"
 )
 
+// The CMake configuration (also called "build type") to use for fuzzing runs.
+const cmakeBuildConfiguration = "RelWithDebInfo"
```
adombeck, line 30: Could you explain in the comment why we want to use this build type?

adombeck (approved):

internal/cmd/run/run.go
```diff
@@ -289,15 +300,31 @@ func commonBuildEnv() ([]string, error) {
...
+	// On Windows, our preferred compiler is MSVC, which can't easily be run
+	// from an arbitrary terminal as it requires about a dozen environment
+	// variables to be set correctly. Thus, we assume users to run cifuzz from
+	// a developer command prompt anyway and thus don't need to set the
+	// compiler explicitly.
+	if runtime.GOOS != "windows" {
+		// Set the C/C++ compiler to clang/clang++, which is needed to build a
+		// binary with fuzzing instrumentation (gcc doesn't have
+		// -fsanitize=fuzzer).
+		cc, err := runfiles.Finder.ClangPath()
```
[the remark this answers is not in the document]
fmeum: That is one of two arguments for doing it, the other being that without using the runfiles finder, this won't work with our shipped clang in core.

internal/cmd/run/run.go
fmeum: I will add a reference to the full explanation in the CMake integration.

PR #53, opened 2022-06-07
Add pre-push hook

adombeck:


adombeck:
@jochil: I couldn't find a good YAML formatter to use here. In the core repo we use https://prettier.io/docs/en/install.html but I don't know if we want to depend on npm/yarn here.

jochil:
> @jochil: I couldn't find a good YAML formatter to use here. In the core repo we use https://prettier.io/docs/en/install.html but I don't know if we want to depend on npm/yarn here.

I thought about that too... if prettier is the only npm package we would use, this might be overkill, especially for 3 yaml files.
If we add npm we can also use https://typicode.github.io/husky/ for the git hooks

adombeck:
> If we add npm we can also use https://typicode.github.io/husky/ for the git hooks

Please not, husky was super annoying when we introduced it in the core repo

jochil:
> > If we add npm we can also use https://typicode.github.io/husky/ for the git hooks
> 
> Please not, husky was super annoying when we introduced it in the core repo

So what's about running prettier only in the CI Pipeline? https://github.com/marketplace/actions/prettier-action

adombeck:
> So what's about running prettier only in the CI Pipeline? https://github.com/marketplace/actions/prettier-action

A bit annoying when the pipeline fails because you didn't format your file the way we want but we don't provide the means to actually format the file that way, no?

adombeck:
> I meant to actually format the file and commit the changes... during the pipeline

The CI automatically adding commits to your branch? I'm skeptical

fmeum:
I also don't like pipelines modifying my commits. Have you taken a look at https://github.com/UltiRequiem/yamlfmt?

jochil:
> > I meant to actually format the file and commit the changes... during the pipeline
> 
> The CI automatically adding commits to your branch? I'm skeptical

It can be done on the commit (amend) who changed the yaml file. But I agree... it can cause problems/confusion. Another option would be on the merge into main... 
I am not sure if this is worth the potential troubles

adombeck:
> Have you taken a look at https://github.com/UltiRequiem/yamlfmt?

Yes, but same as `yq` - and in contrast to prettier - it removes empty lines which can help make the file more readable.

jochil (approved):

PR #56, opened 2022-06-07
Add Apache 2.0 license

HenrichN:


fmeum:
@HenrichN Do we need license headers on all source files? Jazzer [has them](https://github.com/CodeIntelligenceTesting/jazzer/blob/6e61ef2e76213097b36b14507a905e5dcd5a78bd/agent/src/main/java/com/code_intelligence/jazzer/agent/Agent.kt#L1), but they definitely do add noise.

HenrichN:
@fmeum Good question. According "to the internet" (https://softwareengineering.stackexchange.com/questions/317041/should-i-add-the-license-in-every-header-and-source-file) it is not needed. Question remains, if it is something that _we_ might want. I am unsure what additional benefits this brings. Do some come to your mind?

fmeum:
I don't see a benefit - most likely only lawyers would. ;-) We can always add them later if we should decide they are needed after all.

PR #57, opened 2022-06-08
installer: Install CMake integration

fmeum:


tools/install/installer.go
```diff
@@ -64,6 +65,11 @@ func NewInstaller(opts *Options) (*installer, error) {
 		fileutil.Cleanup(opts.InstallDir)
 		return nil, errors.WithStack(err)
 	}
+	err = os.MkdirAll(i.shareDir(), 0700)
```
adombeck, line 79: I see that this is my fault, because I used the same permissions for `lib` and `bin`, but I think we should install all files and directories readable by group and others (unless `umask` removes those bits, which os.MkdirAll takes care of).

internal/cmd/init/init.go
```diff
@@ -47,9 +47,27 @@ func run(cmd *cobra.Command, args []string, opts *cmdOpts) (err error) {
...
 	log.Info(`
 Use 'cifuzz create' to create your first fuzz test`)
 	return
 }
+
+func printBuildSystemInstructions(cwd string) {
+	// Printing build system instructions is best-effort: Do not fail on errors.
+	cfg, err := config.ReadProjectConfig(cwd)
+	if err != nil {
+		return
```
adombeck, line 63: Should we at least print the error via `log.Debug`?

adombeck: If it's something else than a `os.ErrNotExist`

tools/install/installer.go
```diff
@@ -96,9 +97,11 @@ func (i *installer) InstallCIFuzzAndDeps() error {
...
-	if err != nil {
-		return err
+	if runtime.GOOS != "windows" {
+		err = i.InstallMinijail()
+		if err != nil {
+			return err
+		}
 	}
 
 	err = i.InstallProcessWrapper()
```
adombeck, line 117: We also don't have to install the process wrapper on Windows because it's only used along with minijail.

internal/cmd/run/run.go
```diff
@@ -404,6 +405,17 @@ func setBuildFlagsEnvVars(env []string) ([]string, error) {
 		return nil, err
 	}
 
+	// Users should pass the environment variable FUZZ_TEST_CFLAGS to the
+	// compiler command building the fuzz test.
```
adombeck, line 409: Do we tell them that anywhere?

tools/install/installer.go
```diff
@@ -96,9 +97,11 @@ func (i *installer) InstallCIFuzzAndDeps() error {
 		return err
 	}
 
-	err = i.InstallMinijail()
-	if err != nil {
-		return err
+	if runtime.GOOS != "windows" {
```
jochil, line 110: Do we know that minijail is working on MacOSX/darwin? If not we should change it to `runtime.GOOS == "linux"`

tools/install/installer.go
```diff
@@ -262,6 +263,20 @@ func (i *installer) InstallCMakeIntegration() error {
...
+	if runtime.GOOS == "windows" {
+		// TODO: On Windows, users generally don't expect having to fiddle with their PATH. We should update it for
+		//       them, but that requires asking for admin access.
+		fmt.Fprintf(os.Stderr, `
+Please add the following directory to your PATH:
+    %s`, i.binDir())
+	} else {
+		fmt.Fprintf(os.Stderr, `
+Please add the following to ~/.profile or ~/.bash_profile:
+    export PATH="$PATH:%s"`, i.binDir())
```
jochil, line 276: The overall indention of this block is very confusing... maybe we should think about using something like https://github.com/makenowjust/heredoc

adombeck: I didn't have any issues reading this block and understanding the indentation but I would be fine with using that package

tools/install/installer.go
fmeum: Good point, seems like it's only working on Linux. Changed.

tools/install/installer_unix.go
```diff
@@ -0,0 +1,30 @@
...
+)
+
+func registerCMakePackage(packageDir string) error {
+	// Create ~/.cmake/packages/CIFuzz/CIFuzz containing the path to the root directory of the CMake integration.
+	// See https://cmake.org/cmake/help/latest/manual/cmake-packages.7.html#user-package-registry
+	homeDir, err := os.UserHomeDir()
+	if err != nil {
+		return errors.WithStack(err)
+	}
+	cmakePackagesDir := filepath.Join(homeDir, ".cmake", "packages", "CIFuzz")
```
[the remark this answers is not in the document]
fmeum: There is no system registry on non-Windows platforms, which seems weird. I'm using the system package registry on Windows now. On Linux, at least using `sudo` should work for the current user since the value of `HOME` is not changed by `sudo`, right?

internal/cmd/init/init.go
fmeum: Done

internal/cmd/run/run.go
fmeum: We don't tell them anything about our make integration yet. We should, but not in this PR :-)

tools/install/installer.go
jochil, line 276: It seems like there is a linebreak needed... I have the following output (see the % at the end):
```
Please add the following to ~/.profile or ~/.bash_profile:
    export PATH="$PATH:/home/jochen/cifuzz/bin"%  
```

fmeum: The package documentation is very minimal and doesn't really define how it handles relative indentation and new lines. I agree that it makes for code that is visually easier to parse, but at the same time makes it impossible to determine what the actual output on screen will be like without reading the code of the package itself.

If there is a Go package for this that defines what it does as succinctly and completely as https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/trim-indent.html, I would consider using it.

tools/install/installer_unix.go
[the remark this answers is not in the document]
adombeck: > On Linux, at least using sudo should work for the current user since the value of HOME is not changed by sudo, right?

No, sudo does not preserve $HOME

I think we should install to a directory where CMake will find the package due to it's weird search procedure, like `/usr/local/share/cifuzz/cmake`.

tools/install/installer.go
jochil: sounds reasonable... I will take a look

tools/install/installer_unix.go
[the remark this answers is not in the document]
fmeum: Pushed a new commit that installs to `/usr/local/CIFuzz` as root and tested it manually. `/usr/local/share` isn't guaranteed to work, the standard paths include only `/usr/local` and `/usr`.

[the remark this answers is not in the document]
adombeck: From https://cmake.org/cmake/help/latest/command/find_package.html#config-mode-search-procedure:
> <prefix>/(lib/<arch>|lib*|share)/cmake/<name>*/                 (U)
> <prefix>/(lib/<arch>|lib*|share)/<name>*/                       (U)
> <prefix>/(lib/<arch>|lib*|share)/<name>*/(cmake|CMake)/         (U)

Those include `/usr/local/share`, when `<prefix>` is `/usr/local`, no?

And I would prefer the lowercase `cifuzz`, it's uncommon for directories on linux to contain uppercase letters, and cmake searches case-insensitive:
> In all cases the <name> is treated as case-insensitive and corresponds to any of the names specified (<PackageName> or names given by NAMES).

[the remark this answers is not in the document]
fmeum: It does, it just requires restructuring the CMake module contents. Now that we rely on the registry mostly and don't need to fit the patch search procedure on both Unix and Windows, this is a reasonable refactoring I will now do.

I will rename to cifuzz throughout.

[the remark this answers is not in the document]
fmeum: Done, PTAL.

internal/cmd/init/init.go
```diff
@@ -47,9 +47,30 @@ func run(cmd *cobra.Command, args []string, opts *cmdOpts) (err error) {
...
+		if !errors.Is(err, os.ErrNotExist) {
+			log.Debug(err)
+		}
+		return
+	}
+	if cfg.BuildSystem == config.BuildSystemCMake {
+		log.Infof(`
+Enable fuzz testing in your CMake project by adding the following lines to the top-level CMakeLists.txt:
+
+    find_package(cifuzz)
```
adombeck, line 72: I didn't mean that you should rename the CMake package itself, I only meant the `/usr/local/share/cifuzz` directory which should be lowercase IMO, because this is a standard location where apps can install data files and those directories are usually lowercase. Note that this directory  is not CMake-specific, we could also put other files in there. The CMake-specific files in there should follow the CMake naming conventions, which might be to use uppercase (?).  Same for `cifuzzTargets.cmake` and `cifuzzFunctions.cmake`.

tools/install/installer.go
```diff
@@ -224,3 +289,42 @@ func findProjectDir() (string, error) {
 	}
 	return projectDir, nil
 }
+
+// copyCmakeIntegration copies the CMake integration to destDir and returns the
```
adombeck, line 293:
```suggestion
// copyCMakeIntegration copies the CMake integration to destDir and returns the
```

internal/cmd/init/init.go
fmeum: I wasn't so sure about it because our tool is called `cifuzz`, not `CIFuzz`. But an idiomatic CMake module name would certainly look more like `CIFuzz`, so I guess I'll revert that part.

fmeum: Done

adombeck (approved):

fmeum:
@adombeck [This](https://github.com/CodeIntelligenceTesting/cifuzz/runs/6834598731?check_suite_focus=true#step:9:918) might be an interesting failure on macOS for you to investigate.

PR #58, opened 2022-06-09
improve documentation

jochil:
improve the documentation so new users have an easy start


examples/cmake/my_fuzz_target.cpp
```diff
@@ -0,0 +1,26 @@
...
+    a = data[0] | (data[1] << 8) | (data[2] << 16) | (data[3] << 24);
+
+    if (size >= 8) {
+      b = data[4] | (data[5] << 8) | (data[6] << 16) | (data[7] << 24);
+
+      if (size >= 12) {
+        c = string(reinterpret_cast<const char *>(data + 8), size - 8);
+      }
+    }
+  }
```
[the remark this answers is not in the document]
adombeck: Why are we shifting the bytes?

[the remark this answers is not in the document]
jochil: I added some comments

docs/Glossary.md
```diff
@@ -1,3 +1,44 @@
...
+## Fuzzer aka Fuzzing Engine
+Is a piece of software that generates input to feed into the system under test (SUT) via the fuzz targets.
+A coverage guided fuzzer will also collect information about the covered code during the execution and will try to 
+improve coverage by adjusting the inputs accordingly.
+
+## Fuzz Target
+Usally a function that takes an array of bytes from the fuzzer and calls functions from the system under test.
+
+## Fuzzing aka Fuzzing Run
+Fuzzing describes the act of running the fuzzer against a fuzz target.
```
adombeck, line 13: We should get rid of the "aka"s here IMO. We'll have to have a discussion about which terms to keep and also about "Fuzz Target" vs "Fuzz Test".

docs/Glossary.md
```diff
@@ -1,3 +1,44 @@
...
+
+## Corpus
+A set of useful test inputs
+
+### Seed Corpus
+A small corpus usually provided by the user. Findings can be added to this corpus to use it during regression tests.
```
adombeck, line 19:
```suggestion
A small corpus usually provided by the user. Findings can be added to this corpus for use in regression tests.
```

docs/Glossary.md
```diff
@@ -1,3 +1,44 @@
...
+
+### Generated Corpus
+While executing a fuzzing run the fuzzer might discover useful inputs. These inputs are added to the generated corpus.
```
adombeck, line 22:
```suggestion
The generated corpus contains interesting inputs found by the fuzzer while running the fuzz test.
```

docs/Glossary.md
```diff
@@ -1,3 +1,44 @@
...
+
+## Replayer
+A Replayer just applies inputs stored in the (seed) corpus to the fuzz target without generating new input.
```
adombeck, line 25:
```suggestion
A replayer runs inputs from the seed corpus against the fuzz target without generating new input.
```
(using wording similar to https://go.dev/doc/fuzz/#command-line-output)

docs/Glossary.md
```diff
@@ -1,3 +1,44 @@
...
+It can be used during debugging and/or for regression testing.
```
adombeck, line 26:
```suggestion
It can be used for debugging and/or regression testing.
```

docs/Glossary.md
```diff
@@ -1,3 +1,44 @@
...
+## Sanitizer
+Santizers use code instrumentation to detect bugs during the execute of an application. 
+
+## Finding
+A finding describes a bug or vulnerability found by the fuzzer and includes the input that causes it.
+
+## Metrics
+While using **cifuzz** you will get in touch with metrics descriping the progress or outcome of a fuzzing run. 
+Some of them might have different meanings depending on the context, previous knowledge or might even be used 
+differntly by the various fuzzing engines. To avoid confusion we provide the following definitions:
```
adombeck, line 37:
```suggestion
differently by the various fuzzing engines. To avoid confusion we provide the following definitions:
```

README.md
```diff
@@ -1,9 +1,65 @@
 # cifuzz
 
+> **_IMPORTANT:_** This project is under active development. Be aware that the behavior of the commands
```
adombeck, line 3: For better readability in terminals, I would prefer to wrap all READMEs on 72 characters.

docs/Glossary.md
jochil: I removed it and I agree to the Test vs Target topic... we should discuss it and adjust the glossary then

docs/Glossary.md
jochil: done

docs/Glossary.md
jochil: done

docs/Glossary.md
jochil: done

docs/Glossary.md
```diff
@@ -1,3 +1,44 @@
...
+
+### Generated Corpus
+While executing a fuzzing run the fuzzer might discover useful inputs. These inputs are added to the generated corpus.
+
+## Replayer
+A Replayer just applies inputs stored in the (seed) corpus to the fuzz target without generating new input.
+It can be used during debugging and/or for regression testing.
+
+## Sanitizer
+Santizers use code instrumentation to detect bugs during the execute of an application.
```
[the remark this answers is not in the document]
jochil: done

docs/Glossary.md
jochil: done

README.md
jochil: done

adombeck:
Good work! I'll take another quick look tomorrow.

README.md
```diff
@@ -1,9 +1,74 @@
...
+make install
+```
+If everything went fine, you will find the newly created directory 
+`~/cifuzz`. Do not forget to add `~/cifuzz/bin` to your `$PATH`.
+
+### Setup / Create your first fuzz target
+**cifuzz** commands will interactively guide you through the needed 
+options and shows additional next steps. You will also find a complete 
+list of the available commands with all supported options and 
+parameters by calling `cifuzz command --help` or
```
adombeck, line 53:
```suggestion
options and show next steps. You can find a complete 
list of the available commands with all supported options and 
parameters by calling `cifuzz command --help` or 
```

or 

```suggestion
options and show next steps. A complete 
list of the available commands with all supported options and 
parameters can be found by calling `cifuzz command --help` or 
```

README.md
```diff
@@ -1,9 +1,74 @@
...
+will create a stub for your fuzz test (aka fuzz target), lets say it 
+is called `my_fuzz_test.cpp`.
+
+3. Edit `my_fuzz_test.cpp` so it actually calls the function you want 
+to test with the input generated by the fuzzer. To learn more about 
+writing fuzz tests you can take a look at our 
+[Tutorial](docs/How-To-Write-A-Fuzz-Test.md) or one of the 
+[example projects](examples).
+
+4. Start the fuzzing by executing `cifuzz run my_fuzz_test`.
```
adombeck, line 70: I think there's a step missing here: The user needs to make sure that the fuzz target can be built, either via `cmake` or any other build system if they provide the `--build-command` flag to `cifuzz run`.

README.md
jochil: done

README.md
adombeck: What do you mean with "make it more explicit on the 4."? I think the user should not run into failures when they follow these steps.

jochil: > Start the fuzzing by executing cifuzz run my_fuzz_test. cifuzz now tries to build the fuzz test and starts a fuzzing run.

This is currently 4. Maybe changing it to something like this:

> Start the fuzzing by executing cifuzz run my_fuzz_test. cifuzz will build the fuzz test before executing it, so make sure it is compiling without errors (cifuzz will tell you there is a problem).

I would like to avoid adding another point because of the following reasons:
1. Without changing the fuzz target `cifuzz run` should just run fine
2. When changing code everyone knows that it has to be compilable and usually knows how to do it. Especially when we tell him to add the CMake commands when creating the target

adombeck: > Without changing the fuzz target cifuzz run should just run fine

It doesn't though if all you do is follow the steps listed here, because cifuzz can't build the fuzz target. If the build system is cmake, you have to add the "add_fuzz_test" line printed by `cifuzz create` to the `CMakeLists.txt` in order for `cifuzz run` to work, and if the build system is unknown the user has to figure out on their own how to build the fuzz target. So IMO there is clearly a "configure your build system to build the fuzz target" step missing here (with a pointer to the output of `cifuzz create`).

jochil: I expect the user to follow the instructions given in the command... if not we also have to point it out in step 1 for 
```
find_package(cifuzz)
enable_fuzz_testing()
```
as well and basically for all build systems / cases later on.
And I think we should not cover the "unknown/other" build system part for now, as we are currently focusing on the CMake integration. If we supporting other build system I would prefer linking to an additional page covering all the build system specific stuff.

I would be ok with writing something like "please follow the instructions printed out by the command" on Step 3.

jochil: And we can also rework the steps after having actual user feedback.

adombeck: > I would be ok with writing something like "please follow the instructions printed out by the command" on Step 3.

I would be fine with that

adombeck (approved):

PR #71, opened 2022-06-13
Remove bazel dependency

adombeck:


jochil:
Ok... but there is no update on the go.mod and we still have this dependency `github.com/bazelbuild/rules_go v0.33.0`
And the runner is still using `fileutil.TempDir` with the bazel dep
What am I missing?

adombeck:
> Ok... but there is no update on the go.mod and we still have this dependency `github.com/bazelbuild/rules_go v0.33.0` And the runner is still using `fileutil.TempDir` with the bazel dep What am I missing?

You are right, I forgot to actually delete`fileutil.TempDir` and `fileutil.TempFile`. Did that now.

PR #74, opened 2022-06-14
feat: support c/c++ language server (ccls)

jochil:


adombeck:
How can I test this?

adombeck:
Please rebase on main to get rid of the merge commit

jochil:
> Please rebase on main to get rid of the merge commit

that happens if you let github update the PR :) Fixed it

jochil:
> How can I test this?

You need an editor/ide with ccls (https://github.com/MaskRay/ccls) installed. vscode extension: https://marketplace.visualstudio.com/items?itemName=ccls-project.ccls

If this is the case the editor should be able to resolve the includes, for example in `examples/cmake/my_fuzz_target.cpp`

adombeck:
> that happens if you let github update the PR :) Fixed it
Yeah that's annoying, you have to click the arrow next to the "Update branch" button and choose "Update with rebase" in order to avoid the merge commit. I looked for an option in the repository settings to default to rebasing but couldn't find any.

adombeck:
> If this is the case the editor should be able to resolve the includes, for example in examples/cmake/my_fuzz_target.cpp

Ok but that's only used by the editor, not during build? So then the editor will not show errors for missing includes but the build still fails if CMake wasn't configured to also know about the include dir? What happens when CMake was configured but this `.ccls` file does not exist, will the editor (say, vscode) know about the include dir?

jochil:
> Ok but that's only used by the editor, not during build? 

Yes. It has no effects on our CMake build

> So then the editor will not show errors for missing includes but the build still fails if CMake wasn't configured to also know about the include dir? 

Yes that's right too. But CMake is handled by the installer

> What happens when CMake was configured but this `.ccls` file does not exist, will the editor (say, vscode) know about the include dir?

It depends heavily on the editor. vscode with the C/C++ ms-vscode extension is still able to find the include dir. But vim with the ccls language server has no clue about the dir without the `.ccls` file. I guess it will be the same for emacs

adombeck:
> But vim with the ccls language server has no clue about the dir without the .ccls file. I guess it will be the same for emacs

Ok, so this is useful for vim and probably emacs. Fine by me.

adombeck (approved):

jochil:
> Yeah that's annoying, you have to click the arrow next to the "Update branch" button and choose "Update with rebase" in order to avoid the merge commit. I looked for an option in the repository settings to default to rebasing but couldn't find any.

I was very progressive and did it in the app... unfortunately there is no option to chose the merge strategy :)

PR #75, opened 2022-06-14
finder: Fix cifuzz include path

adombeck:
The path uses the lowercase "cifuzz" instead of "CIFuzz".


jochil:
Shouldn't this result in an error with case sensitive filesystems before? I did not notice a problem... 
And maybe a test would have been nice covering this

adombeck:
> Shouldn't this result in an error with case sensitive filesystems before? I did not notice a problem...

It didn't fail because there was no test which uses this code and apparently no one tried building with an unknown buildsystem since this was merged last week (this function is only called when building with an unknown buildsystem).

> And maybe a test would have been nice covering this

That will be covered by an integration test using an unknown buildsystem, which is already planned for this sprint.

jochil (approved):

PR #76, opened 2022-06-14
Fuzz test shell completion

adombeck:


jochil:
A short summary of the solution would have been nice. So this is what I got:
For cmake you are checking the build directory for a file pattern. The build directory can be the result of `cifuzz run` or `cifuzz reload`. For an unknown build system it falls back to the file/dir completion.
The files used in the pattern are created by invoking `cmake` and this is part of our cmake shared lib
It this right?

jochil:
I think it makes sense to add some unit tests for `internal/build/cmake/cmake.go` and `internal/build/build.go`

adombeck:
> A short summary of the solution would have been nice. So this is what I got:
For cmake you are checking the build directory for a file pattern. The build directory can be the result of cifuzz run or cifuzz reload. For an unknown build system it falls back to the file/dir completion.

That's correct.

> The files used in the pattern are created by invoking cmake and this is part of our cmake shared lib

If by "cmake shared lib" you mean the new cmake package, then that's also correct (more specifically, it's `cmake.Configure()`).

adombeck:
> I think it makes sense to add some unit tests for internal/build/cmake/cmake.go and internal/build/build.go

I took another look but I don't see a way to add unit tests which make sense. I don't want to add unit tests like "Check that func X sets the environment variably Y and Z" or "Check that func X executes command Y", because those have little value IMO. They would just duplicate the string values of the function under test and that would make changes annoying because the tests have to be kept in sync.

jochil:
> I took another look but I don't see a way to add unit tests which make sense. I don't want to add unit tests like "Check that func X sets the environment variables Y and Z" or "Check that func X executes command Y", because those have little value IMO. They would just duplicate the string values of the function under test and that would make changes annoying because the tests have to be kept in sync.

In general: The tests might be simple and not covering a lot of code, but that should be the nature of unit tests... this is what make them more stable. I understand the argument witch checking strings/envs and it can be annoying to adjust the tests (but in my experience this kind of stuff does not change a lot), but even in this case a tests protects you against a change in the function under tests that unintentionally drops/change/whatever one of the strings/envs. Especially while doing a refactoring.
Of course we do reviews,, but you can't be sure that we don't make mistakes. 
Just imagine all the ways some can mess up a function :)

In detail: 
`internal/build/build.go` behaves different on windows... I think that can be secured by a test.

`internal/build/cmake/cmake.go`:
- L44: add tests that make sure that the build dir is unique on different option sets
- Are all the relevant arguments/ cache variables added to the command call?
- Does it catches problematic strings, for example when some of the opts used for build dir contain a `/`

It is not a lot, but I think it is a good style to just do it ... even for small packages/functions

internal/cmd/run/run.go
```diff
@@ -511,3 +513,30 @@ func countSeeds(seedDirs []string) (numSeeds uint, err error) {
 	}
 	return numSeeds, nil
 }
+
+func validFuzzTests(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+	// Run the PersistentPreRunE function which takes care of changing
+	// the directory if the `--directory` flag was used.
```
jochil, line 435: Can calling the `PersistentPreRunE` causes problems when the parent commands changes or if the function will do more than changing the directory?

jochil:
If we want to get rid of the `reload` command soon we should create a ticket for it or ad least have a comment pointing out why it exists and that it is only temporary 

If we keep it for an unknown period, we should add an integration test

internal/cmd/run/run.go
adombeck: I extracted a `cmdutils.Chdir()` function, please take a look.

adombeck:
> If we want to get rid of the reload command soon we should create a ticket for it or at least have a comment pointing out why it exists and that it is only temporary

I created https://code-intelligence.atlassian.net/browse/CLI-230 and added a TODO comment.

internal/cmd/run/run.go
jochil: looks better, thanks

internal/cmd/run/run.go
```diff
@@ -507,3 +509,22 @@ func countSeeds(seedDirs []string) (numSeeds uint, err error) {
 	}
 	return numSeeds, nil
 }
+
+func validFuzzTests(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+	// Change the directory if the `--directory` flag was set
+	err := cmdutils.Chdir()
```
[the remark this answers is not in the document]
adombeck: It is necessary, `validFuzzTests` is used as the `ValidArgFunction`, which is called when you have the autocompletion script installed and press `<Tab>` after typing `cifuzz run` in the terminal. `PersistentPreRunE` is not executed before the `ValidArgFunction` is called. That's the reason I explicitly called the `PersistentPreRunE` here before, see https://github.com/CodeIntelligenceTesting/cifuzz/pull/76#pullrequestreview-1007444590.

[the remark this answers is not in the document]
jochil: Ah of course... I forgot that already :)

adombeck:
> internal/build/build.go behaves different on windows... I think that can be secured by a test.

I really don't see any benefit in adding a test for `CommonBuildEnv()`, any test I can imagine will only test that `envutil.Setenv` behaves as expected and that is already tested in `util/envutil/envutil_test.go`.

> internal/build/cmake/cmake.go:
> * L44: add tests that make sure that the build dir is unique on different option sets

I added a test for that.

> Are all the relevant arguments/ cache variables added to the command call?

We can't really test that currently, because we can't check which arguments and environment variables are used for `exec.Command`. For that we would have to do something similar to `gh` for that, which is to create and use a wrapper for `exec.Command` (see https://github.com/cli/cli/blob/0e3c3bb4a428eff7fcff472a34e5208561a9406c/internal/run/run.go#L21-L21) and then use something that can check if the correct arguments were used in tests (see https://github.com/cli/cli/blob/0e3c3bb4a428eff7fcff472a34e5208561a9406c/internal/run/stub.go#L18-L18). I don't think the tests for this file are worth the effort.

> Does it catches problematic strings, for example when some of the opts used for build dir contain a /

I don't think a build dir containing a `/` is problematic, it will just create a subdirectory and use that as the build dir.

jochil:
> I really don't see any benefit in adding a test for `CommonBuildEnv()`, any test I can imagine will only test that `envutil.Setenv` behaves as expected and that is already tested in `util/envutil/envutil_test.go`.

It will test if different variables are set on windows and linux.

> We can't really test that currently, because we can't check which arguments and environment variables are used for exec.Command. For that we would have to do something similar to gh for that, which is to create and use a wrapper for exec.Command (see https://github.com/cli/cli/blob/0e3c3bb4a428eff7fcff472a34e5208561a9406c/internal/run/run.go#L21-L21) and then use something that can check if the correct arguments were used in tests (see https://github.com/cli/cli/blob/0e3c3bb4a428eff7fcff472a34e5208561a9406c/internal/run/stub.go#L18-L18). I don't think the tests for this file are worth the effort.

Ok, but we should keep that in mind... stubbing/mocking the `exec.Command` can actually be helpful to speed up and isolate some tests. If we come to that point

These were just a few examples that quickly came to my mind. I am fine with the current state

PR #78, opened 2022-06-17
Handle expected error that cifuzz.yaml can't be found

adombeck:


jochil:
I think this is a good change, but we should do the check for basically every command except `init`, or?
Is there a way to do it in the `PersistentPreRun` or `PreRun` of the root command? This can avoid a lot of duplicated code

adombeck:
> I think this is a good change, but we should do the check for basically every command except `init`, or? Is there a way to do it in the `PersistentPreRun` or `PreRun` of the root command? This can avoid a lot of duplicated code

We would still have to call `config.FindProjectDir()` in the commands which use the path to the project dir, because we can't pass that value from the `PersistentPreRun` function, and then we would have to handle errors there again. We could add an `AssertProjectDir()` function which calls `config.FindProjectDir()` and logs the error, which we then call in all commands except `init` to reduce code duplication. WDYT?

jochil:
> We would still have to call `config.FindProjectDir()` in the commands which use the path to the project dir, because we can't pass that value from the `PersistentPreRun` function, and then we would have to handle errors there again. We could add an `AssertProjectDir()` function which calls `config.FindProjectDir()` and logs the error, which we then call in all commands except `init` to reduce code duplication. WDYT?

Sounds already good and I like the `AssertProjectDir`... but I am still thinking that there might be a less redundant solution to this.
What's about working with a pointer to the `ProjectConfig` created in `init()` of the root command and injected to the subcommands when creating them
```
	rootCmd.AddCommand(initCmd.New())
	rootCmd.AddCommand(createCmd.New(config))
	rootCmd.AddCommand(runCmd.New(config))
```
And now loading the config in the 'PersistentPreRun` of the root?

adombeck:
@jochil I tried to implement your suggestion, please take a look

jochil:
thanks... looks good to me. 

As we know I am a little bit obsessed with unit tests ;), so personally I would like to have one for `pkg/cmdutils/config_check.go` but I am fine without... so I already approved

jochil (approved):

PR #80, opened 2022-06-17
Move the runner package integration tests to the runner package

adombeck:
To make the relationship more clear and to avoid confusion with the
integration test for `cifuzz run` which we're about to add.

Also adds "_testutils" to the files containing test utilities to make it
easier to identify them as test utils by their basename (for example
when doing a global search in Intellij).


jochil:
I am fine with the `_testutils` renaming, but I would prefer leaving the tests in the integration directory. Tests inside a package should match the scope of the package, but as this are integration tests they actually cover more than this single package. Also I think it is good having the tests close to the testdata... but moving the `testdata` folder to the package would feel very wrong.

And I do not see how this can become confusing when adding the `run` integration tests... which btw are not going to the `internal/cmd/run/` package, or?

jochil:
> I agree that tests inside packages should match the scope of the package, but IMO that's the case here, we are only testing that the runner package works as expected with various libfuzzer binaries.

Ok, I would even go so far and say I prefer only unit tests directly in the package. And these tests are also testing at least the `report` and the `parser` package (but ok, mocking this is sometimes not worth the effort), but more importantly they are calling external commands... this is why I would call them integration tests and they should be more toplevel.

Another point of view. If I see a `foo_test.go` besides a `foo.go` I would expect that I can run it immediately without any external dependcies, which is not the case as the fuzz targets need to get compiled and minijail has to be installed.

And my last reason: Stuff in `pkg` is meant to get used by other tools/libraries (see https://github.com/golang-standards/project-layout/tree/master/pkg). And these tests are not really transferable... 

> As long as the testdata directory is only used by these tests, we should move it to the same package IMO. If at some point we decide that it makes sense to share this testdata between multiple tests, we should move it up to a level above these tests.

that means moving c/cpp files, a corpus, gitignore, some random file and a Makefile into the go package structure. For me this sounds very messy and not so desirable

> That will be a test which executes the cifuzz binary and has therefore dependencies on a lot of packages, so I'm leaning towards putting that into a top-level integration test directory, but I would also be fine with putting it in internal/cmd/run/integration-test.

I totally agree... For me this absolutely belongs on the top-level :)

jochil:
> The parser package is only used by the runner package, so we should probably move that into a subdirectory of runner.
I would like that

> > Another point of view. If I see a `foo_test.go` besides a `foo.go` I would expect that I can run it immediately without any external dependcies, which is not the case as the fuzz targets need to get compiled and minijail has to be installed.
> 
> True, I would also prefer to do that as part of the test, like we do [here](https://github.com/CodeIntelligenceTesting/cifuzz/blob/move-runner-integration-tests/tools/cmake/cmake_test.go#L38-L48).

Calling the Makefile from go seems a little bit more complicated than needed... I think the current setup with invoking it from the root Makefile is quite simple and very straightforward 

To wrap it up a little bit: I understand the reasons behind moving the tests to the package, but it still seems not very intuitive and clean to me. But that might be only my impression. 
But if we move the tests to the package I would prefer to keep the testdata as close as possible and the test invocation simple

adombeck:
> > The parser package is only used by the runner package, so we should probably move that into a subdirectory of runner.
> 
> I would like that

Another point we can agree on then :)

> > > Another point of view. If I see a `foo_test.go` besides a `foo.go` I would expect that I can run it immediately without any external dependcies, which is not the case as the fuzz targets need to get compiled and minijail has to be installed.
> > 
> > 
> > True, I would also prefer to do that as part of the test, like we do [here](https://github.com/CodeIntelligenceTesting/cifuzz/blob/move-runner-integration-tests/tools/cmake/cmake_test.go#L38-L48).
> 
> Calling the Makefile from go seems a little bit more complicated than needed... I think the current setup with invoking it from the root Makefile is quite simple and very straightforward
> 
> To wrap it up a little bit: I understand the reasons behind moving the tests to the package, but it still seems not very intuitive and clean to me. But that might be only my impression. But if we move the tests to the package I would prefer to keep the testdata as close as possible and the test invocation simple

Maybe @fmeum has an opinion which could help us make a decision here. I will prepare another PR which contains the changes we can agree on (renaming the util files and making the `parser` package to a subpackage of `runner`) and would keep this open until we hear back from @fmeum, would that be ok for?

jochil:
> Maybe @fmeum has an opinion which could help us make a decision here. I will prepare another PR which contains the changes we can agree on (renaming the util files and making the `parser` package a subpackage of `runner`) and would keep this open until we hear back from @fmeum, would that be ok for?

Thanks... Sounds good for me.
But if you feel blocked or something please go ahead and just move the testdata, too. No need to wait, as we should be able to easily revert that change, or?

fmeum:
Chiming in from the sidelines due to having been mentioned a few times ;-)


I generally prefer to have tests of any kind as close to the core functionality they test as possible (and their required data right next to them, unless reused by tests on a higher level). As long as we put test data in the `testdata` subdirectory, I wouldn't say that this pollutes the package structure.

I wouldn't actually consider the runner tests to be special in any way, they just happen to require more data than an average test. Since the job of the runner package is to start and watch binaries, that's also what any test needs to do.

In general, I would prefer to rely on Make as little as possible. Ideally, `go test ./...` just works after installing dependencies once. Go seems well suited to do cross-platform work, including test setup. With our test helpers, Go covers caching reasonably well.

jochil:
Okay guys... I am convinced ;)
But can we agree on using Make (an actual build system) for building the binaries? I am ok with invoking the make target from go...

adombeck:
I implemented what we discussed, please take another look

PR #86, opened 2022-06-22
Simplify cifuzz create

adombeck:
Instead of having one flag to specify the output directory and one to
specify the filename, there is now only one flag to specify the output
path.

This also simplifies the code, allowing to get rid of the "pkg/storage"
package.

Note that the behavior changed a bit, in that output directories are not
created automatically anymore and creation of the fuzz target stub fails
if the output directory doesn't exist yet - which is what I would expect
from a command which creates a file.


pkg/dialog/dialog.go
```diff
@@ -46,3 +48,37 @@ func Input(message string, defaultValue string, inReader io.Reader) (string, err
...
+		return "", errors.WithStack(err)
+	}
+
+	if defaultValue == "" {
+		fmt.Printf("%s: \n", message)
+	} else {
+		fmt.Printf("%s [%s]: \n", message, defaultValue)
+	}
+
+	cmd := exec.Command(bash, "-c", "read -e && echo \"${REPLY}\"")
```
jochil, line 71: IMO this implementation is not very stable... for example I have bash installed but I am using zsh (oh-my-zsh). So now bash is printing into my zsh and it breaks with the normal completion ui/behavior I am used to.
I think we should not follow the path of making it compatible to the different shells.... I would prefer building it with the `promptui` package (but that looks complicated on a first view) or using another package offering a autcompletion feature already... like https://github.com/AlecAivazis/survey#suggestion-options. So we would have at least a consistent behavior in cifuzz

adombeck: > IMO this implementation is not very stable... for example I have bash installed but I am using zsh (oh-my-zsh). So now bash is printing into my zsh and it breaks with the normal completion ui/behavior I am used to.

I'm also using zsh with oh-my-zsh and yes, the completion behavior is slightly different: zsh prints the suggestions already when tab is pressed once while bash requires two tabs.

> I think we should not follow the path of making it compatible to the different shells....

The `complete` command we get from cobra can generate autocompletion scripts for bash, fish, zsh, and powershell. Maybe we could try to support completion for those shells in our dialogs as well. The zsh equivalent to `read -e` seems to be `vared` (see https://web.cs.elte.hu/zsh-manual/zsh_17.html). You can test it like this:
```
echo "Please enter filename [my_fuzz_test_1.cpp]:" && vared -c REPLY
```

> I would prefer building it with the promptui package (but that looks complicated on a first view) or using another package offering a autcompletion feature already... like https://github.com/AlecAivazis/survey#suggestion-options. So we would have at least a consistent behavior in cifuzz

At first I tried to use https://github.com/chzyer/readline but found it was too much effort to implement a nice completion function with that (and when I tried it garbled my terminal output).

jochil: > I'm also using zsh with oh-my-zsh and yes, the completion behavior is slightly different: zsh prints the suggestions already when tab is pressed once while bash requires two tabs.

Also I am not able to cycle through the entries. I was confused first when trying it out. Maybe we can avoid this happening to our users.

> The complete command we get from cobra can generate autocompletion scripts for bash, fish, zsh, and powershell. Maybe we could try to support completion for those shells in our dialogs as well.

That would be nice and the most consistent way I guess

I like the other commit... so what about moving the autocomplete thing to a separate task and do some research on it?

adombeck: > That would be nice and the most consistent way I guess

Ok, I'll try to add support for zsh now and later we could add support for fish and powershell.

> I like the other commit... so what about moving the autocomplete thing to a separate task and do some research on it?

Done, see https://github.com/CodeIntelligenceTesting/cifuzz/pull/90

jochil (approved):

PR #87, opened 2022-06-23
executil: Fix race when using CommandContext

adombeck:
By using exec.CommandContext, two go routines were started which waited
for the context to be done, one from the exec package which immediately
sends a SIGKILL to the started process and one from the executil package
which first sends a SIGTERM to the process group and after a grace
period a SIGKILL.

As a result, the started process itself usually didn't have time to shut
down gracefully, because it received a SIGKILL at about the same time as
the SIGTERM.

To fix this, executil.CommandContext now doesn't use exec.CommandContext
anymore but exec.Command, so the exec package doesn't handle killing the
process anymore. Because callers expect that using
executil.CommandContext will terminate the process when the context is
done, the process group is now always terminated when the context is
done and therefore the Cmd.TerminateProcessGroupWhenContextDone field
was removed.

### DIFF
--- a/pkg/runner/libfuzzer/libfuzzer_runner.go
+++ b/pkg/runner/libfuzzer/libfuzzer_runner.go
@@ -191,7 +191,6 @@ func (r *Runner) RunLibfuzzerAndReport(ctx context.Context, args []string, env [
 	}
 	defer cancelCmdCtx()
 	r.cmd = executil.CommandContext(cmdCtx, args[0], args[1:]...)
-	r.cmd.TerminateProcessGroupWhenContextDone = true
 	r.cmd.Env = env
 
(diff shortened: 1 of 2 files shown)


jochil:
Is there any chance for adding a test... maybe even a regression test for the old race condition?

fmeum:
While I would like a test, it would be very difficult: In addition to a test verifying that there is no race, we would ideally need a test to test that the test would have found a race if there was one - otherwise it is very easy to write a test that doesn't really have a chance to cover the original race.

Given that this race is pretty clear in hindsight (only) and the fix isn't difficult, I will thus merge this without a test.

fmeum (approved):

PR #90, opened 2022-06-24
Autocompletion in dialogs

adombeck:


adombeck:
Relevant comments: https://github.com/CodeIntelligenceTesting/cifuzz/pull/86#pullrequestreview-1018180449

fmeum:
@adombeck I refactored the code a bit so that adding support for other shells should become simpler (although the good old rule of three come back to hurt us here).

I also added a commit that suppresses stack traces if the user hits Ctrl + C.

PR #96, opened 2022-06-30
integration-tests: Enable on Windows

fmeum:


fmeum:
@adombeck I think we should merge this after your review and then, asynchronously, work on fixing the remaining issues on Windows.

PR #98, opened 2022-07-01
runner: Print startup stderr in case of an unexpected exit error

fmeum:
Without the stderr of the fuzzer printed during startup, it isn't
possible for users to understand why it exits abnormally, e.g. due to a
missing shared library dependency.

With this commit, the parsed output up to the point where the fuzzer has
been initialized is stored and made available to the runner to print in
case of an error.

### DIFF
--- a/pkg/runner/libfuzzer/libfuzzer_runner.go
+++ b/pkg/runner/libfuzzer/libfuzzer_runner.go
@@ -265,6 +265,12 @@ func (r *Runner) RunLibfuzzerAndReport(ctx context.Context, args []string, env [
 			}
 
 			if !IsExpectedExitError(err) {
+				// Print the stderr output of the fuzzer up to the point where
+				// it has been successfully initialized to provide users with
+				// the context of this abnormal exit even without verbose mode.
+				if !r.Verbose {
+					_, _ = log.NewPTermWriter().Write(reporter.StartupOutput())
+				}
 				return errors.WithMessagef(err, "Unexpected exit code %d", exitErr.ExitCode())
 			}
 
--- a/pkg/runner/libfuzzer/libfuzzer_runner.go
+++ b/pkg/runner/libfuzzer/libfuzzer_runner.go
@@ -269,7 +269,7 @@ func (r *Runner) RunLibfuzzerAndReport(ctx context.Context, args []string, env [
 				// it has been successfully initialized to provide users with
 				// the context of this abnormal exit even without verbose mode.
 				if !r.Verbose {
-					_, _ = log.NewPTermWriter().Write(reporter.StartupOutput())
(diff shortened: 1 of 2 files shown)


pkg/runner/libfuzzer/libfuzzer_runner.go
```diff
@@ -265,6 +265,12 @@ func (r *Runner) RunLibfuzzerAndReport(ctx context.Context, args []string, env [
 			}
 
 			if !IsExpectedExitError(err) {
+				// Print the stderr output of the fuzzer up to the point where
+				// it has been successfully initialized to provide users with
+				// the context of this abnormal exit even without verbose mode.
+				if !r.Verbose {
+					_, _ = log.NewPTermWriter().Write(reporter.StartupOutput())
```
adombeck, line 272:
```suggestion
					log.Print(reporter.StartupOutput())
```

pkg/runner/libfuzzer/libfuzzer_runner.go
```diff
@@ -265,6 +265,12 @@ func (r *Runner) RunLibfuzzerAndReport(ctx context.Context, args []string, env [
 			}
 
 			if !IsExpectedExitError(err) {
+				// Print the stderr output of the fuzzer up to the point where
+				// it has been successfully initialized to provide users with
+				// the context of this abnormal exit even without verbose mode.
```
adombeck, line 270: In case that the fuzzer was successfully initialized and an unexpected error occurred afterwards, we're only printing the stderr up to the successful initialization, potentially leaving out the actual error message. Maybe we should just print all stderr output when an unexpected error occurs?

fmeum:
@adombeck The new integration test is relatively flaky on macOS. Do you think that we could somehow inject a deterministic seed that is known to work well?

pkg/runner/libfuzzer/libfuzzer_runner.go
fmeum: We could, but I don't really know how this could happen - shouldn't libFuzzer catch all kinds of abnormal exits after it has successfully started? The cost would be to keep a buffer with the entire stderr around, which may get really large really quickly in case the user left in some print statements.

adombeck: Lets leave it as is, we can think about a solution (e.g. a ring buffer) in case that we do see unexpected errors after initialization

adombeck (approved):

adombeck:
> @adombeck The new integration test is relatively flaky on macOS. Do you think that we could somehow inject a deterministic seed that is known to work well?

I created https://github.com/CodeIntelligenceTesting/cifuzz/pull/106 for that, lets see if that helps

PR #101, opened 2022-07-01
run: Use CMake integration info file to find fuzz test executable

fmeum:


internal/build/cmake/cmake.go
```diff
@@ -130,3 +132,30 @@ func (b *Builder) Build(fuzzTest string) error {
...
+	}
+	if err != nil {
+		return "", errors.WithStack(err)
+	}
+	if !exists {
+		return "", errors.Errorf("failed to find executable for fuzz test %q", fuzzTest)
+	}
+	fuzzTestExecutable, err := ioutil.ReadFile(infoFileCandidate)
+	if err != nil {
+		return "", err
```
adombeck, line 158:
```suggestion
		return "", errors.WithStack(err)
```

adombeck (approved):

PR #107, opened 2022-07-06
envutil: Use os.PathListSeparator

adombeck:
The purpose of the AddToColonSeparatedList function is to add paths to
the PATH or LD_LIBRARY_PATH environment variables, which are not
colon-separated on Windows, so we now use os.PathListSeparator instead.

### DIFF
--- a/util/envutil/envutil.go
+++ b/util/envutil/envutil.go
@@ -1,6 +1,7 @@
 package envutil
 
 import (
+	"os"
 	"strings"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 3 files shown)


util/envutil/envutil.go
```diff
@@ -9,11 +10,11 @@ import (
...
-// AddToColonSeparatedList appends a string to another string containing
-// a list of colon-separated strings (like the PATH and LD_LIBRARY_PATH
-// environment variables do). It doesn't add duplicates and removes any
-// empty strings from the list.
-func AddToColonSeparatedList(list string, value ...string) string {
+// AppendPathList appends a string to another string containing a list
+// of paths, separated by os.PathListSeparator (like the PATH and
+// LD_LIBRARY_PATH environment variables). It doesn't add duplicates and
+// removes any empty strings from the list.
+func AppendPathList(list string, value ...string) string {
```
fmeum, line 17: Nit: I would like `AppendToPathList` a bit better.

fmeum (approved):

PR #108, opened 2022-07-06
Makefile: Add rm-unused flag to goimports-reviser

florianGla:


fmeum:
@adombeck The deterministic seed doesn't seem to suffice, the test is still flaky.

fmeum:
@adombeck It is also really strange that the fuzzer gets stuck at "features: 1". There are at least to `if`s that should be trivial to pass.

adombeck:
@fmeum: I created a ticket for the flakiness issue: https://code-intelligence.atlassian.net/browse/CLI-247

florianGla:
removes used imports and shall therefore not be used

PR #110, opened 2022-07-06
Implement cifuzz bundle

fmeum:


internal/build/cmake/cmake.go
```diff
@@ -211,3 +211,73 @@ func (b *Builder) FindFuzzTestExecutable(fuzzTest string) (string, error) {
...
+	)
+	stdout, err := cmd.Output()
+	if err != nil {
+		return nil, errors.WithStack(err)
+	}
+
+	var resolvedDeps []string
+	var unresolvedDeps []string
+	var conflictingDeps []string
+	lines := strings.Split(string(stdout), "\n")
```
jochil, line 230: What do you think about using `bufio.Scanner` (https://pkg.go.dev/bufio#example-Scanner-Lines) ?
As far as I know this is the preferred way reading lines from an `io.Reader` and it should handle different line endings better

internal/build/cmake/cmake.go
```diff
@@ -211,3 +211,73 @@ func (b *Builder) FindFuzzTestExecutable(fuzzTest string) (string, error) {
 	}
 	return string(fuzzTestExecutable), nil
 }
+
+func (b *Builder) GetRuntimeDeps(fuzzTest string) ([]string, error) {
```
jochil, line 215: IMO we should cover this functionality with a unit test, not the `cmake` command invocation but the output parsing.

internal/cmd/bundle/bundle.go
```diff
@@ -0,0 +1,249 @@
...
+	fuzzTest   string
+	outputPath string
+}
+
+func New(conf *config.Config) *cobra.Command {
+	opts := &bundleOpts{}
+	cmd := &cobra.Command{
+		Use:               "bundle",
+		Short:             "Bundles the fuzz target into an archive",
+		Long:              "Bundles all runtime artifacts required by the given fuzz target into a self-contained archive that can be executed by a remote fuzzing worker",
```
jochil, line 57: we should use `fuzz test` instead of `fuzz target`

fmeum:
> I am not sure if we already agreed on it, but I think it would be useful to use some of the `archive` unit tests I already created: https://github.com/CodeIntelligenceTesting/cifuzz/blob/jochil/CLI-217-artifacts/pkg/artifact/archive/archive_test.go

I added tests for the Git and archive functionality. Since we have to ensure that the archives we create can be extracted by anyone, I am relying on host `tar` instead of our own extraction function in the test. Please take a look!

integration-tests/cmake/cmake_test.go
```diff
@@ -257,6 +296,7 @@ func modifyFuzzTestToCallFunction(t *testing.T, fuzzTestPath string) {
 	f, err := os.OpenFile(fuzzTestPath, os.O_RDWR, 0700)
 	require.NoError(t, err)
 	defer f.Close()
+	defer f.Sync()
```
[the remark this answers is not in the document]
fmeum: Good catch, but this commit shouldn't have been there to begin with. Please ignore it, I will drop it before merging.

fmeum:
I dropped the commit and rebased onto main.

jochil (approved):

PR #111, opened 2022-07-07
Call fsync after copying

adombeck:
We've seen some errors in CI pipelines indicating that the files copied
by the installer are being used before they were synced to disk, so we
now sync after copying.

The Copy function from github.com/otiai10/copy already supports syncing
after copying, so we're using that now instead of fileutil.CopyFile.

Note that this might not fix all synchronization issues, quoting the
fsync(2) manual:

    Calling fsync() does not necessarily ensure that the entry in the directory
    containing the file has also reached disk.  For that an explicit fsync() on
    a file descriptor for the directory is also needed.

So in case we still see synchronization issues, we should also call fsync
on the directory of the file we copied.

### DIFF
--- a/tools/install/installer.go
+++ b/tools/install/installer.go
@@ -160,15 +160,15 @@ func (i *installer) InstallMinijail() error {
 	// Install minijail binaries
 	src := filepath.Join(i.projectDir, "third-party/minijail/minijail0")
 	dest := filepath.Join(i.binDir(), "minijail0")
-	err = fileutil.CopyFile(src, dest, 0700)
+	err = copy.Copy(src, dest, copy.Options{Sync: true})
 	if err != nil {
-		return err
+		return errors.WithStack(err)
 	}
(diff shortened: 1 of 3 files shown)


tools/install/installer.go
```diff
@@ -159,14 +159,12 @@ func (i *installer) InstallMinijail() error {
...
 	if err != nil {
-		return err
+		return errors.WithStack(err)
 	}
 	src = filepath.Join(i.projectDir, "third-party/minijail/libminijailpreload.so")
-	dest = filepath.Join(i.libDir(), "libminijailpreload.so")
-	err = fileutil.CopyFile(src, dest, 0600)
+	err = copy.Copy(src, i.libDir(), copy.Options{Sync: true})
 	if err != nil {
 		return err
```
fmeum, line 169:
```suggestion
		return errors.WithStack(err)
```

fmeum (approved):

tools/install/installer.go
adombeck: Good catch, applied

PR #116, opened 2022-07-07
installer: Call fsync on the directory after copying

adombeck:
We still saw synchronization issues since we implemented calling fsync on the file in 65b1930006a819dd56b0243e995fb1244cdd9833.

### DIFF
--- a/util/fileutil/fileutil.go
+++ b/util/fileutil/fileutil.go
@@ -63,6 +63,22 @@ func Cleanup(path string) {
 	}
 }
 
+func Sync(path string) error {
+	f, err := os.Open(path)
+	if err != nil {
+		return errors.WithStack(err)
+	}
+	syncErr := f.Sync()
+	closeErr := f.Close()
+	if syncErr != nil {
+		return errors.WithStack(syncErr)
+	}
+	if closeErr != nil {
+		return errors.WithStack(closeErr)
+	}
+	return nil
+}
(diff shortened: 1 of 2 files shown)


util/fileutil/fileutil.go
```diff
@@ -63,6 +63,22 @@ func Cleanup(path string) {
 	}
 }
 
+func Sync(path string) error {
```
fmeum, line 66: Could you add a comment explaining what it does and when it is needed? That could replace the comments at the call sites.

adombeck:
Looking at the error message again, I don't think this change will fix the issue. From https://github.com/CodeIntelligenceTesting/cifuzz/runs/7234917266?check_suite_focus=true#step:9:162:
```
dlopen(): /tmp/cifuzz-install-dir-1347165277/lib/libminijailpreload.so: file too short
```
If the directory entry was not synced, the file should not be found at all. The only reason I see that could cause the file being incomplete is if the source file was not completely written to disk before we copied it. It's written via the minijail Makefile, so that's out of our control. Maybe the installer should just call `sync(1)` once after installing cifuzz and its depedency (if its running on Linux or MacOS - Windows doesn't seem to ship a command for that).

adombeck:
Closing in favor of https://github.com/CodeIntelligenceTesting/cifuzz/pull/118

PR #119, opened 2022-07-07
flogla/add crashing input to seed corpus

florianGla:


internal/cmd/run/report_handler/report_handler.go
```diff
@@ -120,9 +123,17 @@ func (h *ReportHandler) Handle(r *report.Report) error {
...
+		if runtime.GOOS == "windows" {
+			copyCmd = "copy"
+		}
+
 		log.Print("\n")
 		log.Printf("=========================== Finding %d ===========================", h.numFindings)
 		log.Print(strings.Join(r.Finding.Logs, "\n"))
+		log.Print("\n")
+		log.Print("copy crashing input to your seeds directory with the following command:")
+		log.Printf("%s %s %s/", copyCmd, r.Finding.InputDataPath, h.seedsDir)
```
adombeck, line 136: This should only be printed if `r.Finding.InputDataPath` is non-empty, which is the case UBsan findings. For example when I execute the `parser_fuzz_test` from `tools/cmake/testdata` I get the following output:
```
[...]
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /home/adrian/cifuzz/tools/cmake/testdata/src/parser/parser.cpp:14:11 in 
                                                                                                                                      
copy crashing input to your seeds directory with the following command:                                                               
cp  /home/adrian/cifuzz/tools/cmake/testdata/.cifuzz-corpus/parser_fuzz_test/        
[...]
```

adombeck, line 136: We might want to explain the effect of that. Maybe something like this:

```
To add the crashing input to the seed corpus for regression testing, use:

    cp <input> <seeds-dir>

```

Or maybe we should try to explain it in more detail. @fmeum, @jochil: WDYT?

jochil: I agree... maybe explaining it even more

```
To add the crashing input to the seed corpus for regression testing, use:
    cp <input> <seeds-dir>
After adding, the input will always be used when running the fuzz test via the replayer (eg. inside your IDE/editor)
```

internal/cmd/run/report_handler/report_handler.go
```diff
@@ -120,9 +123,17 @@ func (h *ReportHandler) Handle(r *report.Report) error {
...
+		copyCmd := "cp"
+		if runtime.GOOS == "windows" {
+			copyCmd = "copy"
+		}
+
 		log.Print("\n")
 		log.Printf("=========================== Finding %d ===========================", h.numFindings)
 		log.Print(strings.Join(r.Finding.Logs, "\n"))
+		log.Print("\n")
+		log.Print("copy crashing input to your seeds directory with the following command:")
```
jochil, line 135: I think it would be nice to highlight these kind of tips (what to do next) with `log.Info` and/or print them after the "finding block"... so instead of 
```
copy crashing input to your seeds directory with the following command:                                                                      
cp /tmp/minijail-out/crash-0e45924f90184bde44d54baf19d2f3051073b306 /home/jochen/dev/workspace/ci/cifuzz/examples/cmake/.cifuzz-corpus/my_fuzz_test/
=================================================================  
```
I think this would be better readable
```
=================================================================  
copy crashing input to your seeds directory with the following command:                                                                      
    cp /tmp/minijail-out/crash-0e45924f90184bde44d54baf19d2f3051073b306 /home/jochen/dev/workspace/ci/cifuzz/examples/cmake/.cifuzz-corpus/my_fuzz_test/
```

adombeck: Note that there could be multiple findings. In that case, would you prefer this hint to be printed in-between the finding bocks?

jochil: Or more highlighted... by now it looks more like a part of the finding instead of a tip/hint for the user

README.md
```diff
@@ -57,6 +57,7 @@ cd examples/cmake
 cifuzz run my_fuzz_test
 ```
 This should stop after a few seconds with an actual finding.
+The crashing input file gets stored in the current working directory.
```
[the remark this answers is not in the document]
adombeck: True. Maybe we should change that, i.e. move the input file to the current working directory or make the fuzzer put it there in the first place (the latter would require that we allow write access to the current working directory inside the sandbox).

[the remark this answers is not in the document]
jochil: I agree. Moving it after the fuzzing run sounds a little bit better for me... but giving the fuzzer write access to the working dir should also be no problem. Unless your working dir is `/` ;)

internal/cmd/run/report_handler/report_handler.go
florianGla: Sound goods

README.md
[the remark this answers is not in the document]
florianGla: Maybe it is better to make the fuzzer put it in the working directory directly. Replayer contains the ouput path in its log message and we would have to remove it otherwise.
artifact_prefix='/tmp/minijail-out/'; Test unit written to /tmp/minijail-out/crash-72209f5e3795ee3c9d299c19bb51bce700b1d655

jochil:
@florianGla  for the CLI-272 task I added the crashing input file to the `Finding` struct. Maybe this is helpful or relevant for this change.
you can take a look at it in this branch `jochil/CLI-272-store-crashing-input` (still WIP)

internal/cmd/run/report_handler/report_handler.go
```diff
@@ -136,6 +141,20 @@ func (h *ReportHandler) Handle(r *report.Report) error {
...
+		if r.Finding.InputFile != "" {
+			copyCmd := "cp"
+			if runtime.GOOS == "windows" {
+				copyCmd = "copy"
+			}
+			destDir := fileutil.PrettifyPath(filepath.Join(h.corpusDir, r.Finding.Name))
+
+			log.Print("\n")
+			log.Print("Add the crashing input to the corpus directory with:")
+			log.Infof("  %s %s %s", copyCmd, r.Finding.InputFile, destDir)
```
jochil, line 154: To make sure the directory exists we can change the command to
`mkdir -p destDir && cp ...`
at least for linux and mac

internal/cmd/run/report_handler/report_handler.go
```diff
@@ -136,6 +141,20 @@ func (h *ReportHandler) Handle(r *report.Report) error {
 		log.Print("\n")
 		log.Printf("=========================== Finding %d ===========================", h.numFindings)
 		log.Print(strings.Join(r.Finding.Logs, "\n"))
+
+		if r.Finding.InputFile != "" {
+			copyCmd := "cp"
+			if runtime.GOOS == "windows" {
+				copyCmd = "copy"
+			}
+			destDir := fileutil.PrettifyPath(filepath.Join(h.corpusDir, r.Finding.Name))
```
jochil, line 150: the content of this variable is more of a (file)path than a directory

internal/cmd/run/run.go
```diff
@@ -31,15 +31,16 @@ import (
...
-	dictionary     string
-	engineArgs     []string
-	fuzzTargetArgs []string
-	timeout        time.Duration
-	useSandbox     bool
-	printJSON      bool
+	buildCommand       string
+	fuzzTest           string
+	seedCorpusDirs     []string
+	generatedCorpusDir string
```
jochil, line 37: If this is about the seed corpus this field might not be needed at all

internal/cmd/run/run.go
```diff
@@ -132,7 +132,8 @@ func (c *runCmd) run() error {
 	// Initialize the report handler. Only do this right before we start
 	// the fuzz test, because this is storing a timestamp which is used
 	// to figure out how long the fuzzing run is running.
-	c.reportHandler, err = report_handler.NewReportHandler(c.opts.printJSON, viper.GetBool("verbose"))
+	seedCorpusDir := c.opts.fuzzTest + "_seed_corpus"
```
jochil, line 135: maybe calling this `defaultSeedCorpusDir` would avoid confussion with the `seedCorpusDirs` option from this command

README.md
```diff
@@ -100,6 +100,8 @@ directly executing the replayer binary
 (see [here](docs/How-To-Write-A-Fuzz-Test.md#how-to-buildcompile-your-fuzz-tests)
 on how to build that binary).
 This will use the replayer to apply existing input data from the
-seed corpus. In this case the fuzz test will stop immediately after
+seed corpus, which has to be stored in the directory &lt;fuzz-test-name&gt;_seed_corpus
+under your project root. Note that this directory has to be manually created by the user.
```
jochil, line 104: > under your project root
are you sure about this? I remember that we talked about storing it beside the fuzz test (which can be everywhere in the project)

README.md
```diff
@@ -100,6 +100,8 @@ directly executing the replayer binary
 (see [here](docs/How-To-Write-A-Fuzz-Test.md#how-to-buildcompile-your-fuzz-tests)
 on how to build that binary).
 This will use the replayer to apply existing input data from the
-seed corpus. In this case the fuzz test will stop immediately after
+seed corpus, which has to be stored in the directory &lt;fuzz-test-name&gt;_seed_corpus
+beside your fuzz test. Note that this directory has to be manually created by the user.
```
jochil, line 104:
```suggestion
beside your fuzz test. Note that this directory has to be created manually.
```

And please make sure the line width is <= 72

PR #120, opened 2022-07-08
install: Fix relative install dir

florianGla:


tools/install/installer.go
```diff
@@ -199,6 +204,9 @@ func (i *installer) InstallProcessWrapper() error {
 		compiler = "clang"
 	}
 	dest := filepath.Join(i.libDir(), "process_wrapper")
+	log.Print("------------------")
+	log.Print(dest)
+	log.Print("------------------")
```
adombeck, line 209: What's the purpose of those lines?

florianGla: Sorry, I  just printed the dest for testing purposes. I will update the commit.

adombeck (approved):

PR #121, opened 2022-07-08
Include seed directory in artifact bundles

fmeum:


pkg/artifact/archive.go
```diff
@@ -19,7 +20,14 @@ func WriteArchive(out io.Writer, manifest map[string]string) error {
 	tw := tar.NewWriter(gw)
 	defer tw.Close()
 
-	for archivePath, absPath := range manifest {
+	// Sort the archive paths first so that the generated archive is deterministic - map traversals aren't.
+	var archivePaths []string
+	for archivePath := range manifest {
+		archivePaths = append(archivePaths, archivePath)
+	}
```
jochil, line 30: you can replace this with `archivePaths := maps.Keys(manifest)` if you use https://pkg.go.dev/golang.org/x/exp/maps#Keys (which should become part of the standard in 1.19

jochil (approved):

pkg/artifact/archive.go
fmeum: Thanks, used.

PR #122, opened 2022-07-08
integration-tests: Avoid make not rebuilding fuzz test

adombeck:
Check that the fuzz test source file was actually modified. We are
seeing test failures on MacOS which could be caused by the source file
not being modified.

### DIFF
--- a/integration-tests/cmake/cmake_test.go
+++ b/integration-tests/cmake/cmake_test.go
@@ -297,6 +297,7 @@ func modifyFuzzTestToCallFunction(t *testing.T, fuzzTestPath string) {
 	// At the top of the file we add the required headers
 	lines := []string{`#include "parser.h"`}
 	var seenBeginningOfFuzzTestFunc bool
+	var addedFunctionCall bool
 	for scanner.Scan() {
 		if strings.HasPrefix(scanner.Text(), "FUZZ_TEST(") {
 			seenBeginningOfFuzzTestFunc = true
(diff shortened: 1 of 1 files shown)


adombeck:
@fmeum: Commit [fe04f9a](https://github.com/CodeIntelligenceTesting/cifuzz/pull/122/commits/fe04f9ab85ce86b5f0a31aaa6bf982a39acaba37) seems to fix the failure on MacOS. But the timestamps I'm printing on this branch differ between pre-modification and post-modification of the fuzz test, so I don't understand why make sometimes doesn't rebuild the fuzz test unless we're forcing it to. Do you have an idea how to fix this for good (since we don't want to force rebuilding all targets on each build)?

fmeum:
> @fmeum: Commit [fe04f9a](https://github.com/CodeIntelligenceTesting/cifuzz/pull/122/commits/fe04f9ab85ce86b5f0a31aaa6bf982a39acaba37) seems to fix the failure on MacOS (no failures in 5 runs). But the timestamps I'm printing on this branch differ between pre-modification and post-modification of the fuzz test, so I don't understand why make sometimes doesn't rebuild the fuzz test unless we're forcing it to. Do you have an idea how to fix this for good (since we don't want to force rebuilding all targets on each build)?

Funny, I had thought of something like this before but immediately rejected the idea because "certainly macOS file systems and make got this right"...

I don't know how to fix this - introducing a wait may not be enough if the timestamps are already different and it still doesn't rebuild.

fmeum:
https://www.gnu.org/software/autoconf/manual/autoconf-2.63/html_node/Timestamps-and-Make.html makes it sound as if we should just introduce a 1s wait.

adombeck:
> https://www.gnu.org/software/autoconf/manual/autoconf-2.63/html_node/Timestamps-and-Make.html makes it sound as if we should just introduce a 1s wait.

Done, lets see if this also fixes the issue

fmeum (approved):

adombeck:
The test passed in 5 out of 5 runs. Seems good enough for me, so I'll merge this now.

PR #124, opened 2022-07-08
runner tests: Log to buffer instead modifying os.Stderr

adombeck:
Modifying the global variable os.Stderr caused "broken pipe" errors when
running the tests in parallel.

### DIFF
--- a/pkg/runner/libfuzzer/integration-tests/testutils/runner_testutils.go
+++ b/pkg/runner/libfuzzer/integration-tests/testutils/runner_testutils.go
@@ -49,6 +49,7 @@ type RunnerTest struct {
 	FuzzerEnv       []string
 	DisableMinijail bool
 	RunsLimit       int
+	LogOutput       *bytes.Buffer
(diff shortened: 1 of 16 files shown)


pkg/runner/libfuzzer/integration-tests/testutils/runner_testutils.go
```diff
@@ -108,37 +113,12 @@ func (test *RunnerTest) Start(t *testing.T, reportCh chan *report.Report) error
...
-	go func() {
-		_, _ = io.Copy(io.MultiWriter(origStdout, stdout), rStdout)
-		stdoutClosed <- struct{}{}
-	}()
-
-	go func() {
-		_, _ = io.Copy(io.MultiWriter(origStderr, stderr), rStderr)
-		stderrClosed <- struct{}{}
-	}()
-
```
jochil, line 141: very good that this is gone now :)

jochil (approved):

PR #125, opened 2022-07-08
add gocritic linter

jochil:
and improve existing linter setup by using `golangci-lint` instead of managing single linter

### DIFF
--- a/internal/cmd/init/init_test.go
+++ b/internal/cmd/init/init_test.go
@@ -1,7 +1,6 @@
 package init
 
 import (
-	"io/ioutil"
 	"log"
 	"os"
 	"testing"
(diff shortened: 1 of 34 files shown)


internal/cmd/init/init_test.go
```diff
@@ -16,15 +15,15 @@ var baseTempDir string
...
+	baseTempDir, err = os.MkdirTemp("", "init-cmd-test-")
 	if err != nil {
 		log.Fatalf("Failed to create temp dir for tests: %+v", err)
 	}
 	defer fileutil.Cleanup(baseTempDir)
 
 	err = os.Chdir(baseTempDir)
 	if err != nil {
-		log.Fatalf("Failed to change the working directory to %s", baseTempDir)
+		log.Fatalf("Failed to change the working directory to %s", baseTempDir) //nolint:gocritic
```
fmeum, line 26: If this is about defer not running, shouldn't we invoke the cleanup manually?

internal/cmd/root/root.go
```diff
@@ -67,11 +67,11 @@ func New() *cobra.Command {
 
 	rootCmd.PersistentFlags().BoolP("verbose", "v", false,
 		"Show more verbose output, can be helpful for debugging problems")
-	viper.BindPFlag("verbose", rootCmd.PersistentFlags().Lookup("verbose"))
+	viper.BindPFlag("verbose", rootCmd.PersistentFlags().Lookup("verbose")) //nolint: errcheck
```
fmeum, line 70: A brief comment explaining why we need to disable this check would be great here.

fmeum (approved):

internal/cmd/init/init_test.go
jochil: of course... changed it

internal/cmd/root/root.go
jochil: I just added the error handling ;)

PR #127, opened 2022-07-08
test: run unit tests also with a zsh to avoid regression

jochil:


adombeck (approved):
Looks good to me, now we just have to get the tests to pass for zsh

PR #128, opened 2022-07-11
dialog: Read zsh input from stdin

florianGla:


adombeck:
On this branch, I don't get any tab completion for the filename input when using zsh

adombeck:
@florianGla: I still don't get any tab completion for the filename input when using zsh

PR #129, opened 2022-07-11
Revert ineffective integration test fixes

fmeum:
This reverts the following commits as the root cause of the integration
test failures turned out to be something else:
* 373f2de1d840916ed7f1c3fc11d29e07b0dbe1e9
* dcc5fd2ab62b3476f9505f94e21cac619978ba9e

### DIFF
--- a/tools/install/installer.go
+++ b/tools/install/installer.go
@@ -114,19 +114,6 @@ func (i *installer) Cleanup() {
 	fileutil.Cleanup(i.InstallDir)
 }
 
-// Sync executes sync(1) to flush the disk cache
-func (i *installer) Sync() error {
-	// Windows doesn't come with a command like sync(1)
-	if runtime.GOOS == "windows" {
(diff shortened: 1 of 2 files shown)


tools/install/installer.go
```diff
@@ -227,8 +207,7 @@ func (i *installer) InstallCIFuzz() error {
 	if err != nil {
 		return errors.WithStack(err)
```
fmeum, line 208: @adombeck Just FYI because I learned it recently: This could just be `return errors.WithStack(err)` without the guarding if.

fmeum:
Missed that one commit is intended to fix a seperate integrationt test failure.

tools/install/installer.go
adombeck: Thanks, I'm aware, I do use that sometimes in very small functions but in larger functions my personal preference is to use the if.

PR #134, opened 2022-07-12
cmake: Print stderr on exec.Cmd failure

fmeum:
The error handling logic is contained in a new executil helper function.

### DIFF
--- a/util/executil/executil.go
+++ b/util/executil/executil.go
@@ -2,9 +2,11 @@ package executil
 
 import (
 	"context"
+	"fmt"
 	"io"
 	"os"
 	"os/exec"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


util/executil/executil.go
```diff
@@ -21,6 +21,21 @@ const (
...
+// stderr to the message if the error is an exec.ExitError.
+func HandleExecError(cmd *exec.Cmd, err error) error {
+	if exitErr, ok := err.(*exec.ExitError); ok {
+		return errors.Wrapf(
+			exitErr,
+			"Failed with exit code %d: %s\n\n%s\n",
+			exitErr.ExitCode(),
+			cmd.String(),
+			string(exitErr.Stderr),
+		)
```
adombeck, line 34: The `exitErr` message is usually something like `exit status 1`, in which case the `Failed with exit code 1` is duplicate information.

`string(exitErr.Stderr)` is empty if `cmd.Stderr` was non-nil, resulting in just 3 newlines being appended to the error message.

The following change solves both issues (inspired by https://github.com/cli/cli/blob/0e3c3bb4a428eff7fcff472a34e5208561a9406c/internal/run/run.go#L69-L75) but doesn't print the full command. If you think printing the full command is helpful, feel free to change that.

```suggestion
		stderr := string(exitErr.Stderr)
		if stderr != "" && !strings.HasSuffix(stderr, "\n") {
			stderr += "\n"
		}
		err = fmt.Errorf("%s%s: %w", stderr, cmd.Args[0], err)
		return errors.WithStack(err)
```

fmeum: Thanks!

PR #137, opened 2022-07-13
bundle: Allow bundling multiple or all fuzz tests

fmeum:
Since listing fuzz tests requires running configure, the BuildWithCMake
do-it-all method has been removed in favor of having all command run
the builder methods individually. This duplicates hard-coded values for
now, but these have to go away soon anyway.

### DIFF
--- a/internal/build/cmake/cmake.go
+++ b/internal/build/cmake/cmake.go
@@ -14,7 +14,6 @@ import (
 	"github.com/pkg/errors"
 
 	"code-intelligence.com/cifuzz/internal/build"
-	"code-intelligence.com/cifuzz/internal/config"
 	"code-intelligence.com/cifuzz/pkg/log"
 	"code-intelligence.com/cifuzz/util/executil"
 	"code-intelligence.com/cifuzz/util/fileutil"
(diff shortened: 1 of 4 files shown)


jochil:
> This duplicates hard-coded values for now, but these have to go away soon anyway.
It duplicates also the build cmd invocation and error handling. Why not keeping a function encapsulating this?

internal/build/cmake/cmake.go
```diff
@@ -202,6 +163,28 @@ func (b *Builder) FindFuzzTestSeedCorpus(fuzzTest string) (string, error) {
 	return b.readInfoFile(fuzzTest, "seed_corpus")
 }
 
+// ListFuzzTests lists all fuzz tests defined in the CMake project after
+// Configure has been run.
+func (b *Builder) ListFuzzTests() ([]string, error) {
```
jochil, line 168: There is some overlapping functionality with `https://github.com/CodeIntelligenceTesting/cifuzz/blob/main/internal/completion/fuzz_test_completion.go#L52`. Maybe we can generalize that a little bit

fmeum: I thought about this for a while and decided not to as the requirements are different: shell completion is best-effort and conceptually shouldn't be coupled with the `cmake.Builder` type, whereas `ListFuzzTests` has to be precise and can depend on a `Builder` instance.

jochil:
> I removed the function because the existing two callers diverge with this PR: `bundle` has to list the fuzz tests between `Configure()` and `Build(...)`, `run` doesn't - meaning that the leanest abstraction that remains is just to call these functions individually. The error handling is just standard `if err != nil { return err }`.

Ok, but most of the stuff `BuildWithCMake` encapsulated was not the Build/Configure calls... it was the creation of the Builder with the default (and duplicated) values. I meant to put this in one place, maybe `New(Default)CmakeBuilder`

internal/build/cmake/cmake.go
fmeum: I refactored the `cmake` package so that the paths are only constructed once there. I don't see an easy way to reuse that logic in the command-line completion logic though: It requires knowing the precise build directory, which just isn't available at that point.

fmeum:
> Ok, but most of the stuff `BuildWithCMake` encapsulated was not the Build/Configure calls... it was the creation of the Builder with the default (and duplicated) values. I meant to put this in one place, maybe `New(Default)CmakeBuilder`

That would only help to some extent: We already duplicate the hardcoded values in the build method for the unknown build system. As we can't avoid duplication entirely that way and would create an abstraction that is only useful while we keep the hardcoded values around, I would rather not have it: The pain of the hardcoded values should be in plain sight from the commands and we should replace them with values extracted from the config ASAP. That should be possible pretty quickly after @adombeck has figured out the config story for the sandbox flag.

jochil:
> The pain of the hardcoded values should be in plain sight from the commands and we should replace them with values extracted from the config ASAP.

That is a very good reason... I can agree on this

jochil (approved):

PR #138, opened 2022-07-13
executil: Do not print a stack trace for ExitErrors

fmeum:
ExitError's are commonly encountered by users and already contain
enough context for debugging purposes that we don't need to emit stack
traces for them.

### DIFF
--- a/util/executil/executil.go
+++ b/util/executil/executil.go
@@ -35,7 +35,7 @@ func HandleExecError(cmd *exec.Cmd, err error) error {
 			stderr += "\n"
 		}
 		err = fmt.Errorf("%s%s: %w", stderr, cmd.Args[0], err)
-		log.Error(err)
+		log.Error(err, err.Error())
 		return cmdutils.WrapSilentError(errors.WithStack(err))
 	}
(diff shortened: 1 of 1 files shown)


fmeum:
CC @florianGla

util/executil/executil.go
```diff
@@ -25,14 +26,17 @@ const (
...
+// returned as a cmdutils.SilentError.
 func HandleExecError(cmd *exec.Cmd, err error) error {
 	if exitErr, ok := err.(*exec.ExitError); ok {
 		stderr := string(exitErr.Stderr)
 		if stderr != "" && !strings.HasSuffix(stderr, "\n") {
 			stderr += "\n"
 		}
 		err = fmt.Errorf("%s%s: %w", stderr, cmd.Args[0], err)
-		return errors.WithStack(err)
+		log.Error(err)
```
adombeck, line 38: Our `log.Error` function takes the error as the first argument and the message it should print as the second. If no message is provided, it prints an empty string (we should probably change that).
```suggestion
		log.Error(err, err.Error())
```

adombeck (approved):

PR #139, opened 2022-07-13
cmake: Fix for projects that don't enable the C language 

fmeum:
Previously, projects with `project(MyProject LANGUAGES CXX)` in their
top-level CMakeLists.txt would fail to build the replayer.

This is fixed by marking replayer.c as a C++ source file if C is not
in ENABLED_LANGUAGES.

### DIFF
--- a/internal/cmd/init/init.go
+++ b/internal/cmd/init/init.go
@@ -69,7 +69,8 @@ func printBuildSystemInstructions(cwd string) {
 	}
 	if cfg.BuildSystem == config.BuildSystemCMake {
 		log.Infof(`
-Enable fuzz testing in your CMake project by adding the following lines to the top-level CMakeLists.txt:
+Enable fuzz testing in your CMake project by adding the following lines to the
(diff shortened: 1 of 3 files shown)


internal/cmd/init/init.go
```diff
@@ -69,7 +69,7 @@ func printBuildSystemInstructions(cwd string) {
 	}
 	if cfg.BuildSystem == config.BuildSystemCMake {
 		log.Infof(`
-Enable fuzz testing in your CMake project by adding the following lines to the top-level CMakeLists.txt:
+Enable fuzz testing in your CMake project by adding the following lines to the top-level CMakeLists.txt before any add_subdirectory(...) calls:
```
adombeck, line 72: This line is getting a bit long. How about we wrap it at 80 characters?

fmeum: My thought was that users should already have their terminal set to a width they can comfortably read, in which case long lines would break in a way matching their preference. What do you think?

jochil (approved):

PR #143, opened 2022-07-14
dialog: Don't ask for file path in dialogs

adombeck:
Use a default path instead. This reduces unnecessary complexity,
including having to test the shell completion support with different
shells. We also found that there is no popular tool which asks for the
path where it creates a file in a dialog.


jochil (approved):
All this deleted code... I like it

PR #144, opened 2022-07-14
store crashing input to `.cifuzz-findings`

jochil:
relative to the current working directory and with a generated name based on the crashing input. 
And also save the json information of the finding in the same directory


internal/moby/namesgenerator/names-generator.go
```diff
@@ -0,0 +1,863 @@
+// Package namesgenerator generates random names.
```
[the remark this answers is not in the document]
jochil: do you think it is sufficient to add the copyright notice to the top of the files (as it also mentions the license)?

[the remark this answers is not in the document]
adombeck: I think I would just copy the `NOTICE` and `LICENSE` files into the `moby` directory

internal/moby/namesgenerator/names-generator.go
```diff
@@ -0,0 +1,876 @@
...
+		goto begin
+	}
+
+	if retry > 0 {
+		name += strconv.Itoa(rand.Intn(10)) //nolint:gosec // G404: Use of weak random number generator (math/rand instead of crypto/rand)
+	}
+	return name
+}
+
+func GetDeterminsticName(seed []byte) string {
```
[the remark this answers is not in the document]
jochil: done

pkg/report/finding.go
```diff
@@ -0,0 +1,95 @@
...
+		return errors.WithStack(err)
+	}
+
+	return nil
+}
+
+// move the input file to a new location and update
+// the finding and the logs
+func (f *Finding) moveInputFile(findingDir string) error {
+	if f.InputFile != "" {
```
adombeck, line 76: I would move that check to the `Save` function, i.e. only call `moveInputFile` if `f.InputFile != ""`.

pkg/report/finding.go
```diff
@@ -0,0 +1,95 @@
...
+
+	"code-intelligence.com/cifuzz/pkg/log"
+)
+
+const nameCrashingInput = "crashing-input"
+const nameJsonFile = "finding.json"
+const nameFindingDir = ".cifuzz-findings"
+
+type Finding struct {
+	Name               string        `json:"name,omitempty"`
```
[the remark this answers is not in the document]
jochil: it get set in the `report_handler`

pkg/report/finding.go
jochil: done

pkg/report/finding.go
```diff
@@ -0,0 +1,95 @@
...
+	return nil
+}
+
+// move the input file to a new location and update
+// the finding and the logs
+func (f *Finding) moveInputFile(findingDir string) error {
+	if f.InputFile != "" {
+		newPath := filepath.Join(findingDir, nameCrashingInput)
+
+		// because of minijail we can not use os.Rename for moving the file
```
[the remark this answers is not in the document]
jochil: done

pkg/report/finding.go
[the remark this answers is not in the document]
adombeck: Strange, `intellij` doesn't show that reference

pkg/report/finding.go
```diff
@@ -0,0 +1,93 @@
...
+
+	if err := f.saveJson(findingDir); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (f *Finding) saveJson(findingDir string) error {
+	bytes, err := prettyjson.Marshal(f)
```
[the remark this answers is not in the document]
jochil: ups... done

adombeck:
>  feat(names): add function for determinstic name generation 

The commit message contains the same typo as the function name

adombeck (approved):

internal/moby/namesgenerator/names-generator.go
```diff
@@ -861,3 +862,15 @@ begin:
 	}
 	return name
 }
+
+func GetDeterminsticName(seed []byte) string {
+	// make sure the seed is at least 8 bytes long
+	if len(seed) < 8 {
```
fmeum, line 868: I would feel more comfortable if we used https://pkg.go.dev/crypto/sha256#Sum256 to properly hash the seed and then use the hash to feed the RNG. That should avoid collisions very well. With the current scheme, it looks to me like only the first 8 bytes of the seed are used.

fmeum (changes requested):

adombeck:
> I may be out of the loop, but why do we have to vendor the name generator?

I suggested this to avoid pulling in all of docker as a dependency just to be able to use this small package.

util/testutil/testutil.go
```diff
@@ -26,3 +29,21 @@ func RegisterTestDeps(path ...string) {
 		}
 	}
 }
+
+// PrepareTestTempDir creates and changes the working directory to new tmp dir
+func PrepareTestTempDir(prefix string) string {
```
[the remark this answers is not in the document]
jochil: good point. done

internal/moby/namesgenerator/names-generator.go
jochil: > Seed values that have the same remainder when divided by 2³¹-1 generate the same pseudo-random sequence.

This is what the go docs say about that. I don't see how using a hash should reduce the **probability** of a (natural) collision... yes, it would be harder to produce one by intent, but this is not a problem in our scenario.

And in our case the seed is already a hash generated from the crashing input

> With the current scheme, it looks to me like only the first 8 bytes of the seed are used.

Thats correct, but because the `Source` takes an `int64`. I did not find a way to get around this... and to be honest I am not sure if this is even necessary.

adombeck (approved):

internal/moby/namesgenerator/names-generator.go
fmeum: > And in our case the seed is already a hash generated from the crashing input

If that's the case, then my comment was misguided. I expected `seed []byte` to be the *contents* of a seed file, which aren't pseudo-random. But I now see that you are already passing a hash into the function - may I request we rename the parameter to `hash` or `digest`?

fmeum:
> > I may be out of the loop, but why do we have to vendor the name generator?
> 
> I suggested this to avoid pulling in all of docker as a dependency just to be able to use this small package.

This should no longer be an issue with Go 1.18, which loads modules lazily and prunes the module graph to include only the modules transitively required to build the current main module.

I just did `go get github.com/docker/docker` and then used the `namesgenerator` package somewhere. I didn't experience any slowness and the diff is very manageable:

```
diff --git a/go.mod b/go.mod
index ae6af6a..7470ff2 100644
--- a/go.mod
+++ b/go.mod
@@ -23,6 +23,7 @@ require (
 	github.com/containerd/console v1.0.3 // indirect
 	github.com/cpuguy83/go-md2man/v2 v2.0.2 // indirect
 	github.com/davecgh/go-spew v1.1.1 // indirect
+	github.com/docker/docker v20.10.17+incompatible // indirect
 	github.com/lithammer/fuzzysearch v1.1.5 // indirect
 	github.com/mattn/go-colorable v0.1.12 // indirect
 	github.com/mattn/go-isatty v0.0.14 // indirect
diff --git a/go.sum b/go.sum
index 0e5b948..dcb055e 100644
--- a/go.sum
+++ b/go.sum
@@ -69,6 +69,8 @@ github.com/cpuguy83/go-md2man/v2 v2.0.2/go.mod h1:tgQtvFlXSQOSOSIRvRPT7W67SCa46t
 github.com/davecgh/go-spew v1.1.0/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
 github.com/davecgh/go-spew v1.1.1 h1:vj9j/u1bqnvCEfJOwUhtlOARqs3+rkHYY13jYWTU97c=
 github.com/davecgh/go-spew v1.1.1/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
+github.com/docker/docker v20.10.17+incompatible h1:JYCuMrWaVNophQTOrMMoSwudOVEfcegoZZrleKc1xwE=
+github.com/docker/docker v20.10.17+incompatible/go.mod h1:eEKB0N0r5NX/I1kEveEz05bcu8tLC/8azJZsviup8Sk=
 github.com/envoyproxy/go-control-plane v0.9.0/go.mod h1:YTl/9mNaCwkRvm6d1a2C3ymFceY/DCBVvsKhRF0iEA4=
 github.com/envoyproxy/go-control-plane v0.9.1-0.20191026205805-5f8ba28d4473/go.mod h1:YTl/9mNaCwkRvm6d1a2C3ymFceY/DCBVvsKhRF0iEA4=
 github.com/envoyproxy/go-control-plane v0.9.4/go.mod h1:6rpuAdCZL397s3pYoYcLgu1mIlRU8Am5FuJP05cCM98=
diff --git a/internal/cmd/create/create.go b/internal/cmd/create/create.go
index d267d24..35a599a 100644
--- a/internal/cmd/create/create.go
+++ b/internal/cmd/create/create.go
@@ -6,6 +6,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/docker/docker/pkg/namesgenerator"
 	"github.com/spf13/cobra"
 	"golang.org/x/exp/maps"
 
@@ -48,6 +49,7 @@ func New(config *config.Config) *cobra.Command {
 }
 
 func run(cmd *cobra.Command, args []string, opts *cmdOpts) (err error) {
+	namesgenerator.GetRandomName(5)
 	// get test type
 	opts.testType, err = getTestType(args, cmd.InOrStdin())
 	if err != nil {
```

Unless you see any other potential issues with this, I would prefer not to vendor.

internal/moby/namesgenerator/names-generator.go
jochil: I first named the parameter `hash` but I found it a bit misleading. For me `seed` indicates that if I give in the same parameter value I will get the same result... imo it is a seed. Maybe another name with the same semantic meaning or adding a comment making it clear?

fmeum: Yes, I can see why you would prefer `seed` in this context, it's just unfortunate that the term is overloaded in the context of fuzzing. But as I said above, having a comment that states the expected properties of `seed` as well as the fact that only the first 8 bytes of it are used would be sufficient IMO.

adombeck:
> Unless you see any other potential issues with this, I would prefer not to vendor.

Ok I'm fine with that

internal/moby/namesgenerator/names-generator.go
jochil: I will add a comment but names are important ;) what's about `init`, `unique`, `ident`, `initValue`, `base`

fmeum:
> I thought the same first but now I added functionality to the package and unfortunately the name lists from the package are not public. And the only accessible method will not fulfill our requirements of having the same name for the same crashing input (if this is still a requirement). I can work around this with //go:linkname but it is a tradeoff

If all it takes is `//go:linkname` and `import _ "unsafe"`, I would actually prefer that to vendoring. But I can understand if any of you don't.

internal/moby/namesgenerator/names-generator.go
jochil: done

jochil:
> If all it takes is `//go:linkname` and `import _ "unsafe"`, I would actually prefer that to vendoring. But I can understand if any of you don't.

I tried and it looks okay for me... what do you think @fmeum ?

fmeum:
> > If all it takes is `//go:linkname` and `import _ "unsafe"`, I would actually prefer that to vendoring. But I can understand if any of you don't.
> 
> I tried and it looks okay for me... what do you think @fmeum ?

Looks good to me, @adombeck, are you fine with this hack as well?

fmeum (approved):

adombeck:
> Looks good to me, @adombeck, are you fine with this hack as well?

Yes, works for me

PR #145, opened 2022-07-14
bundle: Add a coverage build to the archive

fmeum:


internal/cmd/bundle/bundle.go
```diff
@@ -66,6 +66,18 @@ type bundleOpts struct {
...
+	Engine     string
+	Sanitizers []string
+}
+
+// Nothing stops users from including fuzz tests only conditionally, so we maintain the list of fuzz tests per build
+// directory. We don't expect this to happen, but it decouples the code a bit better.
+type builderAndFuzzTests struct {
+	FuzzTests []string
+	Builder   *cmake.Builder
+}
```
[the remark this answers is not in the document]
adombeck: > Nothing stops users from including fuzz tests only conditionally

What do you mean with that?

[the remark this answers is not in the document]
fmeum: I extended the comment and added an example, let me know whether it's still unclear.

internal/cmd/bundle/bundle.go
```diff
@@ -66,6 +66,19 @@ type bundleOpts struct {
...
+	Engine     string
+	Sanitizers []string
+}
+
+// Nothing stops users from including fuzz tests only conditionally by e.g. wrapping their `add_fuzz_test` target in
+// CMake into an `if(address IN_LIST CIFUZZ_SANITIZERS)`, so we maintain the list of fuzz tests per build configuration.
+// We don't expect this to happen, but it also decouples the code a bit better.
+type builderAndFuzzTests struct {
+	FuzzTests []string
+	Builder   *cmake.Builder
```
adombeck, line 79: I'm not saying we have to do this now, but I think we could use some refactoring which separates the build result (which is what we use here and pass to `assembleArtifacts` from the `builder` struct. The `bundle.Builder` interface is actually only used to access the build configuration and results.

adombeck (approved):

internal/cmd/bundle/bundle.go
fmeum: I generally agree with this, but getting the build result still requires invoking `cmake`. I would prefer to delay this until we have another build system integrated and can think about the right level of abstraction.

internal/cmd/bundle/bundle.go
```diff
@@ -337,20 +402,31 @@ depsLoop:
...
+		Target:       fuzzTest,
+		Path:         fuzzTestArchivePath,
+		BuildDir:     builder.BuildDir(),
+		Seeds:        archiveSeedsDir,
+		LibraryPaths: externalLibrariesPrefix,
+	}
+
+	if builder.Opts().Engine == "coverage" {
+		fuzzer := baseFuzzerInfo
+		fuzzer.Engine = "LLVM_COV"
```
fmeum, line 415: @adombeck Do you know whether we should bundle the full generated corpus instead of only the seed corpus here?

adombeck: `ci-build` only bundles the user-provided seed corpus. It could of course make the remote fuzzing run more efficient if we bundle the full generated corpus. I don't know how large the generated corpus gets for long-running fuzz tests. If they don't become too large, it might make sense to bundle them. Else we could add a flag for that.

adombeck (approved):

PR #146, opened 2022-07-14
Use viper to read cifuzz.yaml

adombeck:
This uses viper to read the config file and set the command options
according to the settings in the config file. The commit also merges
the old project config into the command options struct, because the only
setting of the project config, the build system setting, can just as
well be a command option (and should also be configurable via a
command-line flag - that will be done in a follow-up commit).

Using viper allows us to handle configuration with very little
self-maintained code. What motivated me to make this change is the
viper.SetDefault() function, which allows us to set a default value
which is applied when neither the config file setting nor the flag is set,
a functionality that we would have reimplement otherwise.

### DIFF
--- a/internal/cmd/root/root.go
+++ b/internal/cmd/root/root.go
@@ -82,7 +82,7 @@ func New() (*cobra.Command, error) {
 
 	rootCmd.AddCommand(initCmd.New())
 	rootCmd.AddCommand(createCmd.New(cmdConfig))
-	rootCmd.AddCommand(runCmd.New(cmdConfig))
+	rootCmd.AddCommand(runCmd.New())
 	rootCmd.AddCommand(reloadCmd.New())
 	rootCmd.AddCommand(bundleCmd.New(cmdConfig))
 
--- a/internal/cmd/run/run.go
+++ b/internal/cmd/run/run.go
@@ -36,7 +36,7 @@ type runOptions struct {
 	seedCorpusDirs []string
 	dictionary     string
 	engineArgs     []string
-	fuzzTargetArgs []string
+	fuzzTestArgs   []string
 	timeout        time.Duration
 	useSandbox     bool
 	printJSON      bool
@@ -112,7 +112,7 @@ func New(config *config.Config) *cobra.Command {
 	cmd.Flags().StringArrayVarP(&opts.seedCorpusDirs, "seed-corpus", "s", nil, "Directory containing sample inputs for the code under test.\nSee https://llvm.org/docs/LibFuzzer.html#corpus and\nhttps://aflplus.plus/docs/fuzzing_in_depth/#a-collecting-inputs.")
 	cmd.Flags().StringVar(&opts.dictionary, "dict", "", "A file containing input language keywords or other interesting byte sequences.\nSee https://llvm.org/docs/LibFuzzer.html#dictionaries and\nhttps://github.com/AFLplusplus/AFLplusplus/blob/stable/dictionaries/README.md.")
 	cmd.Flags().StringArrayVar(&opts.engineArgs, "engine-arg", nil, "Command-line argument to pass to the fuzzing engine.\nSee https://llvm.org/docs/LibFuzzer.html#options and\nhttps://www.mankier.com/8/afl-fuzz.")
-	cmd.Flags().StringArrayVar(&opts.fuzzTargetArgs, "fuzz-test-arg", nil, "Command-line argument to pass to the fuzz test.")
+	cmd.Flags().StringArrayVar(&opts.fuzzTestArgs, "fuzz-test-arg", nil, "Command-line argument to pass to the fuzz test.")
 	cmd.Flags().DurationVar(&opts.timeout, "timeout", 0, "Maximum time in seconds to run the fuzz test. The default is to run indefinitely.")
 	useMinijailDefault := runtime.GOOS == "linux"
 	cmd.Flags().BoolVar(&opts.useSandbox, "use-sandbox", useMinijailDefault, "By default, fuzz tests are executed in a sandbox to prevent accidental damage to the system.\nUse --sandbox=false to run the fuzz test unsandboxed.\nOnly supported on Linux.")
@@ -244,7 +244,7 @@ func (c *runCmd) runFuzzTest(fuzzTestExecutable string) error {
 		SeedCorpusDirs:     c.opts.seedCorpusDirs,
 		Dictionary:         c.opts.dictionary,
 		EngineArgs:         c.opts.engineArgs,
-		FuzzTargetArgs:     c.opts.fuzzTargetArgs,
+		FuzzTestArgs:       c.opts.fuzzTestArgs,
 		ReportHandler:      c.reportHandler,
 		Timeout:            c.opts.timeout,
 		UseMinijail:        c.opts.useSandbox,
--- a/internal/cmd/run/run.go
+++ b/internal/cmd/run/run.go
@@ -115,7 +115,7 @@ func New(config *config.Config) *cobra.Command {
 	cmd.Flags().StringArrayVar(&opts.fuzzTargetArgs, "fuzz-test-arg", nil, "Command-line argument to pass to the fuzz test.")
 	cmd.Flags().DurationVar(&opts.timeout, "timeout", 0, "Maximum time in seconds to run the fuzz test. The default is to run indefinitely.")
 	useMinijailDefault := runtime.GOOS == "linux"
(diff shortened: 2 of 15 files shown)


adombeck:
@fmeum @jochil: I'd like some early feedback on the changes I made to `internal/cmd/run/run.go` and the approach to use `viper` to parse the `cifuzz.yaml` into the command options struct. If you agree with those changes, I will make the same changes to the other commands and get rid of `internal/config/config.go` and `internal/config/project.go`.

internal/cmd/run/run.go
```diff
@@ -115,7 +115,7 @@ func New(config *config.Config) *cobra.Command {
 	cmd.Flags().StringArrayVar(&opts.fuzzTargetArgs, "fuzz-test-arg", nil, "Command-line argument to pass to the fuzz test.")
 	cmd.Flags().DurationVar(&opts.timeout, "timeout", 0, "Maximum time in seconds to run the fuzz test. The default is to run indefinitely.")
 	useMinijailDefault := runtime.GOOS == "linux"
-	cmd.Flags().BoolVar(&opts.useSandbox, "sandbox", useMinijailDefault, "By default, fuzz tests are executed in a sandbox to prevent accidental damage to the system.\nUse --sandbox=false to run the fuzz test unsandboxed.\nOnly supported on Linux.")
+	cmd.Flags().BoolVar(&opts.useSandbox, "use-sandbox", useMinijailDefault, "By default, fuzz tests are executed in a sandbox to prevent accidental damage to the system.\nUse --sandbox=false to run the fuzz test unsandboxed.\nOnly supported on Linux.")
```
jochil: what's about flipping the bool and use `--no-sandbox` 
Would be shorter and more intuitive for the user to deactivate the sandbox, especially if `sandbox=true` is the default.
`cifuzz --no-sandbox run ...` instead of `cifuzz --use-sandbox=true run ...`

internal/cmd/root/root.go
```diff
@@ -109,6 +109,7 @@ func Execute() {
 		var usageErr *cmdutils.IncorrectUsageError
 		if errors.As(err, &usageErr) ||
 			strings.HasPrefix(err.Error(), "required flag") ||
+			strings.HasPrefix(err.Error(), "unknown flag") ||
```
jochil, line 112: I am not happy with this approach... there are more errors from pflag we are not handling right now and it just feels wrong doing that by hand. I have no solution for now... but we should think about it.

see also https://github.com/spf13/pflag/blob/master/flag.go for example: unknown shorthand flag, flag needs an argument, ...

internal/cmd/run/run.go
adombeck: > especially if sandbox=true is the default.

That's only the default on Linux, on Windows and MacOS the default is `false`. And its possible to set one value in the `cifuzz.yaml` and override it via the command-line flag. For this flag, I actually agree that `--no-sandbox` would be the better choice and we don't really need a way to enable the sandbox when its disabled - on Windows or MacOS, it will fail anyway. But it would make the flag inconsistent with other boolean flags we support and might support in the future, because viper and pflag don't have support for `--no-[flag]` flags for boolean values. We could try to build that ourselves, I'm not sure if it's worth the effort though.

But it doesn't make sense to use the sandbox on these platforms because it will fail, so

internal/cmd/root/root.go
adombeck: I'm also not happy with that but I think there is no other solution except to patch the `pflag` package. Note that `gh` does the same: https://github.com/cli/cli/blob/0e3c3bb4a428eff7fcff472a34e5208561a9406c/cmd/gh/main.go#L280-L280

internal/cmd/run/run.go
adombeck: I think ideal would be the flags: `--sandbox, --no-sandbox` and the `cifuzz.yaml` setting `use-sandbox: [true|false]`. I could try to achieve that by parsing cobra flags and setting pflag values before `viper.Get` is used to access the values.

adombeck: > I think ideal would be the flags --sandbox, --no-sandbox

Avoiding duplication in the command usage message would be another problem to solve

internal/cmd/root/root.go
jochil: > Note that gh does the same

and they even handle less errors than us...or?
Maybe it is ok for now if we try to have a complete list

internal/cmd/run/run.go
jochil: > Oh and if we call the flag --no-sandbox, then the corresponding option in the cifuzz.yaml would be no-sandbox: true, which is also not ideal..

would be ok for me... but `disable-sandbox: true` might be better.

As no one will ever call `--use-sandbox=true` having it this way is strange. And IMO a intuitive consistent user interface is more important than having all bool values named as "true"... I don't even see any real value in it

internal/cmd/root/root.go
adombeck: > and they even handle less errors than us...or?

That made me curious and I found how `gh` solves handling the flag errors:
https://github.com/cli/cli/blob/0e3c3bb4a428eff7fcff472a34e5208561a9406c/pkg/cmd/root/root.go#L67-L67
https://github.com/cli/cli/blob/0e3c3bb4a428eff7fcff472a34e5208561a9406c/pkg/cmd/root/help.go#L37-L37

I created PR to do the same: https://github.com/CodeIntelligenceTesting/cifuzz/pull/147

internal/cmd/run/run.go
adombeck: >  And IMO a intuitive consistent user interface is more important than having all bool values named as "true"

Having an intuitive consistent user interface is exactly the reason why I want our boolean flags to be consistent. I don't want some of them to be of the form `--[no-]flag` and some of the form `--flag=[true|false]`.

jochil: I spend a little bit more thoughts on it... Maybe the way with multiple params is the best way, like you said.
We can use https://github.com/spf13/pflag#hidden-flags to make it less confusing for the user

internal/cmd/root/root.go
jochil: nice

internal/cmd/run/run.go
fmeum: I personally don't find `no-sandbox: true` that bad. It's mildly unclean, but having flags differ between CLI and YAML seems worse to me.

internal/cmd/run/run.go
```diff
@@ -31,58 +31,74 @@ import (
...
-	buildCommand   string
-	fuzzTest       string
-	seedCorpusDirs []string
-	dictionary     string
-	engineArgs     []string
-	fuzzTestArgs   []string
-	timeout        time.Duration
-	useSandbox     bool
-	printJSON      bool
+	BuildSystem    string        `mapstructure:"build-system"`
```
fmeum, line 34: Our CLI arguments are arranged in a hierarchy by command, but I assume our YAML settings would be flat? So users would specify `timeout` and have it apply to `cifuzz run` only, but if they specify `build-system`, it will be used by all commands?

fmeum (commented):
I generally like this approach and think that we should adopt it for all commands.

internal/cmd/run/run.go
adombeck: Yes, that's how it would work. Now that you mention it, I see that this could potentially lead to problems if multiple commands have a flag of the same name which the user doesn't want to set to the same value. If we ever encounter this issue, we can consider supporting a nested config as well, e.g.:

```
timeout: 5s  # Used for all commands with a timeout flag
run:
  timeout: 20s # Used only for the run command and takes precedence over the above timeout setting
```

internal/cmd/run/run.go
adombeck: I don't think CLI flag names and YAML setting names differing would be an issue, because we document all our config file settings in the template, so we can make it clear which flags override the setting. That's why I would prefer the flags `--sandbox, --no-sandbox` and the `cifuzz.yaml` setting `use-sandbox: [true|false]`. But because getting this to work requires some effort, I would like to keep the `--use-sandbox` flag for now and try to make the change to `--sandbox, --no-sandbox` later.

jochil: sounds good to me

jochil:
> I generally like this approach and think that we should adopt it for all commands.

I agree. But we definitely should mention this behavior/feature in the documentation.

What is about making the options available as environment variables via viper, too?

jochil:
lgtm, but now that the options are available via the config file we should document them here: https://github.com/CodeIntelligenceTesting/cifuzz/blob/main/docs/Configuration.md

adombeck:
> lgtm, but now that the options are available via the config file we should document them here: https://github.com/CodeIntelligenceTesting/cifuzz/blob/main/docs/Configuration.md

Done, see https://github.com/CodeIntelligenceTesting/cifuzz/blob/CLI-274-sandbox-config-setting/docs/Configuration.md#build-system

jochil (approved):

PR #152, opened 2022-07-15
Fix integration test on windows

adombeck:


fmeum:
> @fmeum: I didn't see the "The command we use for process group termination sometimes exits with a non-zero exit code." error yet, the other errors should be fixed.

That one may have been a flake. My fix back then was simply to ignore the error, we can still do that if it shows up again.

fmeum:
@adombeck I pushed a fix for the regex, the "status" group matched too greedily. If you are okay with it, you can squash it in.

fmeum (approved):

adombeck:
> @adombeck I pushed a fix for the regex, the "status" group matched too greedily. If you are okay with it, you can squash it in.

Thanks. It's a separate fix, so I kept it as a separate commit and adjusted the commit message to remove the "FIXUP".

PR #154, opened 2022-07-18
integration-tests: Limit runs rather than time

fmeum:
Limiting runs rather than time should be less flaky and also provide
faster test failures in case of breakages.

The value of 1,000,000 runs is 5x as high as the maximum runs count
found in CI logs across the three platforms.


jochil (approved):
yes, I think this would make more sense too

PR #160, opened 2022-07-20
installer: Fix race when running multiple installers in parallel

adombeck:
This caused non-deterministic failures in tests which use the install
package.


adombeck:
> Would copying the project directory be an alternative to using mutexes? Not saying that would be better, I just want to make sure I understand the underlying issue.

Yes, that should work as well but would mean that each test which uses any of our dependencies has to rebuild them without cache

PR #161, opened 2022-07-20
run: Suggest running without sandboxing

adombeck:
... when the sandboxed command fails unexpectedly.

Also improve handling of libFuzzer errors.


adombeck:
This is based on #146, please review and merge that first

adombeck:
I'm reassigning the review because this touches some code authored by @fmeum

fmeum (approved):

internal/cmd/root/root.go
```diff
@@ -121,6 +121,23 @@ func Execute() {
...
+		var couldBeSandboxError *cmdutils.CouldBeSandboxError
+		if errors.As(err, &couldBeSandboxError) {
+			// Ensure that there is an extra newline between the error
+			// and the following message
+			if !strings.HasSuffix(err.Error(), "\n") {
+				_, _ = fmt.Fprintln(cmd.ErrOrStderr())
+			}
+			msg := `If you don't expect this fuzz test to do any harm to the system
+accidentally (like overwriting files), you might want to try
+running it without sandboxing:
```
jochil, line 133: @adombeck I know that I am late to the party :)

I am not sure if it makes sense here to limit the CLI output to our 72 chars (I guess this was the idea), too. It might be better letting the users terminal handle the line breaks... like all the other tools do. Otherwise this looks a little strange, especially if the user goes under 72 and our line breaks interfere with the terminal line breaks
```
libminijail[15]: clone(CLONE_NEWPID | ...) failed with EPERM; is this process missing CAP_SYS_ADMIN?: Operation not permitted
❌ minijail0: signal: aborted (core dumped)

If you don't expect this fuzz test to do any harm to the system
accidentally (like overwriting files), you might want to try
running it without sandboxing:
```

fmeum: We've had this discussion before ;-) Regardless of the outcome, we should do this consistently between this message and all the other places we print messages (e.g. `cifuzz create`). 

How about we introduce a helper function that takes care of the formatting? Can we detect the width of the terminal and use something like `min(terminal width, 100)`?

adombeck: Quoting the [old discussion](https://github.com/CodeIntelligenceTesting/cifuzz/pull/139#discussion_r920045044):

> Prose is more readable with short lines, but I still prefer a relatively large terminal to be able to view longer lines of structured output.
> 
> Looking at other tools, cobra seems to wrap at 80 characters: https://github.com/spf13/cobra/blob/main/completions.go#L684
> 
> ... and gh at 100: https://github.com/cli/cli/blob/0e3c3bb4a428eff7fcff472a34e5208561a9406c/pkg/cmd/auth/refresh/refresh.go#L51-L51

So it's not correct that all other tools don't wrap long lines, and I do see value in wrapping long lines.

adombeck: > We've had this discussion before ;-) Regardless of the outcome, we should do this consistently between this message and all the other places we print messages (e.g. cifuzz create).

True, if we agree to wrap long lines, I can create a PR which does that consistently.

> How about we introduce a helper function that takes care of the formatting? Can we detect the width of the terminal and use something like min(terminal width, 100)?

I don't think we have to take the terminal width into account here, flowing lines of text become more readable when they are short, no matter how wide the terminal is.

jochil: I think the discussion we had was about line breaks in the help messages of the cobra arguments and how it looks in the source code not about what we actually print on the terminal (if I remember correctly)

> How about we introduce a helper function that takes care of the formatting? Can we detect the width of the terminal and use something like min(terminal width, 100)?

I am not a fan of this. Detecting terminal width and reacting to it, sounds like something we don't want to maintain (same as the bash/zsh stuff).
But I agree that it should be consistent.

I personally would prefer it like we done it in `create` and/or by 
```
test := "Lorem ipsum dolor sit amet, consectetur adipiscing elit," +
        "sed do eiusmod tempor incididunt ut labore et dolore magna aliqua."
```

jochil: > So it's not correct that all other tools don't wrap long lines, and I do see value in wrapping long lines.

Ok, but it depends a little bit on which tools you are looking. Seems like the gnu stuff is pretty consistent with not having artificial line wrapping. But IMO discussing it on tool vs tool will not bring us anywhere :)

So my points are: 

1. Terminals are doing the line wrapping for us, why taking care of it by ourselves (not talking about how it looks in the code)
2. The tools we are instrumenting (CMake, libfuzzer, go stacktraces, cobra UsageString) are not having a line wrapping so it would be more consistent for the overall output to drop it for our messages too

And same is true for the line breaks we added in the cobra argument descriptions:

adombeck: > Ok, but it depends a little bit on which tools you are looking. Seems like the gnu stuff is pretty consistent with not having artificial line wrapping. But IMO discussing it on tool vs tool will not bring us anywhere :)

I made a brief search in the git code and found that it also uses line breaks in the flowing text it prints: https://github.com/git/git/blob/master/branch.c#L179

> Terminals are doing the line wrapping for us, why taking care of it by ourselves (not talking about how it looks in the code)

Because it improves UX.

> The tools we are instrumenting (CMake, libfuzzer, go stacktraces, cobra UsageString) are not having a line wrapping so it would be more consistent for the overall output to drop it for our messages too

That's not correct, another brief in search in the libfuzzer code showed that it's also using line breaks in the flowing text it prints: https://github.com/llvm/llvm-project/blob/main/compiler-rt/lib/fuzzer/FuzzerDriver.cpp#L866

The usage message is indeed a mess, I think we should move the explanations to the command description and keep the flag descriptions short.

jochil: As I said, I think comparing tool by tool will not help us... for example `man`, `ls` or `docker` are not doing it. So what is conclusion here? And git is btw very flexible about it, for exampe `git help commit`

> Because it improves UX.

How? IMO In the best case it is wasting space in the worst case it is scrambling the output
Why does this output have a better UX 

than this

And the latter actually scales better


> That's not correct, another brief in search in the libfuzzer code showed that it's also using line breaks in the flowing text it prints: https://github.com/llvm/llvm-project/blob/main/compiler-rt/lib/fuzzer/FuzzerDriver.cpp#L866

In the same file are multiple examples with long lines not wrapped manually:
https://github.com/llvm/llvm-project/blob/ceb6c23b708d4cae3fbb0a569c5ac14069524a63/compiler-rt/lib/fuzzer/FuzzerDriver.cpp#L355
https://github.com/llvm/llvm-project/blob/ceb6c23b708d4cae3fbb0a569c5ac14069524a63/compiler-rt/lib/fuzzer/FuzzerDriver.cpp#L205
https://github.com/llvm/llvm-project/blob/ceb6c23b708d4cae3fbb0a569c5ac14069524a63/compiler-rt/lib/fuzzer/FuzzerDriver.cpp#L420
For me it looks like they just keep their column limit for the source code and the line breaks are there where sentences end or it makes sense semantically (which btw is what I would prefer for us). Your examples looks like a special case for me

> The tools we are instrumenting (CMake, libfuzzer, go stacktraces, cobra UsageString) are not having a line wrapping

From my point of view, this is still true. I don't see why we should enforce the limit we want on our source code (where I am totally committed to) on the terminal output when all the other output we printing out is not following the same rules.

> The usage message is indeed a mess, I think we should move the explanations to the command description and keep the flag descriptions short.

Sounds like a workaround. If the flag needs that amount of explanation it should be ok to have it

adombeck: > As I said, I think comparing tool by tool will not help us... for example man, ls or docker are not doing it.

* `man` goes a step further and allows the user to specify the maximum line length: `MANWIDTH=80 man man`.
* Where do you see flowing text output of `ls` which is not wrapped? I see for example this in the output of `ls --help`:
```
Using color to distinguish file types is disabled both by default and
with --color=never.  With --color=auto, ls emits color codes only when
standard output is connected to a terminal.  The LS_COLORS environment
variable can change the settings.  Use the dircolors command to set it.
```
The normal output of `ls` or `ls -l` is structured output which should of course use the full terminal width.
* Again, where do you see flowing text output of `docker` which is not wrapped? I see this: https://github.com/moby/moby/blob/master/client/service_create.go#L164

> And git is btw very flexible about it, for exampe git help commit

That is just using `man`. When it's not using `man`, for example `git --help`, it wraps long lines.

> In the same file are multiple examples with long lines not wrapped manually:
> https://github.com/llvm/llvm-project/blob/ceb6c23b708d4cae3fbb0a569c5ac14069524a63/compiler-rt/lib/fuzzer/FuzzerDriver.cpp#L355
> https://github.com/llvm/llvm-project/blob/ceb6c23b708d4cae3fbb0a569c5ac14069524a63/compiler-rt/lib/fuzzer/FuzzerDriver.cpp#L205
> https://github.com/llvm/llvm-project/blob/ceb6c23b708d4cae3fbb0a569c5ac14069524a63/compiler-rt/lib/fuzzer/FuzzerDriver.cpp#L420

Those look like structured log lines to me, because they start with their log level (`INFO:` or `ERROR:`, they also use `WARNING:`), which makes it useful to be in a single line because it allows to grep for specific messages / log levels. That's a different use case from the message we're discussing here IMO: We want to make the user aware of something, not log diagnostical output. The "NOTE:" message I linked above, which *is* wrapped, has the same use case.

> Sounds like a workaround. If the flag needs that amount of explanation it should be ok to have it

Again, I've been looking at how other tools deal with the problem and that's what they do, for example `gh` in the line I already linked above, where they explain the `--scopes` flag in more detail the command description: https://github.com/cli/cli/blob/0e3c3bb4a428eff7fcff472a34e5208561a9406c/pkg/cmd/auth/refresh/refresh.go#L51-L51

jochil: > man goes a step further and allows the user to specify the maximum line length: MANWIDTH=80 man man.
yeah but I would avoid doing that... a sane default is something we should look for

> Again, where do you see flowing text output of docker which is not wrapped?

image

> for example gh in the line I already linked above, where they explain the --scopes flag in more detail in the command description:

And this is what it looks like with 84:
image

From my perspective it boils down to the following: If you choose a width of 100 (like gh) or more it will look shitty in smaller terminals (see picture above), so you have to go for something <=80, which IMO is a waste of space and looks unbalanced in the context of our other output from the tools we are using.
So why even going for it. The "full-width" variant will work the best on all terminals and for me it is as readable as the "short" version. If you need more highlighting why not adding a box, symbol, color or a divider

PR #163, opened 2022-07-21
integration-tests: Fix incorrect argument passed to libFuzzer

adombeck:


jochil (approved):

fmeum:
`1000` is pretty low, I think we'll have to change this to something like `1000000`.

adombeck:
> `1000` is pretty low, I think we'll have to change this to something like `1000000`.

done

fmeum (approved):

PR #166, opened 2022-07-21
run: Fix "Starting from enpty corpus" printed

adombeck:
... even though the corpus is non-empty.

The `err` value was overwritten, which affected the `err == nil` check
below.

### DIFF
--- a/pkg/parser/libfuzzer/libfuzzer_parser.go
+++ b/pkg/parser/libfuzzer/libfuzzer_parser.go
@@ -151,7 +151,7 @@ func (p *parser) parseLine(ctx context.Context, line string) error {
 				// Store all lines printed before the fuzzer has been initialized
 				// so that they can be printed in case of a startup error (e.g.
 				// a missing shared library dependency).
-				_, err = p.StartupOutputWriter.Write(append([]byte(line), '\n'))
+				_, err := p.StartupOutputWriter.Write(append([]byte(line), '\n'))
 				if err != nil {
 					return errors.WithStack(err)
 				}


pkg/parser/libfuzzer/libfuzzer_parser.go
```diff
@@ -151,7 +151,7 @@ func (p *parser) parseLine(ctx context.Context, line string) error {
 				// Store all lines printed before the fuzzer has been initialized
 				// so that they can be printed in case of a startup error (e.g.
 				// a missing shared library dependency).
-				_, err = p.StartupOutputWriter.Write(append([]byte(line), '\n'))
+				_, err := p.StartupOutputWriter.Write(append([]byte(line), '\n'))
```
fmeum, line 154: I don't understand yet why adding a colon here fixes a bug. Could you explain this in a bit more detail? Where is `err` overwritten?

adombeck: The variable `err` which is previously set [here](https://github.com/CodeIntelligenceTesting/cifuzz/blob/fix-number-of-seeds/pkg/parser/libfuzzer/libfuzzer_parser.go#L144-L144) is overwritten in this line when using `=`. By using `:=` instead, a new variable `err` is defined which is scoped to the `if` block.

adombeck: See https://go.dev/ref/spec#Declarations_and_scope:
> 7. The scope of a constant or variable identifier declared inside a function begins at the end of the ConstSpec or VarSpec (ShortVarDecl for short variable declarations) and ends at the end of the innermost containing block.

fmeum: I see, this is all about the check at https://github.com/CodeIntelligenceTesting/cifuzz/blob/2b7194a952aa3e0c02f60b6955bfa649f02cff7e/pkg/parser/libfuzzer/libfuzzer_parser.go#L160, right? Then I get it :-)

fmeum (approved):

pkg/parser/libfuzzer/libfuzzer_parser.go
adombeck: Yes, that is the "err == nil check below".

PR #167, opened 2022-07-21
run: Fix default seed corpus not used

adombeck:


jochil:
Florian and I noticed that a few days ago and thought this was on intent. We encourage the user to copy the crashing input to the seed corpus which will made all following fuzzing runs fail immediately. (which I think is the reason libfuzzer puts the crashing input somewhere else, or?)
If we want this... ok. But then we should change this message too, as it becomes misleading
```
You can add this crashing input to the seed corpus with:                                                                                                   
  mkdir -p my_fuzz_test_seed_corpus && cp .cifuzz-findings/busy_khorana/crashing-input my_fuzz_test_seed_corpus/busy_khorana                               
After adding, the input will be applied every time you run the regression tests / replayer binary (for example during your CI/CD pipeline).                
For more information you can take a look at https://github.com/CodeIntelligenceTesting/cifuzz#regression-testing    
```

internal/cmd/run/run.go
```diff
@@ -66,6 +67,21 @@ func (opts *runOptions) validate() error {
...
+	defaultSeedCorpusDir := cmdutils.DefaultSeedCorpusDir(opts.fuzzTest)
+	defaultSeedCorpusDir, err = filepath.Abs(defaultSeedCorpusDir)
+	if err != nil {
+		return errors.WithStack(err)
+	}
+	exists, err := fileutil.Exists(defaultSeedCorpusDir)
+	if err != nil {
+		return err
+	}
+	if exists && !stringutil.Contains(opts.SeedCorpusDirs, defaultSeedCorpusDir) {
```
fmeum, line 81: Is it problematic if the same directory is added twice as a seed corpus directory? If so, we might want to be more careful here and compare after calling `filepath.Clean` or even `filepath.EvalSymlinks`.

fmeum:
@jochil If I understand the commit correctly, it only adds the default seed corpus to the list of *seed* corpus directories - the *generated* corpus is still kept in a separate directory (`.cifuzz-corpus/<fuzz test name>`).

jochil:
> @jochil If I understand the commit correctly, it only adds the default seed corpus to the list of _seed_ corpus directories - the _generated_ corpus is still kept in a separate directory (`.cifuzz-corpus/<fuzz test name>`).

yes... this is what I understood too and I was not talking about the generated corpus. What I was trying to say: After a finding we prompt the user the message above, about adding the crashing input to the seed corpus... which reads like this will **only** affect the replayer / regression test. With this change this is not true anymore

internal/cmd/run/run.go
adombeck: Well, libFuzzer tries the seeds from that corpus twice if it's added twice. I don't think that is very problematic though, it just wastes some cycles. We already call `filepath.Abs`, which calls `filepath.Clean`, both on `defaultSeedCorpusDir` and all dirs in `opts.SeedCorpusDirs`, so we don't need another call to `filepath.Clean`. Regarding `filepath.EvalSymlinks`, I think we should call that anyway on all seed corpus dirs to avoid adding a minijail binding for a symlink which can't be resolved inside the sandbox.

pkg/cmdutils/paths.go
```diff
@@ -0,0 +1,5 @@
+package cmdutils
+
+func DefaultSeedCorpusDir(fuzzTest string) string {
+	return fuzzTest + "_seed_corpus"
```
fmeum, line 4: This will not find the seed corpus for CMake targets: The seed corpus is stored in the source tree, whereas the fuzzer executable lives in the build tree.

In fact, if `fuzzTest` is just the name of the fuzz test, wouldn't this also not find the default seed corpus with Make if the fuzz test binary lives in a subdirectory?

I think that we should use the `FindFuzzTestSeedCorpus` method for CMake and look next to the actual fuzz test executable for Make.

jochil:
I agree, I was not arguing against the chance itself... just about the missing adjustment of the message ;)

pkg/cmdutils/paths.go
adombeck: I also thought that the plan was to have the seed corpus next to the fuzz test executable, and I would still prefer that, but this is the directory which the command displayed by `cifuzz run` currently copies the input to.

internal/cmd/run/run.go
adombeck: > Regarding filepath.EvalSymlinks, I think we should call that anyway on all seed corpus dirs to avoid adding a minijail binding for a symlink which can't be resolved inside the sandbox.

Done

pkg/cmdutils/paths.go
adombeck: The correct seed corpus for the current build system is now used.

adombeck: @fmeum: I did the refactoring we briefly discussed in https://github.com/CodeIntelligenceTesting/cifuzz/pull/145#discussion_r922032416 to have a build result struct which contains the seed corpus (as well as all other information needed after a build). Please take a look.

internal/build/cmake/cmake.go
```diff
@@ -162,22 +162,47 @@ func (b *Builder) Build(fuzzTests []string) error {
...
+	for _, fuzzTest := range fuzzTests {
+		executable, err := b.findFuzzTestExecutable(fuzzTest)
+		if err != nil {
+			return nil, err
+		}
+		seedCorpus, err := b.findFuzzTestSeedCorpus(fuzzTest)
+		if err != nil {
+			return nil, err
+		}
+		runtimeDeps, err := b.getRuntimeDeps(fuzzTest)
```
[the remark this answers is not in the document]
adombeck: Mmh or we add an `FindRuntimeDeps` option to `BuilderOptions` which we only set if we need the runtime deps in the build result?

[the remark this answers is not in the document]
fmeum: That would work as well for our current use cases. I'm fine with both.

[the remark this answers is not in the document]
adombeck: done

adombeck:
> If we want this... ok. But then we should change this message too, as it becomes misleading

Lets change the seed corpus message in a follow-up, because we have to change it anyway for https://code-intelligence.atlassian.net/browse/CLI-308

fmeum (approved):

PR #169, opened 2022-07-22
cmake: Build the replayer with the fuzz test

fmeum:
This gets rid of a big bag of hacks and cleanly ensures that the
replayer is built with the same flags as the fuzz test, which enables
e.g. cross-compilation scenarios.

Additionally, it saves us from committing to either the legacy or the
modern signature of target_link_library, both of which can result in
warnings or errors depending on the user's CMake version.


tools/install/installer.go
```diff
@@ -400,8 +400,11 @@ func (i *Installer) copyCMakeIntegration(destDir string) (string, error) {
 	if err != nil {
 		return "", errors.WithStack(err)
 	}
-	replayerDst := filepath.Join(replayerDir, "replayer.c")
-	err = copy.Copy(replayerSrc, replayerDst)
+	err = copy.Copy(replayerSrc, filepath.Join(replayerDir, "replayer.c"))
+	if err != nil {
+		return "", errors.WithStack(err)
+	}
+	err = copy.Copy(replayerSrc, filepath.Join(replayerDir, "replayer.cpp"))
```
adombeck, line 407: IMO it would be cleaner to use `os.Symlink` here to create a symlink to the installed `replayer.c` file here instead of copying the file. I think the `copy.Copy` call above in line 390 would already do that if the `OnSymlink` action would return the default, `copy.Shallow`, for `replayer.cpp`.

adombeck (approved):

tools/cmake/cifuzz/cmake/cifuzz-config.cmake
```diff
@@ -7,3 +6,5 @@ set(CIFUZZ_SANITIZERS "" CACHE STRING "The sanitizers to instrument the code wit
 set(CIFUZZ_USE_DEPRECATED_MACROS OFF CACHE BOOL "Whether to use the deprecated FUZZ(_INIT) macros instead of FUZZ_TEST(_SETUP)")
 
 set(CIFUZZ_INCLUDE_DIR "${CMAKE_CURRENT_LIST_DIR}/../include/cifuzz" CACHE INTERNAL "The include directory for the cifuzz headers")
+set(CIFUZZ_REPLAYER_C_SRC "${CMAKE_CURRENT_LIST_DIR}/../src/replayer.c" CACHE INTERNAL "The path of the replayer as a C source file.")
+set(CIFUZZ_REPLAYER_CXX_SRC "${CMAKE_CURRENT_LIST_DIR}/../src/replayer.cpp" CACHE INTERNAL "The path of the replayer as a CXX source file.")
```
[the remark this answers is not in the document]
fmeum: I think I would like to keep `CXX` here: This comment is not shown to users (it's an internal setting) and we specifically use this file to deal with projects declaring `LANGUAGES CXX`.

tools/install/installer.go
fmeum: I will take a look, but if it ends up symlinking on Windows as well, that would be a show-stopper.

fmeum: The problem is that Git on Windows will check out the symlinks as text files containing the symlink target - copying these over is never correct, we have to replace them manually. I thus don't think we can do better than replacing the explicit `copy.Copy` with a utility function that symlinks on Unix and copies on Windows. Do you think that would be better?

adombeck: Ah I didn't know about this issue with symlinks and git for Windows. Lets explain that in a comment. I don't think it's worth treating Windows and Unix differently here, so I'm fine with keeping the copy.

fmeum: The comment already exists in line 384, but I extended it a bit to mention git.

PR #170, opened 2022-07-22
all: Let the caller handle ExecErrors

adombeck:
The executil.HandleExecError function used to log error messages before
the caller could decide what to do with the error. We now wrap the error
in a cmdutils.ExecError and let the caller decide whether to print the
error message or not.

This avoids printing two errors when Ctrl+C is used to interrupt a run:

    ❌ minijail0: exit status 72
    ❌ terminated by signal 2 (interrupt)

and just prints the second error instead (as was intended).


pkg/cmdutils/errors.go
```diff
@@ -86,3 +89,32 @@ func (e CouldBeSandboxError) Unwrap() error {
 func WrapCouldBeSandboxError(err error) error {
 	return &CouldBeSandboxError{err}
 }
+
+// ExecError includes information about the exec.Cmd which failed in the
+// error message.
+type ExecError struct {
+	err error
```
fmeum, line 96: Not a must, but it may be cleaner to embed `error` here.

internal/cmd/run/run.go
```diff
@@ -327,6 +327,14 @@ func (c *runCmd) runFuzzTest(fuzzTestExecutable string) error {
 		return cmdutils.WrapSilentError(signalErr)
 	}
 
+	var execErr *cmdutils.ExecError
+	if errors.As(err, &execErr) {
```
fmeum, line 331: Can we use `errors.Is` instead since we aren't using `execErr` afterwards?

fmeum (approved):

internal/cmd/run/run.go
adombeck: I don't think we can use `errors.Is` here, because it only checks if an error in the chain is *equal* to the target. Quoting https://pkg.go.dev/errors#Is:
> An error is considered to match a target if it is equal to that target or if it implements a method Is(error) bool such that Is(target) returns true.

`errors.As` on the other hand checks if any error in the chain is *assignable* to the value pointed to by target:
> An error matches target if the error's concrete value is assignable to the value pointed to by target, or if the error has a method As(interface{}) bool such that As(target) returns true. In the latter case, the As method is responsible for setting target. 

IIUC, to use `errors.Is`, we would have to use a constant error like this:
```
var execErr = errors.New("exec error")
if errors.Is(err, execErr) {
    ...
}
```
... which we can't because that error doesn't store the information about the command which we want to print.

fmeum: Makes sense. But a simple checked cast would work and allow us to emit the effectively unused `exitErr`, right?

pkg/cmdutils/errors.go
adombeck: I'd prefer to keep it as is, because that's the way we defined our other errors which wrap errors and it's the same way [`gh`](https://github.com/cli/cli/blob/0e3c3bb4a428eff7fcff472a34e5208561a9406c/pkg/cmdutil/errors.go#L23-L23) and the [standard library](https://go.dev/src/fmt/errors.go) do it.

PR #171, opened 2022-07-22
minijail: Fix regex in output filter

adombeck:
The pattern didn't match the line we want to filter.

Also, test that the line is filtered in an integration test.

### DIFF
--- a/integration-tests/make/make_test.go
+++ b/integration-tests/make/make_test.go
@@ -17,6 +17,7 @@ import (
 )
 
 var expectedFinding = regexp.MustCompile(`^==\d*==ERROR: AddressSanitizer: heap-buffer-overflow`)
+var filteredLine = regexp.MustCompile(`child process \d+ exited`)
 
 func TestIntegration_Make_Run(t *testing.T) {
 	if testing.Short() {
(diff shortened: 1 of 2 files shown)


integration-tests/make/make_test.go
```diff
@@ -94,6 +95,9 @@ func runFuzzer(t *testing.T, cifuzz string, dir string, fuzzTest string, expecte
 		if expectedOutput.MatchString(scanner.Text()) {
 			seenExpectedOutput = true
 		}
+		if minijail.IgnoredPattern.MatchString(scanner.Text()) {
```
fmeum, line 98: I can see arguments for both sides, but would prefer to have a (coarse) pattern hardcoded in the test, e.g., just `child process \d+ exited`. Otherwise, wouldn't this have missed the bug fixed in this PR?

adombeck: > Otherwise, wouldn't this have missed the bug fixed in this PR?

That's correct.

adombeck: Done

fmeum (approved):

PR #173, opened 2022-07-25
refactor: drop promptui in favor of pterm

jochil:


pkg/dialog/dialog.go
```diff
@@ -1,26 +1,21 @@
...
-	if err == promptui.ErrInterrupt {
-		return "", cmdutils.WrapSilentError(errors.WithStack(err))
-	}
+func Select(message string, items map[string]string) (string, error) {
+	options := maps.Keys(items)
+	sort.Strings(options)
+	prompt := pterm.DefaultInteractiveSelect.WithOptions(options)
+	prompt.DefaultText = message
+
+	result, err := prompt.Show()
```
adombeck, line 18: The `Show()` function handles `Ctrl+C` by just exiting: https://github.com/pterm/pterm/blob/master/interactive_select_printer.go#L218-L219

That seems problematic because we might have to clean up things first. `promptui` handles this a lot cleaner by returning an `ErrInterrupt` which we handled by wrapping it in a `SilentError`. In fact, this does mess up my terminal: Once I run `cifuzz create` and press `Ctrl+C`, my terminal doesn't clear the current command anymore when I press `Ctrl+C`.

jochil: Oh... happens with my terminal too. I will take a deeper look
Didn't notice as I just compared the Ctrl+C behavior... which from the cifuzz perspective is the same as before (we ended with Exit=1 and no error shown)

jochil: Ok. It is a bug in pterm... in the current implementation pterm itself just doesn't clean up when exiting with Ctrl+C. I made a PR for fixing this (https://github.com/pterm/pterm/pull/383)... now we have to wait :)

adombeck: If we already create a pterm PR, why don't we ensure that we don't run into the same issue again in our code (in case that we ever have to do some cleanup) by returning an ErrInterrupt like promptui does?

jochil: I thought about it, but I decided not to do so because it would be a breaking change for pterm and therefore less likely that they accept it... Especially if we currently not really using it.
But yes in general it would be a nicer solution and I thought about trying to get it in another PR, afterwards

adombeck:
I'm converting this into a draft until the pterm bug is solved

jochil:
> I'm converting this into a draft until the pterm bug is solved

It is solved/released and we updated to the new version with this PR: https://github.com/CodeIntelligenceTesting/cifuzz/pull/177

adombeck (approved):

PR #176, opened 2022-07-26
Add coverage command

adombeck:


internal/build/unknown/unknown.go
```diff
@@ -152,6 +203,50 @@ func setBuildFlagsEnvVars(env []string) ([]string, error) {
...
+	//
+	// Note: Keep in sync with tools/cmake/CIFuzz/share/CIFuzz/CIFuzzFunctions.cmake
+	cflags := []string{
+		// ----- Common flags -----
+		// Keep debug symbols
+		"-g",
+		// Do optimizations which don't harm debugging
+		"-Og",
+		// To get good stack frames for better debugging
+		"-fno-omit-frame-pointer",
```
fmeum, line 220: I would prefer a common slice with the common flags, shared between the libfuzzer and coverage env functions.

internal/cmd/coverage/coverage.go
```diff
@@ -0,0 +1,365 @@
...
+		Use:   "coverage [flags] <fuzz test>",
+		Short: "Generate a coverage report for a fuzz test",
+		// TODO: Write long description
+		Long:              "",
+		ValidArgsFunction: completion.ValidFuzzTests,
+		Args:              cobra.ExactArgs(1),
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			// Bind viper keys to flags. We can't do this in the New
+			// function, because that would re-bind viper keys which
+			// were bound to the flag sof other commands before.
```
fmeum, line 106:
```suggestion
			// were bound to the flags of other commands before.
```

internal/cmd/coverage/coverage.go
```diff
@@ -0,0 +1,365 @@
...
+			return cmd.run()
+		},
+	}
+
+	// Note: If a flag should be configurable via cifuzz.yaml as well,
+	// bind it to viper in the PreRunE function.
+	cmd.Flags().String("build-command", "", "The command to build the fuzz test. Example: \"make clean && make my-fuzz-test\"")
+	cmd.Flags().StringArrayP("seed-corpus", "s", nil, "Directory containing sample inputs for the code under test.\nSee https://llvm.org/docs/LibFuzzer.html#corpus and\nhttps://aflplus.plus/docs/fuzzing_in_depth/#a-collecting-inputs.")
+	cmd.Flags().StringArray("fuzz-test-arg", nil, "Command-line argument to pass to the fuzz test.")
+	cmd.Flags().Bool("use-sandbox", false, "By default, fuzz tests are executed in a sandbox to prevent accidental damage to the system.\nUse --sandbox=false to run the fuzz test unsandboxed.\nOnly supported on Linux.")
```
fmeum, line 131:
```suggestion
	cmd.Flags().Bool("use-sandbox", false, "By default, fuzz tests are executed in a sandbox to prevent accidental damage to the system.\nUse --use-sandbox=false to run the fuzz test unsandboxed.\nOnly supported on Linux.")
```

internal/cmd/coverage/coverage.go
```diff
@@ -0,0 +1,365 @@
...
+	viper.SetDefault("use-sandbox", runtime.GOOS == "linux")
```
fmeum, line 132: In case the defaults are shown in the CLI help message, wouldn't `use-sandbox be shown as disabled by default on Linux?

internal/cmd/coverage/coverage.go
```diff
@@ -0,0 +1,365 @@
...
+	exists, err := fileutil.Exists(generatedCorpusDir)
+	if err != nil {
+		return err
+	}
+	if exists {
+		seedDirs = append(seedDirs, generatedCorpusDir)
+	}
+
+	var seeds []string
+	for _, dir := range seedDirs {
```
fmeum, line 231: This shouldn't be needed: The replayer traverses directories recursively, so passing the dirs should suffice. Only possible issue I see with this is that the directory entry traversal order is clearly defined by `os.ReadDir`.

internal/cmd/coverage/coverage.go
```diff
@@ -0,0 +1,365 @@
...
+			return errors.WithStack(err)
+		}
+		for _, entry := range entries {
+			seeds = append(seeds, filepath.Join(dir, entry.Name()))
+		}
+	}
+
+	// The environment we run the binary in
+	var binaryEnv []string
+	// TODO: Should we put the LLVM profile file in a temporary or
```
fmeum, line 243: I think that we should put it into a temporary location, but print its path with `VERBOSE=1`. I don't expect users to every need it since the indexed profile contains strictly more information.

internal/cmd/coverage/coverage.go
```diff
@@ -0,0 +1,365 @@
...
+	//       hidden directory? Or could it be useful to the user and
+	//       should therefore be visible?
+	binaryEnv, err = envutil.Setenv(binaryEnv, "LLVM_PROFILE_FILE", c.rawProfilePath(fuzzTestExecutable))
```
fmeum, line 246: Using a fixed path will not work if the fuzz test references shared libraries: Each executable or library generates its own profile file, all of which we have to merge in the end. You can use [`%m`](https://clang.llvm.org/docs/SourceBasedCodeCoverage.html#running-the-instrumented-program) to have the runtime generate unique profile names automatically.

Also, we may want to add `%c` to get profile data even with a sanitizer crash.

internal/cmd/coverage/coverage.go
```diff
@@ -0,0 +1,365 @@
...
+		err = cmdutils.WrapExecError(err, cmd)
+		log.Error(err)
+		return cmdutils.ErrSilent
+	}
+	return nil
+}
+
+func (c *coverageCmd) indexRawProfile(fuzzTestExecutable string) error {
+	var err error
+	cmd := exec.Command("llvm-profdata", "merge", "-sparse", c.rawProfilePath(fuzzTestExecutable),
```
fmeum, line 322: We should introduce runfiles finder methods for `llvm-profdata` and `llvm-cov`. Thinking of it, we should then also add one for `clang` since instrumentation and coverage tooling have to remain in sync.

internal/cmd/coverage/coverage_test.go
```diff
@@ -0,0 +1,15 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"code-intelligence.com/cifuzz/pkg/cmdutils"
+)
+
+func TestCoverageCmd(t *testing.T) {
+	_, err := cmdutils.ExecuteCommand(t, New(), os.Stdin)
+	assert.Error(t, err)
```
fmeum, line 14: What does this test actually do? We are not asserting anything about the particular nature of the error, so I don't see what value it adds.

Instead, we should probably add a step to the integration test that e.g. invokes `llvm-cov report -format=text` on the indexed profile and makes some assertions on the coverage summary.

internal/build/unknown/unknown.go
```diff
@@ -0,0 +1,188 @@
+package unknown
```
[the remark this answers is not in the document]
adombeck: I prefer `other` because in contrast to the other options I think that one makes it more clear that this is our fallback mode.

internal/cmd/run/run.go
```diff
@@ -66,6 +67,21 @@ func (opts *runOptions) validate() error {
 		}
 	}
 
+	// If the default seed corpus dir exists, add it to the list of
+	// seed corpus dirs
+	defaultSeedCorpusDir := cmdutils.DefaultSeedCorpusDir(opts.fuzzTest)
```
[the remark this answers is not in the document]
adombeck: Yes, this is #167. I would like to implement the lookup procedure in the refactored `unknown.Builder`, so I'll remove that commit from this branch.

internal/build/unknown/unknown.go
```diff
@@ -21,38 +21,89 @@ import (
...
+	switch opts.Engine {
+	case "libfuzzer":
+		// Set CFLAGS, CXXFLAGS, LDFLAGS, and FUZZ_TEST_LDFLAGS which must
+		// be passed to the build commands by the build system.
+		b.env, err = setLibFuzzerEnv(b.env)
+		if err != nil {
+			return nil, err
+		}
+	case "coverage":
+		err = b.setCoverageEnv()
```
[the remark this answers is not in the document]
adombeck: I now made `setLibFuzzerEnv` a method of `builder` as well, with the same signature.

internal/cmd/run/run.go
```diff
@@ -221,7 +221,7 @@ func (c *runCmd) buildFuzzTest() (string, error) {
 	if c.opts.BuildSystem == config.BuildSystemCMake {
 		builder, err := cmake.NewBuilder(&cmake.BuilderOptions{
 			ProjectDir: c.opts.ProjectDir,
-			// TODO: Do not hardcode this values.
+			// TODO: Do not hardcode these values.
```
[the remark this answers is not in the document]
adombeck: done

internal/build/unknown/unknown.go
adombeck: done

internal/cmd/coverage/coverage.go
```diff
@@ -0,0 +1,365 @@
...
+	ProjectDir string
+	fuzzTest   string
+}
+
+func (opts *coverageOptions) validate() error {
+	var err error
+
+	// Check if the seed dirs exist and can be accessed and ensure that
+	// the paths are absolute
+	for i, d := range opts.SeedCorpusDirs {
```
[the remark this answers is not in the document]
adombeck: done

internal/cmd/coverage/coverage.go
adombeck: fixed

internal/cmd/coverage/coverage.go
adombeck: fixed

internal/cmd/coverage/coverage.go
adombeck: Removed

internal/cmd/coverage/coverage.go
adombeck: Do you think we should put the indexed profile in the current working directory then? Or also in the temporary directory?

adombeck: > I think that we should put it into a temporary location, but print its path with VERBOSE=1.

When we create a temporary directory, we should clean it up before the command exits, which would make it pointless to print its path.

internal/cmd/coverage/coverage.go
adombeck: > Using a fixed path will not work if the fuzz test references shared libraries: Each executable or library generates its own profile file, all of which we have to merge in the end. You can use [%m](https://clang.llvm.org/docs/SourceBasedCodeCoverage.html#running-the-instrumented-program) to have the runtime generate unique profile names automatically.

I'm now using `%m` but still have to test this with a fuzz test which references shared libraries.

> Also, we may want to add %c to get profile data even with a sanitizer crash.

If I add `%c`, no profile files are generated

internal/cmd/coverage/coverage.go
adombeck: Why? The runfiles finder we in cifuzz just looks in the PATH, just as `exec.Command` does. And we don't use the cifuzz packages anywhere else anymore. In fact, I think we should get rid of the runfiles finder interface: https://code-intelligence.atlassian.net/browse/CLI-313

internal/cmd/coverage/coverage.go
fmeum: Maybe we should just not expose the indexed profile for now - so keep it in a temporary directory that we clean up in the end.

internal/cmd/coverage/coverage.go
fmeum: We can leave out the `%c` for now, but this is certainly unexpected. It uses mmap, maybe this interacts badly with minijail?

internal/cmd/coverage/coverage_test.go
adombeck: > What does this test actually do? We are not asserting anything about the particular nature of the error, so I don't see what value it adds.

That was just copy-and-pasted from `internal/cmd/run/run_test.go`.

> Instead, we should probably add a step to the integration test that e.g. invokes llvm-cov report -format=text on the indexed profile and makes some assertions on the coverage summary.

Yes, absolutely.

internal/cmd/coverage/coverage.go
adombeck: Same issue when I run it without minijail

fmeum: Strange, I found it to work very well for Jazzer's tests. We can keep this as is and I will take a look once we have the integration test.

internal/cmd/coverage/coverage_test.go
adombeck: > Instead, we should probably add a step to the integration test that e.g. invokes llvm-cov report -format=text on the indexed profile and makes some assertions on the coverage summary.

If we want to run `llvm-cov report` on the indexed profile, we need to print its output path and create it in a location which can still be accessed after the coverage command exited.

internal/build/unknown/unknown.go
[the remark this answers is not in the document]
adombeck: See https://github.com/CodeIntelligenceTesting/cifuzz/pull/178

internal/cmd/coverage/coverage.go
adombeck: Done

internal/cmd/coverage/coverage.go
adombeck: Done

internal/cmd/coverage/coverage.go
adombeck: > Using a fixed path will not work if the fuzz test references shared libraries: Each executable or library generates its own profile file, all of which we have to merge in the end. 

I tested this and the coverage report looks complete even when I use a fixed path and a shared library.

internal/cmd/coverage/coverage_test.go
adombeck: I added assertions which check that the reports include the source files which contain the function used by the fuzz test.

integration-tests/cmake/cmake_test.go
```diff
@@ -100,6 +100,13 @@ func TestIntegration_CMake_InitCreateRunBundle(t *testing.T) {
 	err = os.WriteFile(filepath.Join(dir, "cifuzz.yaml"), []byte(configFileContent), 0644)
 	runFuzzer(t, cifuzz, dir, regexp.MustCompile(`-rss_limit_mb=1234`), false)
 
+	if runtime.GOOS == "windows" {
+		// Building with coverage instrumentation doesn't work on Windows yet
```
fmeum, line 104: But `cifuzz bundle` should - could we move the `cifuzz coverage` check to the end so that `bundle` is still tested on Windows?

internal/cmd/coverage/coverage.go
```diff
@@ -0,0 +1,411 @@
...
+	} else {
+		err = config.ValidateBuildSystem(opts.BuildSystem)
+		if err != nil {
+			return err
+		}
+	}
+
+	// To build with other build systems, a build command must be provided
+	if opts.BuildSystem == config.BuildSystemOther && opts.BuildCommand == "" {
+		msg := "Flag \"build-command\" must be set when using the build system type \"other\""
```
fmeum, line 77: This could use raw strings.

internal/cmd/coverage/coverage.go
```diff
@@ -0,0 +1,411 @@
...
+		},
+		RunE: func(c *cobra.Command, args []string) error {
+			cmd := coverageCmd{Command: c, opts: opts}
+			return cmd.run()
+		},
+	}
+
+	// Note: If a flag should be configurable via cifuzz.yaml as well,
+	// bind it to viper in the PreRunE function.
+	cmd.Flags().String("build-command", "", "The command to build the fuzz test. Example: \"make clean && make my-fuzz-test\"")
```
fmeum, line 125: This could also use a raw string.

internal/cmd/coverage/coverage.go
```diff
@@ -0,0 +1,411 @@
...
+	//       cmake for that or do it ourselves in Go).
+	var sharedObjects []string
+	err = filepath.Walk(".", func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return errors.WithStack(err)
+		}
+		if info.IsDir() {
+			return nil
+		}
+		if filepath.Ext(info.Name()) == ".so" && (info.Mode()&0111 != 0) {
```
fmeum, line 360: We probably want to allow `.dylib` for macOS. Also, we might want to allow versioned libraries as well, which would call for a regex [such as this one](https://cs.opensource.google/bazel/bazel/+/d44f11be11f3ec12d644eb5f5245bf70c6a65bee:src/main/java/com/google/devtools/build/lib/rules/cpp/CppFileTypes.java;l=219).

fmeum (approved):

integration-tests/cmake/cmake_test.go
adombeck: Oops, good catch. Fixed.

internal/cmd/coverage/coverage.go
adombeck: Done

internal/cmd/coverage/coverage.go
adombeck: Done

internal/cmd/coverage/coverage.go
adombeck: Done

internal/cmd/run/run.go
```diff
@@ -224,46 +226,24 @@ func (c *runCmd) buildFuzzTest() (string, error) {
 		if runtime.GOOS == "windows" {
 			return "", errors.New("CMake is the only supported build system on Windows")
```
jochil, line 215: What was the reason again for not supporting Windows? A build command (provided by the user) should be as universal as it gets, or?
Maybe we can add the reason also as a comment here

internal/build/other/other.go
```diff
@@ -21,38 +21,70 @@ import (
 
 type BuilderOptions struct {
 	BuildCommand string
+	Engine       string
```
jochil, line 24: Is `Engine` a good name here? IMO in our context `Engine` is a synonym for Fuzz Engine/Fuzzer... but here we are using it also for `coverage`. Maybe `Mode` would be a more suitable name

internal/build/other/other.go
```diff
@@ -21,38 +21,70 @@ import (
...
+		err = b.setLibFuzzerEnv()
+		if err != nil {
+			return nil, err
+		}
+	case "coverage":
+		err = b.setCoverageEnv()
+		if err != nil {
+			return nil, err
+		}
+	}
```
jochil, line 63: it's a matter of taste/style... but what's about making one out of the two `if err!=nil` blocks, by moving one directly behind the `switch` statement

internal/build/other/other.go
```diff
@@ -21,38 +21,70 @@ import (
...
 
-	// Set CFLAGS, CXXFLAGS, LDFLAGS, and FUZZ_TEST_LDFLAGS which must
-	// be passed to the build commands by the build system.
-	b.env, err = setBuildFlagsEnvVars(b.env)
+	b.env, err = build.CommonBuildEnv()
 	if err != nil {
 		return nil, err
 	}
 
+	switch opts.Engine {
```
jochil, line 50: If `opts.Engine` is not an optional field we should add a default case. and/or make  sure that it get's validated (by using constants or checking the value)

internal/cmd/run/run.go
```diff
@@ -112,6 +103,18 @@ func New() *cobra.Command {
 		ValidArgsFunction: completion.ValidFuzzTests,
 		Args:              cobra.ExactArgs(1),
 		PreRunE: func(cmd *cobra.Command, args []string) error {
+			// Bind viper keys to flags. We can't do this in the New
+			// function, because that would re-bind viper keys which
+			// were bound to the flag sof other commands before.
```
jochil, line 108:
```suggestion
			// were bound to the flags of other commands before.
```

internal/cmd/run/run.go
```diff
@@ -114,6 +103,18 @@ func New() *cobra.Command {
...
```
jochil, line 108: What do you think about this: instead of separating viper binding and flag declaration, we can check in the `cmdutils.ViperMustBindPFlag` if the flag is already bound...  this way we could keep them together and avoid mistakes by missing a particular flag when changing/adding flags

jochil:
I know I am repeating myself, but there are basically no new unit tests. I do not want to start a detailed discussion about single tests again, so it would be nice giving a second thought if there a some tests that can be added... especially as there a new/changed util functions. If not, I am fine.

jochil:
I think the README should be updated with a section explaining the coverage "workflow"? If not, how should the user know about this feature

internal/build/other/other.go
fmeum: Conceptually speaking, `coverage` is in fact both an engine and a sanitizer: An engine, since it force the use of the replayer and e.g. modifies the seed corpus it uses. A sanitizer, since it applies different instrumentation flags.

I think this could go both ways: Leave as is and make the sanitizer `coverage` implicit in the engine `coverage` or demand that engine `coverage` has to be used together with sanitizer `coverage` (and no other sanitizers).

internal/cmd/run/run.go
fmeum: A build command is universal, but that's only the very basic part: We rely on Make conventions by injecting variables such as `CFLAGS`. Sure, Make exists on Windows, but do people actually use it build Windows projects? MSBuild has completely different conventions/settings.

internal/cmd/coverage/coverage.go
```diff
@@ -0,0 +1,422 @@
...
+	}
+
+	// Add all shared objects in the project directory as arguments, for
+	// the case that the executable uses those and we therefore want
+	// them to be included in the coverage report.
+	// TODO: Only add those objects which are actually used, and which
+	//       might live outside of the project directory, by parsing the
+	//       shared object dependencies of the executable (we could use
+	//       cmake for that or do it ourselves in Go).
+	var sharedObjects []string
```
fmeum, line 355: The CI failure on macOS shows that this logic is actually too simple. We can add some excludes, but I think we should at least use the more fine-grained CMake function for the CMake build system case.

internal/cmd/run/run.go
jochil: Ok. So basically it is not "other", it is make

internal/build/other/other.go
jochil: That `coverage` is a sanitizer is plausible... but a engine, IMO not so much. I think it is way more comprehensible for users when we frame an engine as something that takes/generates input and applies it to the fuzz test. In this scenario the engine is the `replayer` with `coverage` instrumentation

internal/cmd/run/run.go
adombeck: We can discuss whether we could support `cifuzz run` with other build systems on Windows, but that has nothing to do with this PR. IIUC, we don't support `cifuzz coverage` on Windows because coverage instrumentation is not supported by MSVC.

internal/build/other/other.go
adombeck: > I think it is way more comprehensible for users

This option is not exposed to users, it's only set to hardcoded values in our code

internal/build/other/other.go
adombeck: done

internal/build/other/other.go
adombeck: I now added this default case:
```go
default:
		panic(fmt.Sprintf("Invalid engine %q", opts.Engine))
```
Since the `Engine` option is not exposed to users and only set to hardcoded values in our code, IMO panicking is the right way to handle an unexpected value.

internal/cmd/run/run.go
adombeck: fixed

internal/cmd/run/run.go
adombeck: That won't work, because we call the `New` function of each command in the root command, no matter which command the user actually uses. Currently, we call `runCmd.New()` before `coverageCmd.New()`. If we bind the viper flag in the `New` function, the effect would be that e.g. the `build-command` flag of `run` is bound to viper first, and then the `build-command` flag of `coverage` can't be bound to viper as well. So if the user uses `cifuzz coverage`, the `build-command` flag of that command is not bound to viper. The only way I see is to bind the flags to viper in the `RunE` or `PreRunE` methods, because those are only executed if the command is actually executed.

internal/build/other/other.go
jochil: `:s/users/contributors/` ;)

internal/cmd/run/run.go
jochil: ok. thanks for the explanation

internal/cmd/coverage/coverage.go
adombeck: Ignoring shared objects in `.dSYM` directories fixed the test failure on macOS. 

> I think we should at least use the more fine-grained CMake function for the CMake build system case.

Can we reuse the `GetRuntimeDeps` function for that? Or does that include dependencies which are not binaries supported by llvm-cov?

internal/build/other/other.go
adombeck: I don't care much about how we call this option. I would be fine with "mode", but we're already using "engine" in  the cmake builder, where we use it in the build directory path and set the `CIFUZZ_ENGINE` cache variable to the same value and in the bundle command, where we set the `fuzzers.engine` in the artifacts metadata JSON file to `LLVM_COV` if the engine is `coverage` (changing that would require changes to the backend as well). I don't want to change all these places as part of this PR. Using "engine" here is consistent with the usage in our existing code. If you want to change that, I suggest we do that in another PR.

internal/cmd/coverage/coverage.go
fmeum: We should be able to. It will also return system deps, but passing them shouldn't break `llvm-cov`. If it does, we could extract the `bundle` logic that filters them out and use it here as well. But if the PR works as is, we could also create a follow-up ticket instead.

internal/build/other/other.go
fmeum: I tweaked the CMake integration in https://github.com/CodeIntelligenceTesting/cifuzz/pull/181 to represent coverage as `Engine: "replayer"` and `Sanitizers: []string{"coverage"}`. With that, I don't think there is any more reason to rename the fields.

adombeck: Ok, so now the usage of the `Engine` in `other.BuildOptions` is not consistent with the usage in `cmake.BuildOptions` anymore. How should we make it consistent again? We don't support specifying sanitizers with other build systems yet. Should we add a `Sanitizers` field which we ignore if `Engine == "libfuzzer"` and require to be set to `coverage` if `Engine == "replayer"`?

internal/cmd/coverage/coverage.go
adombeck: I created a follow-up ticket: https://code-intelligence.atlassian.net/browse/CLI-318

internal/build/other/other.go
jochil: I think this would be the most consistent way

fmeum: I also think this would be the best way. I would go as far as passing in `"address","undefined"` as `Sanitizers` into the other build system as well and failing if the sanitizers differ from that if `Engine == "libfuzzer"`.

adombeck: done

PR #180, opened 2022-07-28
build installer

florianGla:


tools/install/installer.go
```diff
@@ -17,113 +19,128 @@ import (
...
 }
 
 type Options struct {
-	InstallDir string
+	Version   string
+	TargetDir string
 }
 
-func NewInstaller(opts *Options) (*Installer, error) {
+const defaultVersion = "v0.0.1"
```
jochil, line 36: I think we should not fallback to a default version... IMO If no version number is provided the installer should fail

cmd/installer/main.go
```diff
@@ -1,36 +1,38 @@
...
 
 func main() {
-	opts := &install.Options{}
+	var installDir string
+	fs := &installer_bundle.Bundle
 
 	cmd := &cobra.Command{
 		Use:   "installer",
 		Short: "Install cifuzz",
 		Run: func(cmd *cobra.Command, args []string) {
```
jochil, line 23: maybe can do it similar to the other commands and use the `RunE` function and just return the error

tools/install/installer.go
florianGla: ok, the idea was to use a fallback version for local development builds, but we can think about a better solution

jochil: I would make it required for the `InstallationBundler` and for the development builds we can make sure that a version indicating a dev build is passed (via bundler)

florianGla: done

cmd/installer/main.go
florianGla: this works, but you have to keep in mind that resulting errors are printed together with the command help message

adombeck: Yeah, in the `cifuzz` command we use [`SilenceUsage`](https://github.com/CodeIntelligenceTesting/cifuzz/blob/CLI-305-fix-seed-corpus-not-used/internal/cmd/root/root.go#L38-L38) and [print the usage message ourselves](https://github.com/CodeIntelligenceTesting/cifuzz/blob/CLI-305-fix-seed-corpus-not-used/internal/cmd/root/root.go#L117-L127) to avoid that all errors result in the usage message being printed. A simpler way would be to only return those errors in `RunE` for which we want a usage message to be printed and handle other errors inside the `RunE` and exit immediately (e.g. via `os.Exit(1)`).

florianGla:
Shall I use cobra for the bundler for consistency reasons or are go flags fine?

cmd/installer/main.go
florianGla: I am not sure if RunE is appropriate since only extractBundle can return an error and that error should be handled inside the run function.

adombeck:
> Shall I use cobra for the bundler for consistency reasons or are go flags fine?

I would recommend using pflag. I don't expect that the bundler has subcommands, so I wouldn't recommend cobra.

florianGla:
> I would recommend using pflag. I don't expect that the bundler has subcommands, so I wouldn't recommend cobra.

Ok, does it make sense to also use pflag for the installer, which has only a target directory flag, or is cobra preferred?

cmd/installer/main.go
florianGla: done

pkg/install/installer.go
```diff
@@ -315,43 +304,148 @@ func (i *Installer) InstallCMakeIntegration() error {
...
+				return err
+			}
+
+			content, err := fs.ReadFile(bundleFs, path)
+			if err != nil {
+				return err
+			}
+
+			fileName := filepath.Join(targetDir, d.Name())
+			err = os.WriteFile(fileName, content, 0664)
```
adombeck, line 403:
```suggestion
			err = os.WriteFile(fileName, content, 0644)
```

pkg/install/installer.go
```diff
@@ -315,43 +304,148 @@ func (i *Installer) InstallCMakeIntegration() error {
...
+			if err != nil {
+				return err
+			}
+
+			// Make required files executable
+			for _, executableFile := range executableFiles {
+				if executableFile == path {
+					err = os.Chmod(fileName, 0775)
```
adombeck, line 411:
```suggestion
					err = os.Chmod(fileName, 0755)
```

pkg/install/installer.go
```diff
@@ -315,43 +304,148 @@ func (i *Installer) InstallCMakeIntegration() error {
...
+						return err
+					}
+				}
+			}
+		}
+
+		return nil
+	})
+	if err != nil {
+		return errors.WithStack(err)
```
adombeck, line 422: We should attach the stack trace as soon as possible, so we should attach it to the errors inside the `WalkDirFunc` and not here (else it doesn't tell us which of the functions in the `WalkDirFunc` returned an error).

pkg/install/installer.go
```diff
@@ -315,43 +304,148 @@ func (i *Installer) InstallCMakeIntegration() error {
...
+	executableFiles := []string{
+		"bin/cifuzz",
+		"bin/minijail0",
+		"lib/libminijailpreload.so",
+		"lib/process_wrapper",
+	}
+
+	targetDir, err := validateTargetDir(targetDir)
+	if err != nil {
+		return errors.WithStack(err)
```
adombeck, line 376: This attaches a duplicate stack trace, we already attach one in `validateTargetDir`.
```suggestion
		return err
```

pkg/install/installer.go
```diff
@@ -374,44 +468,36 @@ func findProjectDir() (string, error) {
...
+		}
+	} else {
+		installDir, err = filepath.Abs(installDir)
+		if err != nil {
+			return "", errors.WithStack(err)
+		}
+
+		exists, err := fileutil.Exists(installDir)
+		if err != nil {
+			return "", errors.WithStack(err)
```
adombeck, line 495: Duplicate stack trace
```suggestion
			return "", err
```

cmd/bundler/main.go
```diff
@@ -0,0 +1,38 @@
+package main
```
adombeck, line 1: What's the use case for executing this binary on its own, without running the installer afterwards? Could we merge it with the installer?

florianGla: We need two separate entries points, because the bundled files are embedded into the installer. It would be possible to let the bundler build the installer tho. The chaining of those two commands is currently handled in the Makefile. I am open for both possibilities.

adombeck: Ok, so both the "bundler" and the installer are only useful when called one after the other. How about we add a wrapper executable which does that for us, and hide the two other executables, e.g. in `tools/install`?

adombeck: I'm also fine with the Makefile handling that for us, but I would like to hide the two commands since they are not useful for users/contributors, and IMO `cmd/` should only contains commands which are useful for users/contributors.

florianGla: Great idea

cmd/installer/main.go
```diff
@@ -1,40 +1,47 @@
...
 		Use:   "installer",
 		Short: "Install cifuzz",
-		Run: func(cmd *cobra.Command, args []string) {
-			installer, err := install.NewInstaller(opts)
+		RunE: func(cmd *cobra.Command, args []string) error {
+			err := install.ExtractBundle(installDir, fs)
 			if err != nil {
 				log.Error(err, err.Error())
-				os.Exit(1)
+				return cmdutils.ErrSilent
```
adombeck, line 27: This error is currently printed:
```
$ build/bin/installer 
❌ Install directory '/home/adrian/cifuzz' already exists. Please remove it to continue.
Error: SilentError
Usage:
  installer [flags]

Flags:
  -h, --help                 help for installer
  -i, --install-dir string   The directory to install cifuzz in (default "~/cifuzz")
```

pkg/install/installer.go
```diff
@@ -374,44 +468,36 @@ func findProjectDir() (string, error) {
...
+		}
+
+		exists, err := fileutil.Exists(installDir)
+		if err != nil {
+			return "", err
+		}
+		if exists {
+			return "", errors.Errorf("Install directory '%s' already exists. Please remove it to continue.", installDir)
+		}
```
adombeck, line 499: I don't think we want the bundler to fail if the bundle directory already exists:

```
make installer
go run cmd/bundler/main.go --version dev && go build -tags installer -o build/bin/installer cmd/installer/main.go && rm -r installer-bundle/bundle
❌ Install directory '/home/adrian/cifuzz/installer-bundle/bundle' already exists. Please remove it to continue.
exit status 1
make: *** [Makefile:40: installer] Error 1
```

tools/install/bundler/bundler.go
```diff
@@ -0,0 +1,57 @@
...
+	version := flags.StringP("version", "v", "dev", "the target version of cifuzz")
+	helpRequested := flags.BoolP("help", "h", false, "")
+
+	if err := flags.Parse(os.Args); err != nil {
+		log.Error(err, err.Error())
+		os.Exit(1)
+	}
+
+	if *helpRequested {
+		log.Printf("Usage of %s:\n", "bundler")
```
jochil, line 24:
```suggestion
		log.Print("Usage of bunder")
```

tools/install/installer/installer.go
```diff
@@ -0,0 +1,38 @@
...
+	installDir := flags.StringP("install-dir", "i", "~/cifuzz", "The directory to install cifuzz in")
+	helpRequested := flags.BoolP("help", "h", false, "")
+
+	if err := flags.Parse(os.Args); err != nil {
+		log.Error(err, err.Error())
+		os.Exit(1)
+	}
+
+	if *helpRequested {
+		log.Printf("Usage of %s:\n", "cifuzz installer")
```
jochil, line 24:
```suggestion
		log.Print("Usage of cifuzz installer:")
```

tools/install/installer/installer.go
```diff
@@ -0,0 +1,38 @@
...
+		flags.PrintDefaults()
+		os.Exit(0)
+	}
+
+	fs := &embed.Bundle
+	err := install.ExtractBundle(*installDir, fs)
+	if err != nil {
```
jochil, line 31:
```suggestion
	if err := install.ExtractBundle(*installDir, fs); err != nil {
```
Would be consistent with the error handling above in this file

tools/install/bundler/bundler.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+
+	projectDir, err := install.FindProjectDir()
+	if err != nil {
+		log.Error(err, err.Error())
+		os.Exit(1)
+	}
+	targetDir := filepath.Join(projectDir, "tools", "install", "bundler", "embed", "bundle")
+	err = os.RemoveAll(targetDir)
+	if err != nil {
```
jochil, line 36:
```suggestion
	if err := os.RemoveAll(targetDir); err != nil {
```

tools/install/bundler/bundler.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+
+	bundler, err := install.NewInstallationBundler(opts)
+	if err != nil {
+		log.Error(err, err.Error())
+		os.Exit(1)
+	}
+
+	err = bundler.BuildCIFuzzAndDeps()
+	if err != nil {
```
jochil, line 53:
```suggestion
	if err := bundler.BuildCIFuzzAndDeps(); err != nil {
```

tools/install/bundler/embed/embed.go
```diff
@@ -0,0 +1,7 @@
+//go:build !installer
```
jochil, line 1: it would be nice having a short explanation why we doing the switch with the two embed files... maybe as a comment

integration-tests/cmake/cmake_test.go
```diff
@@ -30,20 +30,40 @@ func TestIntegration_CMake_InitCreateRunCoverageBundle(t *testing.T) {
...
 	require.NoError(t, err)
-	defer installer.Cleanup()
-	err = os.Setenv("CMAKE_PREFIX_PATH", installer.InstallDir)
+
+	opts := install.Options{Version: "dev", TargetDir: targetDir}
+	bundler, err := install.NewInstallationBundler(opts)
+	require.NoError(t, err)
+	err = bundler.BuildCIFuzzAndDeps()
+	require.NoError(t, err)
+	defer bundler.Cleanup()
```
jochil, line 45: I would move the `defer` statement a little bit higher... if `bundler.BuildCIFuzzAndDeps` fails there will be no `bundler.Cleanup`

integration-tests/make/make_test.go
```diff
@@ -29,12 +30,32 @@ func TestIntegration_Make_RunCoverage(t *testing.T) {
...
 	require.NoError(t, err)
-	defer installer.Cleanup()
-	cifuzz := installer.CIFuzzExecutablePath()
+
+	opts := install.Options{Version: "dev", TargetDir: targetDir}
+	bundler, err := install.NewInstallationBundler(opts)
+	require.NoError(t, err)
+	err = bundler.BuildCIFuzzAndDeps()
+	require.NoError(t, err)
+	defer bundler.Cleanup()
```
jochil, line 45: > I would move the defer statement a little bit higher... if bundler.BuildCIFuzzAndDeps fails there will be no bundler.Cleanup

same here

pkg/minijail/process_wrapper/process_wrapper_test.go
```diff
@@ -13,23 +13,23 @@ import (
...
+	err = bundler.BuildProcessWrapper()
 	if err != nil {
 		log.Fatalf("%+v", err)
 	}
-	runfiles.Finder = runfiles.RunfilesFinderImpl{InstallDir: installer.InstallDir}
+	runfiles.Finder = runfiles.RunfilesFinderImpl{InstallDir: bundler.TargetDir}
 
 	res := m.Run()
-	installer.Cleanup()
+	bundler.Cleanup()
```
jochil, line 32: there is a chance that this will not be called... is this on intent?

jochil (approved):

fmeum (approved):

PR #181, opened 2022-07-28
cmake: Represent coverage builds as a combination of engine and sanitizer

fmeum:
Makes it more clear that coverage builds essentially just use the
replayer (albeit with a few more, technical flags) and apply coverage
instrumentation.


jochil (approved):
lgtm

PR #184, opened 2022-07-29
release pipelines

jochil:
This change includes

- A new custom action for generating a pseudo version based on the commit and the previous release
- A new composite action to have a reusable vm setup for the specific operating systems
- A reusable workflow for build the installer (for all supported operating systems) and uploading them as artifacts
- A pipeline generating a new pseudo version and building/uploading the installer (executed on every push to `main`)
- A pipeline triggered by a new tag (following the pattern `v*.*.*`) that also builds the installer and then uploads them to a draft release

The main reasons for the custom action were that it is very easy to access the github api (for the latest release), the commit that triggered the pipeline and the seamless execution inside of github action

This is whats the draft release looks like:
image

EDIT: I renamed the artifacts in the release... now there is no version included and we can use the `/releases/latest/download/asset-name.zip` path


.github/actions/version/action.yml
```diff
@@ -0,0 +1,15 @@
+name: 'cifuzz version'
+description: 'helps generating (prerelease) version numbers for cifuzz'
+inputs:
+  token:
+    description: 'Authorized secret GitHub Personal Access Token. Defaults to github.token'
+    required: false
+    default: ${{ github.token }}
```
fmeum, line 7: Does the action really need this token? If I understand the permission model correctly, this token allows the action to do essentially anything when all it really needs is read-only access to public information (latest releases). In fact, all it would really need is to get the latest tag from the Git history.

Do you know whether this can be restricted further?

.github/workflows/pipeline_installer.yml
```diff
@@ -0,0 +1,32 @@
+name: Installer Pipeline
+
+on:
+  push:
+    # TODO: just run it on main branch
```
fmeum, line 5: Should change this before the merge.

.github/workflows/workflow_installer.yml
```diff
@@ -0,0 +1,36 @@
+on:
+  workflow_call:
+    inputs:
+      version:
+        required: true
+        type: string
+
+jobs:
+  installer:
+    name: build + upload (${{ matrix.os }})
```
fmeum, line 10: If this is the only identifier that is shown, we should probably make it less generic.

.github/actions/version/action.yml
jochil: It seems like the function from github actions toolkit requires the token, even if it is not necessary to access the API for public endpoints/information. I thought using it is just more straightforward than finding a workaround around the authentication or using another/own client.

But restricting the token in general is a good idea. So I generally changed the token (for all workflows) to read permissions. Now we explicitly have to allow write access in the workflow files.

.github/workflows/workflow_installer.yml
jochil: done

florianGla (approved):

fmeum (commented):
Thanks, this looks really good!

I left a single comment.

.github/workflows/workflow_installer.yml
```diff
@@ -0,0 +1,34 @@
...
+        required: true
+        type: string
+
+jobs:
+  installer:
+    name: build installer + upload as artifact (${{ matrix.os }})
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [macos-latest, windows-latest, ubuntu-latest]
```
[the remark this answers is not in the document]
jochil: Good point. I created a separate task for it CLI-321. This way we can already use the installer (pipelines) in the wild

fmeum (approved):

.github/workflows/workflow_installer.yml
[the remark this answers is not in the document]
jochil: extremely good point :) ... the version we released with the installer is not running with ubuntu18:04
It fails when running our cmake project:
```
/home/runner/cifuzz/bin/minijail0: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.28' not found (required by /home/runner/cifuzz/bin/minijail0)
```
@florianGla FYI

PR #188, opened 2022-08-02
Add CMake presets for coverage and regression tests

fmeum:


jochil (approved):
I am not an expert in CMake/C but looks good to me

PR #192, opened 2022-08-03
readme: Remove clang from MacOS prerequisites

florianGla:


fmeum (commented):
Could you explain why it isn't needed? I remember AppleClang being weird and not shipping libfuzzer.

florianGla:
> Could you explain why it isn't needed? I remember AppleClang being weird and not shipping libfuzzer.

Oh ok, Jochen mentioned to me, that clang is not available in brew. I'm not sure if we should try to find another command or write a note message, saying that clang has to be installed manually for MacOS.

jochil:
> Oh ok, Jochen mentioned to me, that clang is not available in brew. I'm not sure if we should try to find another command or write a note message, saying that clang has to be installed manually for MacOS.

It is because clang is included in the llvm homebrew package... if I am not completely wrong

fmeum (approved):

PR #193, opened 2022-08-03
ci(installer): Use Windows Server 2022 os runner version

florianGla:
Visual Studio 17 2022 or higher is required for building the installer.


jochil:
I think this is not necessary... the "Visual Studio 17 2022" requirement only applies to executing cifuzz. Not building the installer.
Sorry If I was not clear with this before.

Only reason to change would be consistently using windows-2022 for all our pipelines

florianGla:
Ok, shall I mention this requirement for windows in the README?

jochil:
> Ok, shall I mention this requirement for windows in the README?

I would like that

PR #196, opened 2022-08-03
cmake: Fix cmake_test cmake installation

florianGla:


tools/cmake/cmake_test.go
```diff
@@ -41,6 +51,12 @@ func TestMain(m *testing.M) {
...
 
+	err = os.Setenv("CMAKE_PREFIX_PATH", installDir)
+	if err != nil {
+		bundler.Cleanup()
+		log.Fatalf("Failed to install CMake integration: %+v", err)
+	}
+
 	m.Run()
 
 	bundler.Cleanup()
```
fmeum, line 62: Could we use `defer bundler.Cleanup()` instead? It seems to be called on all paths.

tools/cmake/cmake_test.go
```diff
@@ -41,6 +51,12 @@ func TestMain(m *testing.M) {
 		log.Fatalf("Failed to install CMake integration: %+v", err)
 	}
 
+	err = os.Setenv("CMAKE_PREFIX_PATH", installDir)
```
fmeum, line 54: Could you add a comment saying that we don't test that our CMake package is registered correctly here, but instead do so in the CMake integration test?

tools/cmake/cmake_test.go
florianGla: I think it is not possible, because `log.Fatalf()` calls `os.Exit(1)`, which prevents `defer bundler.Cleanup()` from beeing executed.

fmeum (approved):

PR #199, opened 2022-08-04
feat(coverage): display coverage metrics when running `cifuzz coverage`

jochil:
Changes:
* Data structures to store coverage information
* lcov parser to convert lcov tracefile report into internal data structures
* Nice output for the coverage metrics
* opens HTML report in Browser

this command can be used to quickly test the output 
`go run cmd/cifuzz/main.go -C ./examples/cmake --use-sandbox=false coverage my_fuzz_test`


internal/cmd/coverage/coverage.go
```diff
@@ -357,11 +364,28 @@ func (c *coverageCmd) generateHTMLReport(fuzzTestExecutable string, runtimeDeps
...
 	output, err := cmd.Output()
 	if err != nil {
-		return cmdutils.WrapExecError(errors.WithStack(err), cmd)
+		return "", cmdutils.WrapExecError(errors.WithStack(err), cmd)
+	}
+	return string(output), nil
+}
+
+func (c *coverageCmd) generateLcovReport(fuzzTestExecutable string, runtimeDeps []string) (string, error) {
+	output, err := c.runLlvmCov([]string{"export", "-format=lcov", "-summary-only"}, fuzzTestExecutable, runtimeDeps)
```
fmeum, line 373: Could you explain why you decided to emit lcov and then parse it? The `json` output format of `llvm-cov` also contains summary information, see https://stackoverflow.com/a/56792192/297261.

jochil: My thought was that we might encounter `lcov` again when working on coverage support for other platforms...  and the format was quite simple to parse so I gave it a shot.
But I am open to switch to the json output

fmeum: While it is true that essentially everything can convert to `lcov`, I think that almost everything can also convert to a summary directly. The main advantage of `lcov` is that we can merge it across languages, so it does serve a purpose. 

The potential problem I see is that the JSON summary, the HTML report and the summary derived from the `lcov` report could differ. Maybe we should compare the three for some test cases to form an opinion?

jochil: sounds reasonable... do you think it should be part of this task/change or should we create a new one?

fmeum: I think that we should do it now - it's unlikely that the JSON summary will differ from the HTML report (both use raw llvm-cov data as their basis), so if it turns out that the `lcov` result is different, we would have to parse the JSON instead.

jochil: Ok. I will come up with a testcase. 
I changed this PR to "draft" meanwhile

florianGla: I think it would be an advantage to also generate a lcov report, which can be easily used by the coverage gutters extension in VSC. Otherwise we could provide a script for VSC users, which seems redundant.

fmeum: With @florianGla's discovery that LCOV generation is best done via `cifuzz coverage` in VSC, we definitely need the functionality to generate a report. At that point, we might as well use it as the basis of our coverage summaries. 

@jochil We should still follow through on the comparison, but I wouldn't say it's blocking this PR anymore.

jochil: Ok. I am on the test... but I created a new task (CLI-342) for it. I "undrafted" this PR

jochil: lcov (+ our package) are working perfectly fine... I documented the results in the ticket.

fmeum (approved):

PR #201, opened 2022-08-05
fix(e2e): add glibc to arch image

jochil:
... as it is needed for the github checkout action


.github/workflows/pipeline_e2e_source.yml
```diff
@@ -24,7 +24,8 @@ jobs:
       - name: Install dependencies with pacman
         if: contains(matrix.container, 'archlinux')
         # please make sure this is in sync with the README.md / installation docs
-        run: pacman -Sy --noconfirm git make cmake clang llvm go libcap
+        # glibc was added because the github checkout action is not working without it
```
fmeum, line 27: I would prefer a more specific comment here: We can say that `node` requires `glibc` and is used to run the action (assuming that's correct).

fmeum (approved):

.github/workflows/pipeline_e2e_source.yml
jochil: Yes. That is correct... I updated the comment

PR #206, opened 2022-08-05
runner: Add build dir to minijail bindings

fmeum:
Without this binding, `cifuzz run parser_fuzz_test` fails when executed
from within `.cifuzz-build/libfuzzer/address+undefined/src/parser` in
`tools/cmake/testdata`.


fmeum:
@adombeck I discovered this issue when using the fuzzing preset introduced in #205 in CLion. The directory in the commit message is the cwd used by CLion to execute the CMake target.

I don't understand why the minijail behavior depends on the working directory, so please let me know if my way to fix this is wrong.

adombeck (approved):

PR #211, opened 2022-08-10
Fix install.sh

adombeck:


jochil (approved):
thanks... shell script, always new stuff to learn :)

PR #213, opened 2022-08-11
Generate lcov coverage report

florianGla:


internal/cmd/coverage/coverage.go
```diff
@@ -158,21 +170,39 @@ func (c *coverageCmd) run() error {
...
+		lcovReport, err := c.generateLcovReport(buildResult.Executable, buildResult.RuntimeDeps, false)
+		if err != nil {
+			return err
+		}
+
+		reportPath, err := c.writeReportToFile(buildResult.Executable, lcovReport)
+		if err != nil {
+			return err
+		}
+		log.Successf("Created coverage LCOV report: %s", reportPath)
```
jochil, line 208: I think we can move this block and the similar one above, out of the switch block... due to the default value there should be always a report to write to a file

jochil (approved):

PR #215, opened 2022-08-11
Canonicalize BuildResult members

fmeum:


internal/cmd/bundle/bundle.go
```diff
@@ -293,6 +293,14 @@ func assembleArtifacts(fuzzTest string, buildResult *build.Result, projectDir st
 	buildArtifactsPrefix := filepath.Join(fuzzTestPrefix(fuzzTest, buildResult), "bin")
 
 	// Add the fuzz test executable.
+	ok, err := fileutil.IsBelow(fuzzTestExecutableAbsPath, buildResult.BuildDir)
+	if err != nil {
+		err = errors.WithStack(err)
```
adombeck, line 298: This attaches a duplicate stack trace
```suggestion
```

internal/cmd/bundle/bundle.go
```diff
@@ -293,6 +293,14 @@ func assembleArtifacts(fuzzTest string, buildResult *build.Result, projectDir st
...
+		return
+	}
+	if !ok {
+		err = errors.Errorf("fuzz test executable (%s) is not below build directory (%s)", fuzzTestExecutableAbsPath, buildResult.BuildDir)
```
adombeck, line 302: I'm surprised the linter didn't catch that the value assigned to `err` here is never used
```suggestion
		err = errors.Errorf("fuzz test executable (%s) is not below build directory (%s)", fuzzTestExecutableAbsPath, buildResult.BuildDir)
		return
```

adombeck (approved):

PR #219, opened 2022-08-12
feat: check command dependencies on demand

jochil:
This change includes the package `dependencies` that holds definitions of the cifuzz dependencies (incl. versions) and offers functionality to perform checks.
Also all commands with dependencies were adjusted to check them at the beginning of the `run` function.


CODEOWNERS
```diff
@@ -22,3 +22,4 @@
 /docs @jochil 
 /examples @jochil
 /pkg/coverage @jochil
+/pgk/dependencies @jochil
```
fmeum, line 25:
```suggestion
/pkg/dependencies @jochil
```

pkg/dependencies/deps.go
```diff
@@ -0,0 +1,87 @@
...
+	"code-intelligence.com/cifuzz/pkg/runfiles"
+)
+
+type Key string
+
+const (
+	CLANG           Key = "clang"
+	CMAKE               = "cmake"
+	LLVM_COV            = "llvm-cov"
+	LLVM_SYMPOLIZER     = "llvm-symbolizer"
```
fmeum, line 17:
```suggestion
	LLVM_SYMBOLIZER     = "llvm-symbolizer"
```

pkg/dependencies/deps.go
```diff
@@ -0,0 +1,87 @@
...
+	Key        Key
+	MinVersion semver.Version
+	// these fields are used to implement custom logic to
+	// retrieve version or installation information for the
+	// specific dependency
+	GetVersion func(*Dependency) (*semver.Version, error)
+	Installed  func(*Dependency) bool
+}
+
+// Comparse MinVersion against GetVersion
```
fmeum, line 34:
```suggestion
// Compares MinVersion against GetVersion
```

pkg/dependencies/definitions_test.go
```diff
@@ -0,0 +1,51 @@
...
+func TestResetDefault(t *testing.T) {
+	oldDefault := Default
+	Default = Dependencies{}
+	ResetDefault()
+	assert.Len(t, Default, 5)
+
+	newKeys := maps.Keys(Default)
+	for key, _ := range oldDefault {
+		assert.Contains(t, newKeys, key)
+
```
fmeum, line 20:
```suggestion
```

pkg/dependencies/definitions.go
```diff
@@ -0,0 +1,104 @@
...
+	deps, err := Define([]Key{
+		CLANG,
+		CMAKE,
+		LLVM_COV,
+		LLVM_PROFDATA,
+		LLVM_SYMPOLIZER,
+	})
+
+	if err != nil {
+		panic("Unable do define default dependencies")
```
fmeum, line 26:
```suggestion
		panic("Unable to define default dependencies")
```

pkg/dependencies/definitions.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+	Default = deps
+}
+
+// Defines a set of dependencies
+func Define(keys []Key) (Dependencies, error) {
+	deps := Dependencies{}
+	for _, key := range keys {
+		if dep, found := all[key]; found {
+			// make a copy of the dependency to be able to modify init
```
fmeum, line 36:
```suggestion
			// make a copy of the dependency to be able to modify it
```

internal/testutil/example_project.go
```diff
@@ -0,0 +1,29 @@
...
+
+	"github.com/otiai10/copy"
+	"github.com/pkg/errors"
+
+	"code-intelligence.com/cifuzz/util/testutil"
+)
+
+// Creates a copy of the the CMake example project in a temp dir and
+// change the workding dir to the new path
+func ChdirToClonedCmakeExamleProject(prefix string) (string, error) {
```
fmeum, line 15:
```suggestion
func ChdirToClonedCmakeExampleProject(prefix string) (string, error) {
```

internal/testutil/example_project.go
```diff
@@ -0,0 +1,29 @@
...
+	"path/filepath"
+	"runtime"
+
+	"github.com/otiai10/copy"
+	"github.com/pkg/errors"
+
+	"code-intelligence.com/cifuzz/util/testutil"
+)
+
+// Creates a copy of the the CMake example project in a temp dir and
```
fmeum, line 13:
```suggestion
// Creates a copy of the CMake example project in a temp dir and
```

internal/testutil/example_project.go
```diff
@@ -0,0 +1,29 @@
...
+// change the workding dir to the new path
```
fmeum, line 14:
```suggestion
// change the working dir to the new path
```

internal/cmd/bundle/bundle.go
```diff
@@ -279,6 +290,14 @@ func (c *bundleCmd) buildAllVariants() ([]map[string]*build.Result, error) {
 	return allVariantBuildResults, nil
 }
 
+func (c *bundleCmd) checkDependencies() (bool, error) {
+	deps := []dependencies.Key{dependencies.CLANG, dependencies.LLVM_SYMBOLIZER}
```
fmeum, line 294: I don't think we need the llvm-symbolizer for `bundle`, only for `run`.

internal/cmd/coverage/coverage_test.go
```diff
@@ -1,15 +1,108 @@
...
+func TestClangMissing(t *testing.T) {
+	deps := dependencies.CreateTestDeps(t, []dependencies.Key{
+		dependencies.CLANG, dependencies.LLVM_SYMBOLIZER, dependencies.LLVM_COV, dependencies.LLVM_PROFDATA, dependencies.CMAKE,
+	})
+	deps[dependencies.CLANG].Installed = func(d *dependencies.Dependency) bool {
+		return false
+	}
+
+	// clone the example project because this command needs to parse an actual
+	// project config... if there is no it will fail before the dependency check
```
fmeum, line 51:
```suggestion
	// project config... if there is none it will fail before the dependency check
```

internal/cmd/coverage/coverage_test.go
```diff
@@ -1,15 +1,108 @@
...
+func TestCMakeMissing(t *testing.T) {
+	deps := dependencies.CreateTestDeps(t, []dependencies.Key{
+		dependencies.CLANG, dependencies.LLVM_SYMBOLIZER, dependencies.LLVM_COV, dependencies.LLVM_PROFDATA, dependencies.CMAKE,
+	})
+	deps[dependencies.CMAKE].Installed = func(d *dependencies.Dependency) bool {
+		return false
+	}
+
+	// clone the example project because this command needs to parse an actual
+	// project config... if there is no it will fail before the dependency check
```
fmeum, line 73:
```suggestion
	// project config... if there is none it will fail before the dependency check
```

internal/cmd/coverage/coverage_test.go
```diff
@@ -1,15 +1,108 @@
...
+func TestLlvmCovVersion(t *testing.T) {
+	deps := dependencies.CreateTestDeps(t, []dependencies.Key{
+		dependencies.CLANG, dependencies.LLVM_SYMBOLIZER, dependencies.LLVM_COV, dependencies.LLVM_PROFDATA, dependencies.CMAKE,
+	})
+	deps[dependencies.LLVM_COV].GetVersion = func(d *dependencies.Dependency) (*semver.Version, error) {
+		return semver.NewVersion("0.0.0")
+	}
+
+	// clone the example project because this command needs to parse an actual
+	// project config... if there is no it will fail before the dependency check
```
fmeum, line 96:
```suggestion
	// project config... if there is none it will fail before the dependency check
```

internal/cmd/reload/reload_test.go
```diff
@@ -1,15 +1,92 @@
...
+
+	conf := config.NewConfig()
+	conf.BuildSystem = config.BuildSystemCMake
+
+	_, err := cmdutils.ExecuteCommand(t, New(conf), os.Stdin)
+	require.Error(t, err)
+
+	output, err := io.ReadAll(testOut)
+	require.NoError(t, err)
+	assert.Contains(t, string(output), "cmake has invalid version")
```
fmeum, line 91: Great tests!

pkg/dependencies/definitions.go
```diff
@@ -0,0 +1,104 @@
...
+
+type Dependencies map[Key]*Dependency
+
+var Default Dependencies
+
+func init() {
+	ResetDefault()
+}
+
+func ResetDefault() {
```
fmeum, line 16: If this is only used in tests, could we call it `ResetDefaultsForTestsOnly`?

pkg/dependencies/version.go
```diff
@@ -0,0 +1,70 @@
...
+)
+
+// returns the currently installed clang version
+func clangVersion(dep *Dependency) (*semver.Version, error) {
+	path, err := dep.finder.ClangPath()
+	if err != nil {
+		return nil, err
+	}
+
+	re := regexp.MustCompile(`(?m)^clang version (?P<version>\d+.\d+.\d+)$`)
```
fmeum, line 19: On my system, the output is `Ubuntu clang version 14.0.0-1ubuntu1`. How about:
```suggestion
	re := regexp.MustCompile(`(?m)clang version (?P<version>\d+.\d+.\d+)$`)
```

pkg/dependencies/version.go
```diff
@@ -0,0 +1,70 @@
...
+	version, err := parseCommandVersionWithRegex(path, []string{"--version"}, re, dep.Key)
+	if err != nil {
+		return nil, errors.WithStack(err)
```
fmeum, line 22: If we fail to detect the version, maybe we should assume it to be valid instead? I'm worried about breaking third-party builds of clang and/or CMake purely because our version parsing isn't lenient enough.

pkg/dependencies/version.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+	return version, nil
+}
+
+// helper for parsing the --version output for different llvm tools,
+// for example llvm-cov, llvm-symbolizer
+func llvmVersion(path string, dep *Dependency) (*semver.Version, error) {
+	re := regexp.MustCompile(`(?m)^\s*LLVM version (?P<version>\d+.\d+.\d+)$`)
```
fmeum, line 30: On my system: `Ubuntu LLVM version 14.0.0`
```suggestion
	re := regexp.MustCompile(`(?m)\s*LLVM version (?P<version>\d+.\d+.\d+)$`)
```

pkg/dependencies/version.go
```diff
@@ -0,0 +1,70 @@
...
+	return version, nil
+}
+
+func cmakeVersion(dep *Dependency) (*semver.Version, error) {
+	path, err := exec.LookPath("cmake")
+	if err != nil {
+		return nil, errors.WithStack(err)
+	}
+
+	re := regexp.MustCompile(`(?m)^\s*cmake version (?P<version>\d+.\d+.\d+)$`)
```
fmeum, line 45: I'm not sure whether CMake 3.22 isn't printing out `cmake version 3.22` rather than `cmake version 3.22.0`. Maybe we should always make the patch version optional?

internal/cmd/bundle/bundle.go
jochil: ok... the symbolizer is only used for the fuzzing run, or?

internal/cmd/reload/reload_test.go
jochil: thanks

pkg/dependencies/version.go
jochil: I checked this and I got this `cmake version 3.24.0`
But in general I think it would be a good idea... should we do it for all version checks?

pkg/dependencies/version.go
jochil: I think you are right... we can still give a warning, but at least let it pass. I would add it to `deps.CheckVersion()` to have it at a more "central" location

pkg/dependencies/definitions.go
jochil: It is also used in the `init()` of this file to actual set the Defaults

fmeum: Is there a situation in which non-test functions in other packages may want to call it? If not, I would still prefer the rename (or have it be internal and expose it via a public `ResetDefaultsForTestsOnly`.

pkg/dependencies/version.go
fmeum: Yes, that sounds like a reasonable approach.

pkg/dependencies/version.go
fmeum: Yes, I would say so. That's a case where being more lenient generally pays off.

pkg/dependencies/definitions.go
jochil: The latter sounds good to me... this would avoid having `ResetDefaultsForTestsOnly` called in the `init()`

pkg/dependencies/deps.go
```diff
@@ -0,0 +1,88 @@
...
+	// specific dependency
+	GetVersion func(*Dependency) (*semver.Version, error)
+	Installed  func(*Dependency) bool
+}
+
+// Compares MinVersion against GetVersion
+func (dep *Dependency) CheckVersion() bool {
+	currentVersion, err := dep.GetVersion(dep)
+	if err != nil {
+		log.Errorf(err, "Unable to get current version for %s", dep.Key)
```
fmeum, line 38: IMO we should use `Warningf` here if we don't fail.

pkg/dependencies/deps.go
```diff
@@ -0,0 +1,88 @@
...
+		// we want to be lenient if we were not able to extract the version
+		return true
+	}
+
+	if currentVersion.Compare(&dep.MinVersion) == -1 {
+		log.Warnf("Dependency %s has invalid version %s (min. %s)", dep.Key, currentVersion.String(), dep.MinVersion.String())
```
fmeum, line 44: How about 
```suggestion
		log.Warnf("cifuzz requires %s %s or higher, have %s", dep.Key, dep.MinVersion.String(), currentVersion.String())
```

pkg/dependencies/deps.go
```diff
@@ -0,0 +1,88 @@
...
+
+	"code-intelligence.com/cifuzz/pkg/log"
+	"code-intelligence.com/cifuzz/pkg/runfiles"
+)
+
+type Key string
+
+const (
+	CLANG           Key = "clang"
+	CMAKE               = "cmake"
```
fmeum, line 15: Are these also of type `Key` implicitly?

pkg/dependencies/deps.go
```diff
@@ -0,0 +1,88 @@
...
+	if currentVersion.Compare(&dep.MinVersion) == -1 {
+		log.Warnf("Dependency %s has invalid version %s (min. %s)", dep.Key, currentVersion.String(), dep.MinVersion.String())
+		return false
+	}
+	return true
+}
+
+func (dep *Dependency) Ok() bool {
+	if !dep.Installed(dep) {
+		log.Warnf("Dependency %s is not installed", dep.Key)
```
fmeum, line 52: I would prefer:
```suggestion
		log.Warnf("cifuzz requires %s, but it is not installed", dep.Key)
```

We could even pass in the command name as an argument and then print `cifuzz foo requires ...` instead.

pkg/dependencies/deps.go
fmeum: Wouldn't we see the warning just like we would the error? It just that if we show an error to the end user rather then a warning, they should be able to assume it's fatal.

pkg/dependencies/deps.go
jochil: No... untyped constants are a little bit counterintuitive
```
type Key string
const (
	C1 Key = "c1"
	C2
	C3 = "c3"
)
func main() {
	fmt.Printf("C1 %T\n", C1) // C1 main.Key
	fmt.Printf("C2 %T\n", C2) // C2 main.Key
	fmt.Printf("C3 %T\n", C3) // C3 string
}
```
It works as `Key` is a string type... but I will make it more explicit and add the types.

pkg/dependencies/deps.go
jochil: so `log.Warnf("Unable to get current version for %s, message: %+v", dep.Key, err)` ?

fmeum: Yes, that's what I meant.

fmeum (approved):

pkg/dependencies/deps.go
jochil: done that

PR #220, opened 2022-08-15
coverage: Support not opening browser

adombeck:
Copy the behavior of `go tool cover`, which is to open the browser iff
the `--output` flag is not specified (in which case the report is
created in a temporary directory).


florianGla:
That could clash with #213. I will close this pr, if the checks pass.

adombeck:
> That could clash with #213. I will close this pr, if the checks pass.

Yes, there are some minor conflicts between this and #213. No need to close the PR though, we can just adjust it.

adombeck:
I fixed the conflicts

internal/cmd/coverage/coverage.go
```diff
@@ -387,6 +388,53 @@ func (c *coverageCmd) indexRawProfile(fuzzTestExecutable string) error {
...
+	}
+
+	err = os.WriteFile(outputPath, []byte(report), 0644)
+	if err != nil {
+		return errors.WithStack(err)
+	}
+
+	// Open the browser iff the output format is HTML and no output path
+	// was specified
+	openBrowser := c.opts.OutputFormat == "html" && c.opts.OutputPath == ""
```
jochil, line 415: `c.opts.OutputFormat == "html"` is unnecessary as it is already in `run()` before calling this function. But if it was intended as some kind of defensive programming, we can keep it

jochil (approved):

internal/cmd/coverage/coverage.go
adombeck: It's a leftover from the refactoring I had to do after #213 was merged, I'll remove it

adombeck: done

PR #223, opened 2022-08-15
Add command to list and view findings

adombeck:


internal/cmd/finding/finding_cmd.go
```diff
@@ -0,0 +1,62 @@
...
+
+func New() *cobra.Command {
+	opts := &cmdOpts{}
+	findingCmd := &cobra.Command{
+		Use:     "finding",
+		Aliases: []string{"findings"},
+		Short:   "List and show findings",
+		Args:    cobra.MaximumNArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return run(cmd, args, opts)
```
jochil, line 38: what do you think about having a `findingCmd` struct as we have in all newer commands... I think this would be more consistent

jochil: For example I just changed it for the `create` command: https://github.com/CodeIntelligenceTesting/cifuzz/pull/219/commits/36073197f3d8b5cf0c96c5c2e5e5d0fb6287c7e0

adombeck:
@CodeIntelligenceTesting/cifuzz-devs: I implemented Variant 2 from https://code-intelligence.atlassian.net/wiki/spaces/CI/pages/123142145/CLI+Workflow here, because that was the most favored variant IIRC. If you prefer another variant, please say so and we can vote on it (changing the code shouldn't be much effort and I will do that of course).

internal/cmd/finding/finding_cmd.go
adombeck: done

pkg/finding/finding.go
```diff
@@ -148,3 +151,49 @@ func (f *Finding) moveInputFile(projectDir, findingDir string) error {
...
+		return []*Finding{}, nil
+	}
+	if err != nil {
+		return nil, errors.WithStack(err)
+	}
+
+	var res []*Finding
+	for _, e := range entries {
+		jsonPath := filepath.Join(findingsDir, e.Name(), nameJsonFile)
+		bytes, err := os.ReadFile(jsonPath)
```
fmeum, line 168: Could this become a bottleneck with many findings? I think it would be better to read and parse the JSONs only when the details of a particular finding are actually needed, not already in `ListFindings`.

pkg/finding/finding.go
```diff
@@ -148,3 +151,49 @@ func (f *Finding) moveInputFile(projectDir, findingDir string) error {
...
+			return nil, errors.WithStack(err)
+		}
+		res = append(res, &f)
+	}
+
+	return res, nil
+}
+
+func LoadFinding(projectDir, findingName string) (*Finding, error) {
+	findingDir := filepath.Join(projectDir, nameFindingDir, findingName)
```
fmeum, line 184: This seems to duplicate parts of `ListFindings`.

pkg/finding/finding.go
adombeck: Valid point, but since we want to sort the findings by date (which I'm working on in [a follow-up PR](#224)), we need to parse them anyway. And I think we want to sort the findings by date in all current use cases of the `ListFindings` function (that's the command which lists the findings and the findings completion function).

pkg/finding/finding.go
adombeck: The error handling is different though. Lets first finish discussing https://github.com/CodeIntelligenceTesting/cifuzz/pull/223#pullrequestreview-1074111139 to see if we keep unmarshalling findings in `ListFindings` as well, then I could try to extract a function here.

adombeck: Nevermind, I changed the error handling a bit and am now reusing `LoadFinding` in `ListFindings`.

pkg/finding/finding.go
fmeum: Sounds good, we can revisit this if it becomes a bottleneck.

fmeum (approved):

PR #224, opened 2022-08-16
Sort findings by date

adombeck:


adombeck:
We currently overwrite an existing finding when the same fuzz test is run twice and it crashes with same input, because we choose the finding name based on the hash of the crashing input. That has the effect that the `CreatedAt` date of an old finding is set to `time.Now` when the fuzz test is re-run, and therefore old findings are listed on top once they are "found again". I'm not sure what the best behavior would be here. Any ideas, @CodeIntelligenceTesting/cifuzz-devs?

fmeum:
My first thought was that we could detect that we already have a finding for that crashing input and simply not write out a JSON.

But that has fundamental issues: Over time, the same crashing input may exercise different behavior (including different stack traces) due to the code changing. Maybe deriving finding names from just the crashing input isn't correct? We should think more about this and have a team discussion.

fmeum (approved):

PR #225, opened 2022-08-16
adding documentation for regression test to the cmake example

vargen:
included a note in README indicating cmake >= 3.20 is required to build the replayer binary.


jochil:
I think we should not add the cmake presets to the example project, as we will offer an easy way to install them with CLI-346 (this PR: https://github.com/CodeIntelligenceTesting/cifuzz/pull/226)
But referencing the command for installing them in the readme would be nice

examples/cmake/README.md
```diff
@@ -11,3 +11,21 @@ You can start the fuzzing with
 cifuzz run my_fuzz_test
 ```
 
+## Create Regression Test
+---
+**NOTE**
+
+Creating a replayer binary requires CMake >= 3.20.
```
jochil, line 18: I think this is not correct... using the cmake presets requires the higher version, the binary can be build with our normal requirement 3.16.

IMO the best way would be something like (I paraphrase):
If you use cmake >=3.19 you can use our presets like this (+ here is how you install them). Otherwise you can do it with this command: `cmake -S . -B build -DCIFUZZ_ENGINE -D...`
@fmeum what do you think?

And if I am not wrong presets were introduced in 3.19: https://cmake.org/cmake/help/v3.19/manual/cmake-presets.7.html

vargen: I see what you mean. I've associated the term 'replayer binary' with regression test (and debugging). Even though it won't work as a regression test, it can still be built with the lower version. I'm happy to adjust the language to be more precise.

You are correct, presets were added in 3.19. I listed 3.20 because that was the minimum version listed in the CMakePresets.json file.

fmeum: What @jochil said is entirely correct: The preset is just a nice way to specify a bunch of CMake command-line flags, so we shouldn't tie the ability to do regression tests to CMake 3.19+. 

Configuration presets were added in CMake 3.19, but build and test presets are only available since 3.20. Since I don't really see how the former would be useful without the latter, I would specify 3.20 as the minimum.

While we are at it, we should probably also explain how the test preset can be used to run all the regression tests. @vargen Do you want to add a section? I could also do it as a follow-up.

vargen:
Since the method in https://github.com/CodeIntelligenceTesting/cifuzz/pull/226 to install the presets appears to still be in flux, I removed mentions of the presets in the README for now along with the CMakePresets.json file. Can update  again after it is done.

jochil:
> Since the method in #226 to install the presets appears to still be in flux, I removed mentions of the presets in the README for now along with the CMakePresets.json file. Can update again after it is done.

sounds good for me... I created CLI-384 as a reminder

examples/cmake/README.md
jochil: see CLI-384 for follow  up

jochil (approved):

PR #226, opened 2022-08-16
add tools integration

florianGla:


share/tasks.json
```diff
@@ -0,0 +1,12 @@
...
+  "tasks": [
+    {
+      "label": "cifuzz coverage (current test)",
+      "type": "shell",
+      "command": "cifuzz",
+      "args": ["coverage", "-f", "lcov", "-o", "lcov.info", "${fileBasenameNoExtension}"],
+      "problemMatcher": []
+    }
+  ]
+}
```
adombeck, line 12: No newline at end of file

internal/cmd/integrate/integrate.go
```diff
@@ -0,0 +1,344 @@
...
+		content := string(bytes)
+
+		gitIgnore, err := os.OpenFile(gitIgnorePath, os.O_APPEND|os.O_WRONLY, 0644)
+		if err != nil {
+			return err
+		}
+		defer gitIgnore.Close()
+
+		for _, fileToIgnore := range filesToIgnore {
+			if !strings.Contains(content, fileToIgnore) {
```
adombeck, line 151: With this check, the file isn't added to `.gitignore` even if it's part of any existing line, which doesn't mean that the file is ignored (e.g. file `foo` isn't added to `.gitignore` if it already contains `subdir/foo`). I think we should instead check if any line of the existing `.gitignore` exactly matches `fileToIgnore` here.

internal/cmd/integrate/integrate.go
```diff
@@ -0,0 +1,344 @@
...
+	}
+
+	if !hasGitIgnore {
+		err = os.WriteFile(gitIgnorePath, []byte(strings.Join(filesToIgnore, "\n")), 0644)
+		if err != nil {
+			return err
+		}
+		log.Printf(`
+.gitignore has been created to prevent cifuzz generated files from
+being put under version control.`)
```
adombeck, line 136: This message feels a bit clunky. How about we print `Added files generated by cifuzz to .gitignore` in both the case that `.gitignore` already exists and doesn't exist?

internal/cmd/integrate/integrate.go
```diff
@@ -0,0 +1,344 @@
...
+		// the existing tasks.json
+		presetsSrc, err := readJSONFile(tasksSrcPath)
+		if err != nil {
+			return err
+		}
+		presetsDst, err := readJSONFile(tasksDestPath)
+		if err != nil {
+			return err
+		}
+		fmt.Println("read json files")
```
adombeck, line 199: Is this supposed to be here? Looks like debug statement

internal/cmd/integrate/integrate.go
```diff
@@ -0,0 +1,344 @@
...
+	scanner := bufio.NewScanner(jsonFile)
+	for scanner.Scan() {
+		line := scanner.Text()
+		if strings.Contains(line, "//") {
+			continue
+		}
+		bytes = append(bytes, []byte(line)...)
+	}
+	if scanner.Err() != nil {
+		return nil, scanner.Err()
```
adombeck, line 288:
```suggestion
		return nil, errors.WithStack(scanner.Err())
```

internal/cmd/integrate/integrate.go
```diff
@@ -0,0 +1,344 @@
...
+	}
+
+	jsonMap := orderedmap.New()
+	err = json.Unmarshal(bytes, &jsonMap)
+	if err != nil {
+		return nil, err
```
adombeck, line 294:
```suggestion
		return nil, errors.WithStack(err)
```

internal/cmd/integrate/integrate.go
```diff
@@ -0,0 +1,344 @@
...
+	}
+
+	return jsonMap, nil
+}
+
+func writeJSONFile(content *JsonMap, path string) error {
+	jsonOut, err := json.MarshalIndent(content, "", "  ")
+	if err != nil {
+		return err
```
adombeck, line 303:
```suggestion
		return errors.WithStack(err)
```

internal/cmd/integrate/integrate.go
```diff
@@ -0,0 +1,344 @@
...
+	}
+	err = os.WriteFile(path, jsonOut, 0644)
+	if err != nil {
+		return err
```
adombeck, line 307:
```suggestion
		return errors.WithStack(err)
```

internal/cmd/integrate/integrate_test.go
```diff
@@ -0,0 +1,15 @@
...
+
+	"github.com/stretchr/testify/assert"
+
+	"code-intelligence.com/cifuzz/pkg/cmdutils"
+)
+
+func TestIntegrateCmd(t *testing.T) {
+	_, err := cmdutils.ExecuteCommand(t, New(), os.Stdin)
+	assert.Error(t, err)
+}
```
adombeck, line 15: This could use an actual test, which checks that the command behaves as expected in the various scenarios.

internal/cmd/integrate/integrate.go
```diff
@@ -0,0 +1,346 @@
...
+
+	return nil
+}
+
+func (c *integrateCmd) addGitIgnoreIntegration() error {
+	filesToIgnore := []string{
+		".cifuzz-build/",
+		".cifuzz-corpus/",
+		".cifuzz-findings/",
+		"*_seed_corpus",
```
fmeum, line 119: We explicitly want the seed corpus to be checked in, so I think we can remove this line.

internal/cmd/integrate/integrate.go
```diff
@@ -0,0 +1,346 @@
...
+		"*.coverage.html",
```
fmeum, line 120: Could we make this pattern more precise, e.g. by prescribing a directory in which these files live? It otherwise feels a bit broad.

jochil:
What's about using https://docs.pterm.sh/printers/interactive/multiselect when the user did not an argument? So it would be possible to install the git and clion integration at once

florianGla:
> What's about using https://docs.pterm.sh/printers/interactive/multiselect when the user did not add an argument? So it would be possible to install the git and clion integration at once It would be similar to the create command

good idea

internal/cmd/integrate/integrate.go
```diff
@@ -0,0 +1,272 @@
...
+
+	https://code.visualstudio.com/docs/editor/tasks
+
+You can download the Coverage Gutters extension from:
+
+	https://marketplace.visualstudio.com/items?itemName=ryanluker.vscode-coverage-gutters`)
+	} else {
+		// Situation: The user does have a tasks.json file set up, so we
+		// assume them to know about the benefits. We add our task to
+		// the existing tasks.json
```
adombeck, line 205:
```suggestion
		// assume them to know about the benefits. We suggest to the user
		// that they add our task to the existing tasks.json.
```

internal/cmd/integrate/integrate.go
```diff
@@ -0,0 +1,272 @@
...
+tests and measure code coverage right from your IDE.
+This file should not be checked in to version control systems.
+To learn more about CMake presets, visit:
+
+    https://github.com/microsoft/vscode-cmake-tools/blob/main/docs/cmake-presets.md
+    https://www.jetbrains.com/help/clion/cmake-presets.html`)
+	} else {
+		// Situation: The user does have a CMake user preset set up, so we
+		// assume them to know about the benefits. We add our presets to
+		// the existing CMakeUserPresets.json
```
adombeck, line 258:
```suggestion
		// assume them to know about the benefits. We suggest to the user
		// that they add our preset to the existing CMakeUserPresets.json.
```

internal/cmd/integrate/integrate.go
```diff
@@ -0,0 +1,272 @@
...
+		presetsSrc, err := os.ReadFile(tasksSrcPath)
+		if err != nil {
+			return errors.WithStack(err)
+		}
+
+		log.Printf(`
+Add the following task to your tasks.json to provide easy access to
+cifuzz coverage runs from within VS Code. You can use the Coverage
+Gutters extension to visualize the generated coverage report.
+
```
adombeck, line 215: Remove whitespace
```suggestion

```

adombeck (approved):

fmeum (approved):

PR #227, opened 2022-08-16
Parse stack trace and attach it to finding

adombeck:


internal/cmd/run/report_handler/stacktrace/stacktrace.go
```diff
@@ -0,0 +1,195 @@
...
+
+var framePattern = regexp.MustCompile(
+	`#(?P<frame_number>\d+)\s+0x[a-fA-F0-9]+\s+in\s+(?P<function>(\(anonymous namespace\))?[^(\s]+).*\s(?P<source_file>\S+?):(?P<line>\d+):?(?P<column>\d*)`)
+
+var sourceLocationPattern = regexp.MustCompile(`^(?P<source_file>\S+?):(?P<line>\d+):(?P<column>\d+)`)
+
+// A StackFrame represents an element of the stack trace
+type StackFrame struct {
+	SourceFile  string
+	Line        int32
```
fmeum, line 23: I would find `uint32` more descriptive.

internal/cmd/run/report_handler/stacktrace/stacktrace.go
```diff
@@ -0,0 +1,195 @@
...
+	Column      int32
+	FrameNumber int32
```
fmeum, line 25: Do we expect this to always equal the index of the `StackFrame` in the slice? If we do, maybe we should remove this number and verify the equality when parsing the stack trace?

internal/cmd/run/report_handler/stacktrace/stacktrace.go
```diff
@@ -0,0 +1,195 @@
...
+	// which we can't rely on.
+	var err error
+	if filepath.IsAbs(path) {
+		path, err = filepath.Rel(p.projectDir, path)
+		if err != nil {
+			return "", errors.WithStack(err)
+		}
+	}
+	if strings.HasPrefix(path, "..") {
+		return "", errors.Errorf("file %q is not below project directory %q", path, p.projectDir)
```
fmeum, line 164: Wouldn't this would break if fuzz tests use third-party libraries that ship with debug symbols? We wouldn't expect their source files to be under the project dir.

internal/cmd/run/report_handler/stacktrace/stacktrace.go
adombeck: Done in a fixup commit (which I will squash before merging)

internal/cmd/run/report_handler/stacktrace/stacktrace.go
adombeck: No, I don't think we expect this to be equal to the index in the slice, because we only include symbolized stack frames with a filename and line number. For example, the complete stack trace of the finding found in our `examples/cmake` project is:

```
#0 0x5627a02c1319 in __asan_memcpy (/home/adrian/cifuzz/examples/cmake/.cifuzz-build/libfuzzer/address+undefined/my_fuzz_test+0xe0319) (BuildId: cb57908f4da32e9b7f0b50d243efd64af9799834)
    #1 0x5627a0300b4a in exploreMe(int, int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >) /home/adrian/cifuzz/examples/cmake/src/explore_me.cpp:24:11
    #2 0x5627a02ff4e7 in LLVMFuzzerTestOneInputNoReturn(unsigned char const*, unsigned long) /home/adrian/cifuzz/examples/cmake/my_fuzz_test.cpp:18:3
    #3 0x5627a02ff328 in LLVMFuzzerTestOneInput /home/adrian/cifuzz/examples/cmake/my_fuzz_test.cpp:8:1
    #4 0x5627a0225583 in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned long) (/home/adrian/cifuzz/examples/cmake/.cifuzz-build/libfuzzer/address+undefined/my_fuzz_test+0x44583) (BuildId: cb57908f4da32e9b7f0b50d243efd64af9799834)
    #5 0x5627a0224ca9 in fuzzer::Fuzzer::RunOne(unsigned char const*, unsigned long, bool, fuzzer::InputInfo*, bool, bool*) (/home/adrian/cifuzz/examples/cmake/.cifuzz-build/libfuzzer/address+undefined/my_fuzz_test+0x43ca9) (BuildId: cb57908f4da32e9b7f0b50d243efd64af9799834)
    #6 0x5627a0226888 in fuzzer::Fuzzer::ReadAndExecuteSeedCorpora(std::vector<fuzzer::SizedFile, std::allocator<fuzzer::SizedFile> >&) (/home/adrian/cifuzz/examples/cmake/.cifuzz-build/libfuzzer/address+undefined/my_fuzz_test+0x45888) (BuildId: cb57908f4da32e9b7f0b50d243efd64af9799834)
    #7 0x5627a0226d72 in fuzzer::Fuzzer::Loop(std::vector<fuzzer::SizedFile, std::allocator<fuzzer::SizedFile> >&) (/home/adrian/cifuzz/examples/cmake/.cifuzz-build/libfuzzer/address+undefined/my_fuzz_test+0x45d72) (BuildId: cb57908f4da32e9b7f0b50d243efd64af9799834)
    #8 0x5627a02151c2 in fuzzer::FuzzerDriver(int*, char***, int (*)(unsigned char const*, unsigned long)) (/home/adrian/cifuzz/examples/cmake/.cifuzz-build/libfuzzer/address+undefined/my_fuzz_test+0x341c2) (BuildId: cb57908f4da32e9b7f0b50d243efd64af9799834)
    #9 0x5627a023f1b2 in main (/home/adrian/cifuzz/examples/cmake/.cifuzz-build/libfuzzer/address+undefined/my_fuzz_test+0x5e1b2) (BuildId: cb57908f4da32e9b7f0b50d243efd64af9799834)
    #10 0x7f0ac0e29209 in __libc_start_call_main ../sysdeps/nptl/libc_start_call_main.h:58:16
    #11 0x7f0ac0e292bb in __libc_start_main ../csu/libc-start.c:389:3
    #12 0x5627a02099e0 in _start (/home/adrian/cifuzz/examples/cmake/.cifuzz-build/libfuzzer/address+undefined/my_fuzz_test+0x289e0) (BuildId: cb57908f4da32e9b7f0b50d243efd64af9799834)
```

... but the first stack frame we store is ` #1 0x5627a0300b4a in exploreMe`, because the one above doesn't have a source file and line number.

internal/cmd/run/report_handler/stacktrace/stacktrace.go
adombeck: Indeed. That raises the question: How do we want to handle stack frames for third party libraries? We will probably use this stack trace information for:
1. The finding summary
2. Finding deduplication
3. Breakpoints

In the core repo, we ignore stack frames which are not below the project directory, so there they are not considered when deduplicating findings and creating breakpoints.

adombeck: We discussed this in a call, with the following results:

> 1. The finding summary
We assume that the topmost line in the user code is more useful than the line in a third-party library, so we want to ignore third-party libraries in the summary. It's important the full stack trace is easily accessible though.

> 2. Finding deduplication
* If the same fuzz test is run with and without debug symbols for third-party libraries, we would create duplicate findings if we include stack frames from third-party libraries.
* We assume that the user wants to find bugs in the code inside the project directory. If the stack frames from those source files are the same, the root cause of multiple findings is probably the same, so it makes sense to deduplicate them.

> 3. Breakpoints
We're not sure yet if / how we want to automatically set breakpoints when debugging via cifuzz. We need to clarify that first.

pkg/finding/finding.go
```diff
@@ -52,6 +53,9 @@ type Finding struct {
 	Tag                uint64        `json:"tag,omitempty"`
 	ShortDescription   string        `json:"short_description,omitempty"`
 	InputFile          string
+
+	// The following fields don't exist in the protobuf representation
```
jochil, line 57: is this comment needed? It might be misleading as we do not have a protobuf definition in this project

adombeck: The core repo converts those structs to protobuf messages, I think it's helpful to note that those fields are lost in that conversion.

jochil: ok. I expected that... what do you think about prefixing these kind of comments with `// CI Core:` (or similar) to make it clear for external developers that this is nothing they have to care about?

adombeck: I don't think that prefix would really clear things up for external developers. How about this?
```
// Note: The following fields don't exist in the protobuf
// representation used in the Code Intelligence core repository.
```
This should at least make it clear that this note is about another Code Intelligence repository.

jochil: I think this will work just fine... I thought the prefix might be helpful (if used in other locations too) to quickly find all connections points, but nevermind :)

internal/cmd/run/report_handler/stacktrace/stacktrace.go
```diff
@@ -0,0 +1,199 @@
...
+		return nil, err
+	}
+	if trace != nil {
+		return trace, nil
+	}
+
+	// Some findings don't produce a stack trace but a single source
+	// location, like this:
+	//
+	//    SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior fuzz-targets/trigger_ubsan.cpp:5:5 in
```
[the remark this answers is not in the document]
adombeck: You are right, the regex doesn't match the summary line. I copied this from our libfuzzer runner test, but the regex actually matches the first line:
```
fuzz-targets/trigger_ubsan.cpp:5:5: runtime error: shift exponent 32 is too large for 32-bit type 'int'
    #0 0x555c482ac32b in TriggerUndefinedBehaviorSanitizer() fuzz-targets/trigger_ubsan.cpp:5:5
    #1 0x555c482ac3f3 in LLVMFuzzerTestOneInput fuzz-targets/trigger_ubsan.cpp:15:3
    #2 0x555c481d24e3 in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned long) (/tmp/libfuzzer-runner-integration-test-45293162/build-182372873/trigger_ubsan+0x434e3) (BuildId: acb3d23da3a70b80171aac27ebbbf66cae67e129)
    #3 0x555c481d1c09 in fuzzer::Fuzzer::RunOne(unsigned char const*, unsigned long, bool, fuzzer::InputInfo*, bool, bool*) (/tmp/libfuzzer-runner-integration-test-45293162/build-182372873/trigger_ubsan+0x42c09) (BuildId: acb3d23da3a70b80171aac27ebbbf66cae67e129)
    #4 0x555c481d3856 in fuzzer::Fuzzer::ReadAndExecuteSeedCorpora(std::vector<fuzzer::SizedFile, std::allocator<fuzzer::SizedFile> >&) (/tmp/libfuzzer-runner-integration-test-45293162/build-182372873/trigger_ubsan+0x44856) (BuildId: acb3d23da3a70b80171aac27ebbbf66cae67e129)
    #5 0x555c481d3cd2 in fuzzer::Fuzzer::Loop(std::vector<fuzzer::SizedFile, std::allocator<fuzzer::SizedFile> >&) (/tmp/libfuzzer-runner-integration-test-45293162/build-182372873/trigger_ubsan+0x44cd2) (BuildId: acb3d23da3a70b80171aac27ebbbf66cae67e129)
    #6 0x555c481c2122 in fuzzer::FuzzerDriver(int*, char***, int (*)(unsigned char const*, unsigned long)) (/tmp/libfuzzer-runner-integration-test-45293162/build-182372873/trigger_ubsan+0x33122) (BuildId: acb3d23da3a70b80171aac27ebbbf66cae67e129)
    #7 0x555c481ec112 in main (/tmp/libfuzzer-runner-integration-test-45293162/build-182372873/trigger_ubsan+0x5d112) (BuildId: acb3d23da3a70b80171aac27ebbbf66cae67e129)
    #8 0x7f76ca029209 in __libc_start_call_main ../sysdeps/nptl/libc_start_call_main.h:58:16
    #9 0x7f76ca0292bb in __libc_start_main ../csu/libc-start.c:389:3
    #10 0x555c481b6940 in _start (/tmp/libfuzzer-runner-integration-test-45293162/build-182372873/trigger_ubsan+0x27940) (BuildId: acb3d23da3a70b80171aac27ebbbf66cae67e129)

SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior fuzz-targets/trigger_ubsan.cpp:5:5 in
```

Since we're currently only storing the source file, line and column, we could just as well use the summary line. I think that we're only trying to match UBSan findings with this regex, so I looked at the UBSan code and AFAICT it does indeed always [print the summary](https://github.com/llvm/llvm-project/blob/73a9dfcee24df959b59a46d75dcbdc0bcfb50fe6/compiler-rt/lib/ubsan/ubsan_diag.cpp#L386-L388) whenever it [prints the line starting with the source file](https://github.com/llvm/llvm-project/blob/73a9dfcee24df959b59a46d75dcbdc0bcfb50fe6/compiler-rt/lib/ubsan/ubsan_diag.cpp#L343-L374). But the latter actually includes more detailed information about the error, so I think we might want to parse that and include that information in a field in the finding.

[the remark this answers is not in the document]
adombeck: > we might want to parse that and include that information in a field in the finding

We could match on `^(\S+):(\d+?):(\d+?): runtime error: (.*)$` because the `runtime error:` part is always printed when UBSan reports an error.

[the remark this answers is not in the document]
fmeum: > We could match on `^(\S+):(\d+:)?(\d+:)? runtime error: (.*)$` because the `runtime error:` part is always printed when UBSan reports an error.

Sounds good, let's use that regex instead.

internal/cmd/run/report_handler/stacktrace/stacktrace.go
```diff
@@ -0,0 +1,195 @@
...
+		frame, err := p.stackFrameFromLine(line)
+		if err != nil {
+			return nil, err
+		}
+		if frame == nil {
+			continue
+		}
+
+		if len(frames) > 0 && frame.FrameNumber < frames[len(frames)-1].FrameNumber {
+			// The frame number is lower than the one from the previous
```
fmeum, line 70: Wouldn't "lower or equal" make more sense here?

internal/cmd/run/report_handler/stacktrace/stacktrace.go
```diff
@@ -0,0 +1,195 @@
...
+func (p *parser) relSourceFilePath(path string) (string, error) {
+	// We have to make sure that the source file path of the break point
+	// is relative to the project directory to be able to debug on a
+	// different system than the one which built the binary.
+	//
+	// Note that we can't use the --relativenames option of
+	// llvm-symbolizer to produce relative paths in the stack trace
+	// because that only produces relative paths if the compiler
+	// command-line also contained relative paths to the source files,
+	// which we can't rely on.
```
fmeum: All the more reason to just not use that option. We should remove this comment when we no longer use the flag.

integration-tests/cmake/cmake_test.go
```diff
@@ -126,6 +127,34 @@ func TestIntegration_CMake_InitCreateRunCoverageBundle(t *testing.T) {
 	findings = getFindings(t, cifuzz, dir)
 	require.Len(t, findings, 1)
 	require.Contains(t, findings[0].Details, "heap-use-after-free")
+	// TODO: This check currently fails on macOS because of
+	// https://code-intelligence.atlassian.net/browse/CLI-371
```
fmeum, line 131: Can we inline at least some of the information on that non-public ticket here?

internal/cmd/run/report_handler/stacktrace/stacktrace.go
```diff
@@ -0,0 +1,228 @@
...
+
+	// To make the stack trace more useful when the finding is shared
+	// with others, and to avoid leaking the directory structure, we
+	// make the source file path relative to the project directory.
+	//
+	// Note that we can't rely on the the --relativenames option of
+	// llvm-symbolizer to produce relative paths in the stack trace
+	// because that only produces relative paths if the compiler
+	// command-line also contained relative paths to the source files.
+	log.Printf("XXX: 1 path: %v", path)
```
fmeum, line 158: You should drop the `XXX` commit.

internal/cmd/run/report_handler/stacktrace/stacktrace.go
adombeck: I thought the plan was to keep using the flag when the LLVM version supports it, because it does produce more usable stack traces in some cases.

internal/cmd/run/report_handler/stacktrace/stacktrace.go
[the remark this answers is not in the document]
adombeck: Done

internal/cmd/run/report_handler/stacktrace/stacktrace.go
adombeck: Done

integration-tests/cmake/cmake_test.go
adombeck: I rebased this on #234 and removed the commit which disables the check

internal/cmd/run/report_handler/stacktrace/stacktrace.go
adombeck: Done

integration-tests/cmake/cmake_test.go
adombeck: I had to add the commit again, forgot that the test still fails because the debug info for the shared library is missing / only available in the object files. I adjusted the comment to explain the issue instead of referring to the non-public ticket.

internal/cmd/run/report_handler/stacktrace/stacktrace.go
fmeum: True, let's keep it.

fmeum (approved):

PR #229, opened 2022-08-17
Update Configuration.md

pnoltof:
It seems like the default unit is ms so timeout: 300 would result in a test stopping after 300ms.
I adjusted the example to set the timeout to 300 seconds.


jochil:
I just went quickly over the code and it seems that we just use it as `https://pkg.go.dev/time#Duration`... so actually it should be even nano seconds (if I did not miss a conversion somewhere) :)

IMO the best way would be changing the code to work with seconds (I think we do not need a smaller unit and libfuzzer uses seconds too). Otherwise we also have to change comments (`cifuzz run --help`) and templates (`internal/config/cifuzz.yaml.tmpl`)

@adombeck what do you think?

adombeck:
Thanks @pnoltof and @jochil for reporting and investigating this. I created #231 to fix the issue. We do indeed expect a duration string, not an integer. That was incorrectly documented and it was an error that parsing the config file didn't fail if the unit was missing. Note that using the `--timeout` flag did already fail with a descriptive error message:

```
$ cifuzz run my_fuzz_test --timeout 5
invalid argument "5" for "--timeout" flag: time: missing unit in duration "5"
```

With #231, parsing a `cifuzz.yaml` with the content `timeout: 5` now fails with:
```
Failed to parse cifuzz.yaml: error decoding 'timeout': time: missing unit in duration "5"
```

PR #234, opened 2022-08-18
cmake: Link debug symbols with dsymutil on macOS

fmeum:


fmeum:
@adombeck This is untested. If it doesn't fix your issues after you have rebased your branch onto this PR, let me know and I will debug it.

adombeck:
The stack frames from the fuzz test executable contain source file and line number now, but the ones from the `libparser.dylib` do not: https://github.com/CodeIntelligenceTesting/cifuzz/runs/7897008904?check_suite_focus=true#step:4:244

Maybe we also have to call `dsymutil` for the dynamic libraries?

fmeum:
> The stack frames from the fuzz test executable contain source file and line number now, but the ones from the `libparser.dylib` do not: https://github.com/CodeIntelligenceTesting/cifuzz/runs/7897008904?check_suite_focus=true#step:4:244
> 
> Maybe we also have to call `dsymutil` for the dynamic libraries?

I pushed a commit that should handle this.

tools/cmake/cifuzz/cmake/cifuzz-functions.cmake
```diff
@@ -51,6 +51,18 @@ function(enable_fuzz_testing)
           -UNDEBUG
       )
     endif()
+
+    if(APPLE)
+      # On macOS, debug information is only contained in the object files by default. llvm-symbolizer, which we use to
+      # resolve addresses in stack traces to source location, doesn't read the object files. We thus need to invoked
```
adombeck, line 57:
```suggestion
      # resolve addresses in stack traces to source location, doesn't read the object files. We thus need to invoke
```

tools/cmake/cifuzz/cmake/cifuzz-functions.cmake
```diff
@@ -191,6 +203,16 @@ function(add_fuzz_test name)
     message(FATAL_ERROR "cifuzz: Unsupported value for CIFUZZ_ENGINE: ${CIFUZZ_ENGINE}")
   endif()
 
+  # On macOS, debug information is only contained in the object files by default. llvm-symbolizer, which we use to
+  # resolve addresses in stack traces to source location, doesn't read the object files. We thus need to invoked
```
adombeck, line 207:
```suggestion
  # resolve addresses in stack traces to source location, doesn't read the object files. We thus need to invoke
```

adombeck:
> @adombeck This runs into https://github.com/CodeIntelligenceTesting/cifuzz/runs/7898399551?check_suite_focus=true#step:4:113, which I have seen multiple times already. Do you have an idea how to fix it?

It looks like the process didn't terminate within 3 seconds after the SIGTERM was sent, so it was killed. I have no idea why it would take that long to terminate.

adombeck:
>> Maybe we also have to call dsymutil for the dynamic libraries?

> I pushed a commit that should handle this.

That doesn't seem to be working, the dynamic library is still not symbolized with filename and line number:
https://github.com/CodeIntelligenceTesting/cifuzz/runs/7898842990?check_suite_focus=true#step:4:244

fmeum:
@adombeck I force-pushed away everything except for the first commit, feel free to put everything else back if you need it.

adombeck (approved):

PR #241, opened 2022-08-24
cmake: Further clarify cmake_minimum_required requirement

fmeum:


tools/cmake/cifuzz/cmake/cifuzz-functions.cmake
```diff
@@ -2,7 +2,7 @@
 #       Explanations of these flags are provided in that file.
 function(enable_fuzz_testing)
   if(${CMAKE_MINIMUM_REQUIRED_VERSION} VERSION_LESS 3.16)
-    message(FATAL_ERROR "cifuzz: cmake_minimum_required(VERSION 3.16) or higher required, got ${CMAKE_MINIMUM_REQUIRED_VERSION}")
+    message(FATAL_ERROR "cifuzz: cmake_minimum_required(VERSION 3.16) or higher required, but the current project declares a minimum version of ${CMAKE_MINIMUM_REQUIRED_VERSION}")
```
adombeck, line 5: Is the `cmake_minimum_required` not always declared in a `CMakeLists.txt`?

```suggestion
    message(FATAL_ERROR "cifuzz: cmake_minimum_required(VERSION 3.16) or higher required, but the CMakeLists.txt declares a minimum version of ${CMAKE_MINIMUM_REQUIRED_VERSION}")
```

fmeum: Well, that's almost always true but technically not required. You could also call it in a function loaded from a package. Also, there are usually many `CMakeLists.txt` files, so "the" one doesn't exist.

adombeck (approved):

PR #242, opened 2022-08-24
add success message if installer succeeds

wojtekCodeInt:


HenrichN:
I would propose to keep the styling of the success message consistent with our other messages like 
image
image

pkg/install/installer.go
```diff
@@ -373,13 +373,14 @@ func PrintPathInstructions(binDir string) {
...
 		// TODO: On Windows, users generally don't expect having to fiddle with their PATH. We should update it for
 		//       them, but that requires asking for admin access.
-		_, _ = fmt.Fprintf(os.Stderr, `Please add the following directory to your PATH:
+		_, _ = fmt.Fprintf(os.Stderr, `Installation successful. Please add the following directory to your PATH:
     %s
-`, binDir)
+If you haven't already done so.`, binDir)
 	} else {
-		_, _ = fmt.Fprintf(os.Stderr, `Please add the following to ~/.profile or ~/.bash_profile:
+        binDirWithHome := strings.Replace(binDir, "~", "$HOME", 1)
```
fmeum, line 380: Can you explain why this replacement is necessary?

adombeck: @wojtekCodeInt ping

jochil:
It would be nice if we can get rid of these "fix go syntax" commits before merging. For the future it can be avoided using this git hooks: https://github.com/CodeIntelligenceTesting/cifuzz/tree/main/githooks

pkg/install/installer.go
jochil: what shell are you using?

wojtekCodeInt: GNU bash, version 5.0.17(1)-release (x86_64-pc-linux-gnu)

adombeck: >  For me, the commands itself work, but the "which" command does not show their location (just empty output).

I can't reproduce that

jochil: in theory `~` should be a substitution for $HOME (with some extras) and there might be shells not supporting it. But I can reproduce it neither and GNU bash definitely has it. 
So as long we do not know the reason for this behavior, I would prefer going with `~`

wojtekCodeInt: ok I removed this part.

jochil:
> It would be nice if we can get rid of these "fix go syntax" commits before merging. For the future it can be avoided using this git hooks: https://github.com/CodeIntelligenceTesting/cifuzz/tree/main/githooks

Same for the `remove the ~ to $HOME change as it seems it is not needed that much` commit
@wojtekCodeInt I can squash the commits and force push the changes, if this is ok for you

jochil (approved):

PR #243, opened 2022-08-25
replayer: Output doctest xml if launched by CLion

florianGla:


tools/replayer/src/replayer.c
```diff
@@ -364,6 +367,22 @@ static void run_file_or_dir(const char *path) {
   }
 }
 
+static void print_clion_doctest_xml() {
+  /*
+   * Output Doctest xml, which enables the CLion test framework integration
+   * to indicate the test success status.
+   */
+  fprintf(stdout,
```
fmeum, line 375: When printing to stdout you can simply use `printf`.

tools/replayer/src/replayer.c
```diff
@@ -364,6 +367,22 @@ static void run_file_or_dir(const char *path) {
...
+          "<?xml version=\"1.0\" encoding=\"UTF-8\"?>"
+          "\n<doctest>"
```
fmeum, line 377: Are the `\n`s significant to CLion? If not, then I would vote for leaving them out. The visual line breaks help, but the ones in the string literal will never be seen by a human.

tools/replayer/src/replayer.c
```diff
@@ -69,6 +69,9 @@
  * be executed. */
 static const char *SEED_CORPUS_SUFFIX = "_seed_corpus";
 
+// A nonzero value indicates that replayer was launched as a Doctest target by the CLion test framework integration.
```
fmeum, line 72: You have to use `/* ... */` for compatibility with ancient C standards.

florianGla: That explains the failing tests.

tools/replayer/src/replayer.c
florianGla: You are right, only the last \n is significant.

fmeum (approved):

fmeum (approved):

PR #244, opened 2022-08-26
run: Make fuzz test path available to build-command 

fmeum:


fmeum:
@pnoltof Could you give this a try?

fmeum:
I forgot about the line length limit we agree on - the long description for run seemed to use 80.

pnoltof:
> @pnoltof Could you give this a try?

Works for me!

internal/cmd/run/run.go
```diff
@@ -111,10 +108,13 @@ depends on the build system configured for the project:
    or after running 'cifuzz reload'. The '--build-command' flag is ignored.
 
  * For other build systems, a command which builds the fuzz test executable
-   must be provided via '--build-command' and <fuzz test> is the path to
-   the fuzz test executable created by the build command. Example:
+   must be provided via the '--build-command' argument or the 'build-command'
+   YAML field. In this case, <fuzz test> is the path to the fuzz test executable
```
adombeck, line 112:
```suggestion
   setting in cifuzz.yaml. In this case, <fuzz test> is the path to the fuzz test executable
```

internal/cmd/run/run.go
```diff
@@ -111,10 +108,13 @@ depends on the build system configured for the project:
...
+   created by the build command. The value specified for <fuzz test> is
+   available to the build command in the FUZZ_TEST environment variable.
+   Example:
 
-       cifuzz run --build-command="make clean && make my_fuzz_test" my_fuzz_test
+       cifuzz run --build-command="make clean && make $FUZZ_TEST" my_fuzz_test
```
adombeck, line 117: The `$` needs to be escaped here to avoid that the variable is expanded by the shell which executes `cifuzz run`.

```suggestion
       cifuzz run --build-command="make clean && make \$FUZZ_TEST" my_fuzz_test
```

adombeck: In this example, using `$FUZZ_TEST` instead of `my_fuzz_test` only makes it more complex. Maybe we should use this example instead?
```
echo "build-command: make clean && make \$FUZZ_TEST" >> cifuzz.yaml
cifuzz run my_fuzz_test
```

adombeck:
> The value specified for <fuzz test> is available to the build command in the FUZZ_TEST environment variable.

I like that!

internal/cmd/run/run.go
fmeum: Great point, updated the commit.

internal/cmd/run/run.go
```diff
@@ -111,10 +108,15 @@ depends on the build system configured for the project:
...
-   must be provided via '--build-command' and <fuzz test> is the path to
-   the fuzz test executable created by the build command. Example:
-
-       cifuzz run --build-command="make clean && make my_fuzz_test" my_fuzz_test
+   must be provided via the '--build-command' argument or the 'build-command'
+   setting in cifuzz.yaml. In this case, <fuzz test> is the path to the fuzz
+   test executable created by the build command. The value specified for
+   <fuzz test> is available to the build command in the FUZZ_TEST environment
+   variable.
+   Example:
```
adombeck, line 116:
```suggestion
   variable. Example:
```

adombeck (approved):

PR #250, opened 2022-08-29
Build cmake projects with multiple cores

florianGla:


internal/build/cmake/cmake.go
```diff
@@ -151,10 +152,17 @@ func (b *Builder) Build(fuzzTests []string) (map[string]*build.Result, error) {
...
+	flags := append([]string{
 		"--build", buildDir,
 		"--config", cmakeBuildConfiguration,
-		"--target"}, fuzzTests...)...)
+		"--target"}, fuzzTests...)
+
+	flags = append(flags, "--parallel")
+	if b.NumJobs != 0 {
+		flags = append(flags, fmt.Sprint(b.NumJobs))
+	}
```
adombeck, line 163: This does not implement the suggestion from @fmeum, which was to build with a single core by default

internal/cmd/bundle/bundle.go
```diff
@@ -99,6 +107,8 @@ func New(conf *config.Config) *cobra.Command {
 		},
 	}
 
+	cmd.Flags().Uint("build-jobs", 1, "Maximum number of concurrent processes to use when building.\nNo value or 0 sets no limit. Only available when the build system is CMake.")
+	cmd.Flags().Lookup("build-jobs").NoOptDefVal = "0"
```
adombeck, line 111: IIUC, the idea was that by default we just keep the default of the build system, which means building with a single core in the case of Make and CMake, bazel on the other hand builds in parallel by default.

florianGla: IIUC from the CMake documentation, using --parallel without any arguments uses the default number of processes,
even though on my system --parallel uses the maximum number of processes.

--parallel [<jobs>], -j [<jobs>]
The maximum number of concurrent processes to use when building. If <jobs> is omitted the native build tool's default number is used (see https://cmake.org/cmake/help/latest/manual/cmake.1.html)

florianGla: I am not sure how to resolve this issue. Maybe we need three states (without parallel flag, with parallel flag without arguments, with parallel flag with arguments). What do you think?

fmeum: But doesn't that mean that there is a difference between using `--parallel` without argument and not using it at all? If that's true, we could decide to mirror that in cifuzz: No flag means build system default, flag without argument means "let the build system figure out a reasonable default for the number of processes" and a numeric argument means "let the build system use this many processes".

fmeum: Just saw your comment. Yes, I think we might need (and want) that behavior.

internal/cmd/bundle/bundle.go
```diff
@@ -246,9 +256,13 @@ func (c *bundleCmd) buildAllVariants() ([]map[string]*build.Result, error) {
...
 	for _, variant := range configureVariants {
 		builder, err := cmake.NewBuilder(&cmake.BuilderOptions{
-			ProjectDir:      c.config.ProjectDir,
-			Engine:          variant.Engine,
-			Sanitizers:      variant.Sanitizers,
+			ProjectDir: c.config.ProjectDir,
+			Engine:     variant.Engine,
+			Sanitizers: variant.Sanitizers,
+			Parallel: cmake.ParallelOptions{
+				Enabled: c.Flags().Lookup("build-jobs").Changed,
```
adombeck, line 263: `c.Flags().Lookup("build-jobs").Changed` is only true if the flag was set, not if the `build-jobs` setting was set in the `cifuzz.yaml`. I think we want this instead:

```suggestion
				Enabled: viper.IsSet("build-jobs"),
```

(same for the other usages of `c.Flags().Lookup("build-jobs").Changed`)

internal/cmd/bundle/bundle.go
```diff
@@ -83,6 +86,11 @@ func New(conf *config.Config) *cobra.Command {
 		ValidArgsFunction: completion.ValidFuzzTests,
 		Args:              cobra.ArbitraryArgs,
 		PreRunE: func(cmd *cobra.Command, args []string) error {
+			// Bind viper keys to flags. We can't do this in the New
+			// function, because that would re-bind viper keys which
+			// were bound to the flags of other commands before.
+			cmdutils.ViperMustBindPFlag("build-jobs", cmd.Flags().Lookup("build-jobs"))
+
```
adombeck, line 93: There is a call to `config.ParseProjectConfig(opts)` missing here.

adombeck (approved):

PR #251, opened 2022-08-29
runner: Fix race condition

adombeck:
It was possible that the fuzzer command was not started yet when a
signal is handled, which can result in the command starting afterwards
and not being terminated.


fmeum:
Is it possible that the command never starts (e.g. the file system operations required get blocked by antivirus software) and we deadlock in the termination routine?

adombeck:
> Is it possible that the command never starts (e.g. the file system operations required get blocked by antivirus software) and we deadlock in the termination routine?

Yes, but I don't see a good way to handle that case

adombeck:
I amended the commit to unblock the termination routine in case that starting the runner fails. The case that starting the runner is blocked in some way is still not handled though.

fmeum:
Looks good enough for now, thanks! We could look into using a timeout if that should actually become a problem.

fmeum (approved):

PR #253, opened 2022-08-29
style: activate nonamedreturns linter

jochil:
... and fix/check all warnings


util/sliceutil/sliceutil.go
```diff
@@ -14,7 +14,7 @@ import (
 // return (false, false) if impossible.
 // return (true, false) if element was not found.
 // return (true, true) if element was found.
-func Contains(slice interface{}, element interface{}) (found bool) {
+func Contains(slice interface{}, element interface{}) (found bool) { //nolint:nonamedreturns
```
florianGla, line 17: I am not sure if it is necessary to name the return parameter for that function.

florianGla (approved):

pkg/parser/libfuzzer/libfuzzer_parser.go
```diff
@@ -497,8 +497,8 @@ func parseAsSlowInput(log string) *finding.Finding {
 	return nil
 }
 
-func parseAsSeedCorpusMessage(line string) (numSeeds uint, err error) {
-	numSeeds, err = parseAsNonEmptyCorpusMessage(line)
+func parseAsSeedCorpusMessage(line string) (uint, error) {
```
adombeck, line 500: Here, it's not clear from the function signature what the returned `uint` is without the variable name. Same for `parseAsNonEmptyCorpusMessage`.

jochil: Would you prefer changing the function name or adding named returns?

adombeck: I can't think of a better name so I'd add the named returns back

util/sliceutil/sliceutil.go
jochil: I was not sure about this... especially because of this `defer` function and the missing return
```
defer func() {
		if e := recover(); e != nil {
			found = false
		}
}()
```
@adombeck do you know what the exact behavior is in this case?

adombeck: This is some "clever" implementation from https://github.com/stretchr/testify. IMO it's not worth the time trying to understand if things will break here if we're not using named returns: Lets just keep the named return.

jochil: Ok, especially without tests I would also prefer to keep it like this.

PR #254, opened 2022-08-29
report_handler: Fix usage of atomic.Value

adombeck:
Fixes

    panic: sync/atomic: store of inconsistently typed value into Value


fmeum (approved):
Well, there goes compile-time type safety. I hope they reimplement this API with generics at some point.

PR #255, opened 2022-08-29
build: Add -I for cifuzz.h in coverage builds

fmeum:
Achieved by moving the FUZZ_TEST_CFLAGS logic into the common build env.


fmeum:
@pnoltof

pnoltof:
That works!
Thanks for the quick fix!

fmeum:
Switching to draft as `CommonBuildEnv` is also used in CMake.

fmeum:
Fixed by duplicating the include path logic. That's probably the better approach anyway as it protects us from overwriting the variable without noticing.

jochil (approved):

PR #249, opened 2022-08-28
Check ErrSilent during coverage test

MarkusZoppelt:
This generates and opens a coverage report even if `runFuzzTest(buildResult)` returns a silent error.

**Example (example/cmake fails on Mac arm64):**
```
Building my_fuzz_test
-- Configuring done
-- Generating done
-- Build files have been written to: /Users/mz/code/cifuzz/examples/cmake/.cifuzz-build/replayer/coverage
Consolidate compiler generated dependencies of target exploreMe
[ 40%] Built target exploreMe
Consolidate compiler generated dependencies of target my_fuzz_test
[100%] Built target my_fuzz_test
Running my_fuzz_test on corpus
❌ my_fuzz_test: exit status 1

✅ Coverage Report:
              File | Functions Hit/Found |  Lines Hit/Found | Branches Hit/Found
  my_fuzz_test.cpp |      2 / 2 (100.0%) |   8 / 8 (100.0%) |     0 / 0 (100.0%)
src/explore_me.cpp |      1 / 1 (100.0%) | 19 / 21  (90.5%) |     4 / 8  (50.0%)
                   |                     |                  |                   
                   | Functions Hit/Found |  Lines Hit/Found | Branches Hit/Found
             Total |               3 / 3 |          27 / 29 |              4 / 8

```

IMHO this is a more helpful behaviour compared to only logging an `exit status 1` message.

PR #256, opened 2022-08-29
coverage: Print stderr on error

adombeck:


MarkusZoppelt:
This is clearly an improvement as it facilitates debugging target exits as described in, e.g., #249 
👍🏼

fmeum (approved):

PR #259, opened 2022-08-30
installer: Install shell completion script

adombeck:


pkg/install/installer.go
```diff
@@ -472,6 +472,21 @@ func ExtractBundle(targetDir string, bundle *embed.FS) error {
 		return err
 	}
 
+	// Install the autocompletion script for the current shell (if the
+	// shell is supported)
+	cifuzzPath := filepath.Join(targetDir, "bin", "cifuzz")
+	switch filepath.Base(os.Getenv("SHELL")) {
```
jochil, line 478: Why not supporting powershell, too?

adombeck: Because it took long enough to figure out how to correctly set up the completion for the other shells and I didn't have the nerves to do the same for powershell on Windows, which I have zero experience with. IMO we can do that in a follow-up when we see the need for it (e.g. when we know that we have some Windows user who want to have that feature).

jochil: According to https://github.com/spf13/cobra/blob/main/shell_completions.md#creating-your-own-completion-command and https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_profiles?view=powershell-7.2 I think an easy way to do it is just writing a cifuzz.ps file and tell the user to source it

adombeck: If this requires user action anyway, doesn't it make sense to just suggest them to read `cifuzz completion powershell` and follow the instructions?

jochil: Indeed... thats true

jochil (approved):

pkg/install/installer.go
adombeck: I implemented it but I can't test it on Windows because of https://code-intelligence.atlassian.net/browse/CLI-429

adombeck: To unblock this PR, I'll create a separate ticket and do that as a follow-up once we fixed the issue with the installer on Windows

PR #260, opened 2022-08-30
feat: optimize findings output/summary during fuzzing run

jochil:


internal/cmd/run/report_handler/report_handler.go
```diff
@@ -185,51 +190,44 @@ func (h *ReportHandler) handleFinding(f *finding.Finding, print bool) error {
...
 
-    %s
+	return nil
+}
 
-`, f.Name, fileutil.PrettifyPath(f.GetSeedPath()))
+func (h *ReportHandler) PrintFindingInstruction() {
+	log.Printf(`
+For further details on a finding, use:
+    cifuzz finding <name of finding>`)
```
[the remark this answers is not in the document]
adombeck:
```suggestion
	log.Note(`
Use 'cifuzz finding <finding name>' for details on a finding
```

internal/cmd/run/report_handler/report_handler.go
```diff
@@ -186,52 +190,44 @@ func (h *ReportHandler) handleFinding(f *finding.Finding, print bool) error {
...
 	switch {
+	case !isDuplicate:
+		log.Printf(" - NEW %s", f.ShortDescription())
+
 	case isDuplicate && f.GetSeedPath() == "":
 		// The finding is a duplicate and the input is also already known
-		log.Notef(`
-Note: This seems to be a duplicate of finding [%s].
-`, f.Name)
+		log.Printf(" - REGRESSION [%s]", f.Name)
```
adombeck, line 199: Why do we call this "regression"? When the user just executes the same fuzz test twice, the same findings are reported as regressions the second time. Do we expect that users fix all findings before the fuzz test is run again? Else I would prefer to call this "duplicate".

internal/cmd/run/report_handler/report_handler.go
```diff
@@ -239,13 +237,7 @@ including remote runs with artifacts created via 'cifuzz bundle' and
 regression tests. For more information on regression tests, see:
 
     https://github.com/CodeIntelligenceTesting/cifuzz#regression-testing
```
adombeck, line 239: I think we should use the plural in this message if more than one crashing input has been copied:
```
Note: The crashing inputs have been copied to the seed corpus at:

    my_fuzz_test_seed_corpus/affectionate_banach-1
    my_fuzz_test_seed_corpus/admiring_kalam-2

They will now be used a seed inputs for all runs of the fuzz test,
including remote runs with artifacts created via 'cifuzz bundle' and
regression tests. For more information on regression tests, see:

    https://github.com/CodeIntelligenceTesting/cifuzz#regression-testing
```

internal/cmd/run/report_handler/report_handler.go
jochil: If the crashing input is still in the seed corpus it has to be fixed to continue with fuzzing, or? So I think it is legit to call it a regression until it was fixed

adombeck: > If the crashing input is still in the seed corpus it has to be fixed to continue with fuzzing, or?

That depends, UBSan findings for example allow the fuzzer to continue

adombeck: > So I think it is legit to call it a regression until it was fixed

To me a regression is something that was working before and is now broken. That's not the case here. We could use a different definition of "regression" here, but IMO without any further explanations, the term is less intuitive here than "duplicate".

internal/cmd/run/report_handler/report_handler.go
jochil: Or just using plural for all cases. `All crashing inputs have been copied to the seed corpus at...` sounds fitting even with one input.

But is there actual a case where you can have two crashing inputs? Even after adding ubsan to the example project and having two findings, there is only one crashing input... which makes sense in this scenario, or?

internal/cmd/run/report_handler/report_handler.go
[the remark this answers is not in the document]
jochil: ok, done

internal/cmd/run/report_handler/report_handler.go
adombeck: > But is there actual a case where you can have two crashing inputs? Even after adding ubsan to the example project and having two findings, there is only one crashing input... which makes sense in this scenario, or?

I took a look at the libFuzzer code and it looks like the crashing input is indeed only written for findings which cause the fuzzer to stop. So you're right, we don't have to consider the case of multiple crashing inputs for now (until we try to continue fuzzing after crashes as well, e.g. by using libFuzzer's fork mode).

internal/cmd/run/report_handler/report_handler.go
jochil: > That depends, UBSan findings for example allow the fuzzer to continue

But UBSan findings are not added as a crashing input to the seed corpus... as they are not crashing, or? (basically same question as below)

> To me a regression is something that was working before and is now broken. That's not the case here (well maybe it is, but we don't know that). We could use a different definition of "regression" here, but IMO without any further explanations, the term is less intuitive here than "duplicate".

I think in our scenario adding the crashing input to the seed corpus is like writing a regression test. So the next step usually is to fix the bug before your tests getting green again.
But I agree... this is less intuitive as "duplicate". Another option could be "OLD", as counterpart to "NEW".

internal/cmd/run/report_handler/report_handler.go
```diff
@@ -185,51 +190,43 @@ func (h *ReportHandler) handleFinding(f *finding.Finding, print bool) error {
...
 
-`, f.Name, fileutil.PrettifyPath(f.GetSeedPath()))
+func (h *ReportHandler) PrintFindingInstruction() {
+	log.Note(`
+Use 'cifuzz finding <finding name>' for details on a finding.`)
 
-	case !isDuplicate:
-		log.Printf(`
-Further details:
+	log.Print("\nFindings:")
```
adombeck, line 213: When using `--verbose`, the list of findings is intertwined with output from the fuzzer. We could avoid that by not printing this `Findings:` header and somehow making it clear in each printed line that the line describes a finding.

internal/cmd/run/report_handler/report_handler.go
jochil: We already have the code for adding multiple entries to the message, so we can make it plural... this would be more generic

internal/cmd/run/report_handler/report_handler.go
jochil: What's about: With `--verbose` we print the list of findings only on the end of the run and for the findings we add some kind of visual separation again?

adombeck: I thought about that as well but I would prefer printing the findings without the header because:
- Using the same format allows users to parse the output in the same way both in verbose and non-verbose mode
- Even in verbose mode it's useful to print these summaries while the fuzzing run is still running, because hopefully they allow the user to get a quick understanding of a finding without having to read all the logs. (IMO we should also highlight these lines visually for that.)

jochil: I agree... makes sense. What do you think about this (incl. removed `Findings:\n`) : 

image

image

adombeck: I like it!

jochil: ok, I added that to this PR

internal/cmd/run/report_handler/report_handler.go
jochil: I changed it

internal/cmd/run/report_handler/report_handler.go
jochil: As a result of todays discussion, the different cases are obsolete now

internal/cmd/run/report_handler/report_handler.go
```diff
@@ -185,67 +188,42 @@ func (h *ReportHandler) handleFinding(f *finding.Finding, print bool) error {
...
-		fallthrough
+	if len(crashingInputs) == 0 {
+		return
+	}
 
-	case f.GetSeedPath() != "":
-		log.Notef(`
-Note: The crashing input has been copied to the seed corpus at:
+	log.Notef(`
+Note: All crashing inputs have been copied to the seed corpus at:
```
adombeck, line 217: Since we only ever have one crashing input, as you correctly stated, lets change this back to the singular form

internal/cmd/run/report_handler/report_handler.go
```diff
@@ -185,67 +188,42 @@ func (h *ReportHandler) handleFinding(f *finding.Finding, print bool) error {
...
-	lenLeftBar := (65 - len(title)) / 2
-	lenRightBar := (65 - len(title) + 1) / 2
-	log.Print("\n", strings.Repeat("=", lenLeftBar), title, strings.Repeat("=", lenRightBar))
-	// TODO: Print short summary instead
-	log.Print(f.Details)
-	if len(f.StackTrace) > 0 {
-		st := f.StackTrace[0]
-		log.Printf("in %s (%s:%d:%d)", st.Function, st.SourceFile, st.Line, st.Column)
-	}
+	log.Printf("💥 %s", f.ShortDescription())
```
adombeck, line 191: The output looks good, but when using the `--verbose` flag with a fuzz target which prints a lot of output (like our cmake example), these finding lines are hard to find and now that we're not prefixing it with "NEW" anymore, they don't contain anything to search for except for the emoji (which just doesn't feel right to use on the command-line). I currently can't think of anything we could add to make the finding lines searchable, do you have an idea?

jochil: I thought about that too... "NEW" is not really correct in all cases, so I would avoid that. "Finding" is long and it looks weird repeating it multiple times, without `--verbose`. Maybe something like `| 💥 foo`, `-> 💥 foo` or `|> 💥 foo`

I tend to say that the normal output is for humans, even with `--verbose` (like you said recently)... so if you want to parse our output use `--json` or live with grepping for a unicode emoji :)

adombeck:
Because I think it's something that's not immediately clear and which we might ask ourselves again later why we did it, I took the liberty to push a commit which explains in a comment why we now base the finding names on the crashing input as well. Please take a look, @jochil.

internal/cmd/run/report_handler/report_handler.go
adombeck: > I tend to say that the normal output is for humans, even with --verbose (like you said recently)... so if you want to parse our output use --json or live with grepping for a unicode emoji :)

Yes, for parsing the `--json` flag is definitely a good solution, but as a human looking at the verbose output I still want to be able to find the interesting lines. But since we don't have a good solution now, I suggest we postpone this and come back to it if we notice that it's hurting UX in practice.

pkg/finding/finding.go
```diff
@@ -258,6 +214,29 @@ func (f *Finding) moveInputFile(projectDir, seedCorpusDir string) error {
...
+	default:
+		errorType = f.Details
+	}
+
+	output := fmt.Sprintf("[%s] %s", f.Name, errorType)
+
+	// add location (file, function, line)
+	if len(f.StackTrace) > 0 {
+		f := f.StackTrace[0]
+		output = fmt.Sprintf("%s in %s (%s:%d:%d)", output, f.Function, f.SourceFile, f.Line, f.Column)
```
adombeck, line 235: There are cases in which in which ASan / libFuzzer don't include the column in the stack trace, so lets do:
```suggestion
		if f.Column != 0 {
			output = fmt.Sprintf("%s in %s (%s:%d:%d)", output, f.Function, f.SourceFile, f.Line, f.Column)
		} else {
			output = fmt.Sprintf("%s in %s (%s:%d)", output, f.Function, f.SourceFile, f.Line)
		}
```

jochil: ok. I structured the code a little bit different... but should work the same

PR #261, opened 2022-08-31
lower clang/llvm min version to v10

florianGla:


pkg/runner/runnerutils.go
```diff
@@ -177,9 +179,15 @@ func FuzzerEnvironment() ([]string, error) {
...
 	// See https://llvm.org/docs/CommandGuide/llvm-symbolizer.html#cmdoption-llvm-symbolizer-relativenames
-	env, err = envutil.Setenv(env, "LLVM_SYMBOLIZER_OPTS", "--relativenames")
-	if err != nil {
-		return nil, err
+	dep := dependencies.Default[dependencies.LLVM_SYMBOLIZER]
+	version, err := dep.FindVersion(runfiles.Finder)
+
+	// This flag is only available for llvm-symbolizer v11 or greater
+	if err == nil && version.Compare(semver.MustParse("11.0.0")) > -1 {
+		env, err = envutil.Setenv(env, "LLVM_SYMBOLIZER_OPTS", "--relativenames")
```
jochil, line 187: IIRC @adombeck mentioned that the flag is not needed anymore as we replace all paths in the stack trace later 
Is that correct? If so, we can skip this whole block

jochil:
If we want our pipelines to consistently use llvm v10, we should also try to lower the version in the following files for at least the linux vms/containers:

- .github/actions/vm_setup/action.yml (this is used in most of the pipelines (release, push, pr, installer))
- .github/workflows/pipeline_e2e_installer.yml
- .github/workflows/pipeline_e2e_source.yml

If the downgrading to v10 is to complicating I would be fine if we only do it in the `/pipeline_e2e_source.yml` but we should at least have one pipeline running with v10

pkg/runner/runnerutils.go
adombeck: We never relied on this option, but it does produce more convenient logs in some cases, so I'd prefer to keep it unless that causes us too much effort.

adombeck:
@florianGla and I briefly discussed this PR, with Florian mentioning that it's hard to even install LLVM 10 on Ubuntu and that it's a somewhat broken version which was quickly superseded by LLVM 11. Given that our current minimum version is LLVM 11, we found that it's not really worth the effort to lower the version to LLVM 10, so we're in favor of closing this PR. Did I summarize that correctly, @florianGla?

jochil:
> @florianGla and I briefly discussed this PR, with Florian mentioning that it's hard to even install LLVM 10 on Ubuntu and that it's a somewhat broken version which was quickly superseded by LLVM 11. Given that our current minimum version is LLVM 11, we found that it's not really worth the effort to lower the version to LLVM 10, so we're in favor of closing this PR. Did I summarize that correctly, @florianGla?

after giving it a second thought, I agree. Having a lot of quirky workarounds in the CI is not worth it for a version that was out there for 6 months 

But while working on the PR it came up that we are not having any pipeline/test that is actually verifying that our min. requirements are working. So maybe we should create a follow up task for that... I think hacking together one container with all the minimal requirements should be feasible. What do you think @florianGla @adombeck ?

florianGla:
> But while working on the PR it came up that we are not having any pipeline/test that is actually verifying that our min. requirements are working. So maybe we should create a follow up task for that... I think hacking together one container with all the minimal requirements should be feasible. What do you think @florianGla @adombeck ?

I think that is a good idea.

PR #263, opened 2022-09-02
run regression tests from the cli

florianGla:


docs/How-To-Write-A-Fuzz-Test.md
```diff
@@ -1,27 +1,5 @@
 # How to write a fuzz test
 
-
-## How to build/compile your fuzz tests
```
jochil, line 4: In the main README is a link to this section. We should update that too

jochil (approved):

PR #266, opened 2022-09-02
integrate: Remove multiselect view

florianGla:


internal/cmd/integrate/integrate.go
```diff
@@ -52,7 +51,7 @@ CMakeUserPresets.json file and by adding tasks to your tasks.json:
 Missing files are generated automatically.
 `,
 		ValidArgs: supportedTools(),
-		Args:      cobra.MatchAll(cobra.MaximumNArgs(3), cobra.OnlyValidArgs),
+		Args:      cobra.MatchAll(cobra.RangeArgs(1, 3), cobra.OnlyValidArgs),
```
jochil, line 54: can we use something like `len(tools)` for the max value? I think this is a thing that you forgot to change if adding support for a new tool

jochil (approved):

PR #269, opened 2022-09-05
runner: Use ASAN_OPTIONS and UBSAN_OPTIONS from environment

adombeck:
To allow users to configure the sanitizers themselves.


fmeum (commented):
Could you add a test to make it easier for is to preserve the property that users can set sanitizer options?

florianGla (approved):

adombeck:
> Could you add a test to make it easier for is to preserve the property that users can set sanitizer options?

We already have a test for that [here](https://github.com/CodeIntelligenceTesting/cifuzz/blob/CLI-444-asan-options/pkg/runner/libfuzzer/integration-tests/with_env_vars_test.go#L39-L39). That test uses the `libfuzzer.RunnerOptions.EnvVars` field to set `ASAN_OPTIONS` instead of setting it in the process environment - which is the better thing to do in a test IMO, because setting the global environment can have unexpected side effects.

fmeum:
> We already have a test for that [here](https://github.com/CodeIntelligenceTesting/cifuzz/blob/CLI-444-asan-options/pkg/runner/libfuzzer/integration-tests/with_env_vars_test.go#L39-L39). That test uses the `libfuzzer.RunnerOptions.EnvVars` field to set `ASAN_OPTIONS` instead of setting it in the process environment - which is the better thing to do in a test IMO, because setting the global environment can have unexpected side effects.

I agree that this is the better technique for smaller scoped tests in general, but we should still consider adding coverage for this to an integration test. Jazzer recently broke OSS-Fuzz pretty badly precisely because of a bug in code that wasn't exercised in tests because of concerns like this.

How about setting the [`coverage` and `coverage_dir` options](https://github.com/google/sanitizers/wiki/SanitizerCommonFlags) in one of the integration tests and verifying that they are honored by checking that the coverage dir has been created?

adombeck:
> I agree that this is the better technique for smaller scoped tests in general, but we should still consider adding coverage for this to an integration test. Jazzer recently broke OSS-Fuzz pretty badly precisely because of a bug in code that wasn't exercised in tests due to concerns like this.
> 
> How about setting the [coverage and coverage_dir options](https://github.com/google/sanitizers/wiki/SanitizerCommonFlags) in one of the integration tests and verifying that they are honored by checking that the coverage dir has been created?

I added a check to the cmake integration test. Using the `coverage` didn't have any effect, I assume because we don't build with coverage instrumentation in the `run` command. I used the `print_stats=1` option instead, which we already use in the libfuzzer test (https://github.com/CodeIntelligenceTesting/cifuzz/blob/CLI-444-asan-options/pkg/runner/libfuzzer/integration-tests/with_env_vars_test.go#L39-L39).

fmeum (approved):

PR #274, opened 2022-09-05
docs: update glossary (metrics)

jochil:


docs/Glossary.md
```diff
@@ -46,9 +46,26 @@ different meanings depending on the context, previous knowledge or
 might even be used differently by the various fuzzing engines. 
 To avoid confusion we provide the following definitions:
 
-### Features
+### exec/s
+This number points out how often your fuzz target gets called with a
+generated input per second (usually as an average).
```
fmeum, line 51: This clarifies that the number is more about the past rather than the present performance:
```suggestion
The average number of times per second the fuzz target has been called
with a generated input since fuzzing started. 
```

docs/Glossary.md
```diff
@@ -46,9 +46,26 @@ different meanings depending on the context, previous knowledge or
 might even be used differently by the various fuzzing engines. 
 To avoid confusion we provide the following definitions:
 
-### Features
+### exec/s
+This number points out how often your fuzz target gets called with a
+generated input per second (usually as an average). 
 
-### Branches
+### Paths
```
fmeum, line 53: Given that we refer to specific output of `cifuzz run`, we should use consistent lowercase:
```suggestion
### paths
```

docs/Glossary.md
```diff
@@ -46,9 +46,26 @@ different meanings depending on the context, previous knowledge or
...
+While the fuzzer feeds generated inputs to an application, more and more
```
fmeum, line 54: Since this particular term is very important to understand, I would provide some more details:

```
While the fuzzer feeds generated inputs to a fuzz test, it progressively explores the code under test. The "paths" metric captures the different kinds of progress the fuzzer can make, such as

* reaching new lines of code;
* executing a loop body a different number of times;
* satisfying equality for larger parts of a failing comparison.

While this number increasing can be taken as a sign that the fuzzer is still making progress, it is not meaningful to compare across different fuzz tests.

For the libFuzzer engine, the "paths" metric coincides with the engine's "ft" (feature) count.
```

docs/Glossary.md
```diff
@@ -46,9 +46,26 @@ different meanings depending on the context, previous knowledge or
...
+code gets executed. This amount is described as "paths". 
+Sometimes also referred to as features.
+
+During a fuzzing run, **cifuzz** will inform you also about the 
+time passed since the last new path was discovered. This might give 
+you a hint if there are any problems for the fuzzer to reach and 
+execute more code.
```
fmeum, line 61: Could we move that into a separate section with the same title as reported by `cifuzz run` (probably "last new path")? 

```
The time that has passed since the last increase of the "paths" metric.
If this number keeps increasing, it is likely that the fuzzer isn't making progress anymore. In this case, use `cifuzz coverage` to get an idea of where the fuzzer got stuck.
```

docs/Glossary.md
```diff
@@ -46,9 +46,26 @@ different meanings depending on the context, previous knowledge or
...
 
 ### Coverage
```
fmeum, line 63: Maybe this should become a separate section since it isn't reported by `cifuzz run`? In general, I don't think that we should have to explain the concept of "code coverage" as long as it really is just that: coverage as usually measured by llvm-cov/jacoco and unrelated to any fuzzing concepts.

docs/Glossary.md
jochil: way better, thanks

docs/Glossary.md
jochil: good point... but I changed `If this number keeps increasing,` to `If this number stops increasing,`

docs/Glossary.md
jochil: I moved it out of the metrics section. In general I agree, but maybe there is a user not knowing about coverage and we already have written it... if the glossary is getting to big, we can still remove it

docs/Glossary.md
fmeum: Isn't that wrong? This section is about "time since last path", which keeps increasing if the fuzzer gets stuck.

jochil: :facepalm: of course... my mind was still in "paths". Fixed

PR #279, opened 2022-09-07
bundle: Mention that fuzz test names are supported as arguments

florianGla:


internal/cmd/bundle/bundle.go
```diff
@@ -81,9 +81,11 @@ type configureVariant struct {
 func New(conf *config.Config) *cobra.Command {
 	opts := &bundleOpts{}
 	cmd := &cobra.Command{
-		Use:               "bundle",
-		Short:             "Bundles the fuzz test into an archive",
-		Long:              "Bundles all runtime artifacts required by the given fuzz test into a self-contained archive that can be executed by a remote fuzzing worker",
+		Use:   "bundle [flags] <fuzz test>...",
```
adombeck, line 84: The `<fuzz test>` argument is optional, which is denoted by brackets:
```suggestion
		Use:   "bundle [flags] [<fuzz test>]...",
```

adombeck (approved):

PR #282, opened 2022-09-07
Add options to bundle command

adombeck:


internal/cmd/bundle/bundle.go
```diff
@@ -433,26 +480,49 @@ depsLoop:
...
-	archiveSeedsDir := ""
 	if exists {
+		seedCorpusDirs = append(seedCorpusDirs, buildResult.SeedCorpus)
+	}
+	var archiveSeedsDir string
+	if len(seedCorpusDirs) > 0 {
 		archiveSeedsDir = filepath.Join(fuzzTestPrefix(fuzzTest, buildResult), "seeds")
-		err = artifact.AddDirToManifest(archiveManifest, archiveSeedsDir, seedCorpus)
+		for _, dir := range seedCorpusDirs {
+			err = artifact.AddDirToManifest(archiveManifest, archiveSeedsDir, dir)
```
fmeum, line 505: I'm a bit worried that this might result in seeds being override that don't follow the `crash-<hash>` naming convention. Could we stage each directory in a separate subdirectory?

fmeum (approved):

internal/cmd/bundle/bundle.go
adombeck: done

internal/cmd/bundle/bundle.go
```diff
@@ -496,13 +497,27 @@ depsLoop:
...
+		var targetDirs []string
+		for _, sourceDir := range seedCorpusDirs {
+			// Put the seeds into subdirectories of the "seeds" directory
+			// to avoid seeds with the same name to override each other.
+
+			// Choose a name for the target directory which wasn't used
+			// before
+			basename := filepath.Join(archiveSeedsDir, filepath.Base(sourceDir))
+			targetDir := basename
+			for i := 1; sliceutil.Contains(targetDirs, targetDir); i++ {
```
fmeum, line 514: This use of `sliceutil.Contains` is a bit tricky, could we explain it in the comment?

adombeck: What exactly should we explain in the comment? The `sliceutil.Contains` is used as the termination condition of the for loop. It checks whether the current value of `targetDir` is in the list of existing target dirs.

fmeum: I would vote for making this a `for sliceutil.Contains(targetDirs, targetDir)` and manually increment the counter. I find index-based `for` loops where the termination condition doesn't clearly reference the index somewhat hard to understand.

adombeck: like this?

fmeum: Yes, thanks!

PR #285, opened 2022-09-09
Fix fuzzing run from clion

florianGla:
I am not sure if `envutil.RemoveFromEnv()` is the best function name.
The problem is that some environment variables have multiple key/value pairs as values (e.g. ASAN_OPTIONS=log_path=/tmp/log). This function should remove one value or key/value pair. What do you think?


adombeck:
@florianGla I pushed a commit which should achieve the same in a way that's more consistent with how we enforce other sanitizer options. Feel free to squash into your commit.

adombeck:
And if we add a function to remove sanitizer options (which I don't currently see the need for, because we can just override the option with `log_path=stderr` instead), that function should live in `runnerutils.go` and probably be named `UnsetSanitizerOption`.

adombeck (approved):

PR #288, opened 2022-09-09
installer: Refactor and rename "bundler" to "builder"

adombeck:
It was confusing that we had two bundlers in cifuzz, one which creates the fuzz test artifacts bundle and one which builds cifuzz and its dependencies.

The term builder is more common and it's not as confusing to have a cifuzz builder and a fuzz test builder in the same project.


cmd/installer/.gitignore
```diff
@@ -0,0 +1 @@
+build
```
fmeum, line 1: Could this be `/build`? It's also missing a newline.

adombeck: done

florianGla (approved):

fmeum (approved):

PR #290, opened 2022-09-12
Adjust launcher.c to prevent old-style-cast warning for cpp fuzz targets

pnoltof:


tools/cmake/cifuzz/src/launcher.c
```diff
@@ -38,7 +38,12 @@ void __msan_scoped_disable_interceptor_checks() {
     return;
   }
   /* Not running within cifuzz, replace the process with cifuzz running this fuzz test. */
-  POSIX_EXECLP("cifuzz", /*argv[0]=*/ "cifuzz", "run", CIFUZZ_TEST_NAME, (char *)0);
+  #ifdef __cplusplus
+  // Prevent old-style-casting warning for cpp build
+  POSIX_EXECLP("cifuzz", /*argv[0]=*/ "cifuzz", "run", CIFUZZ_TEST_NAME, static_cast<char*>(0));
+  #else
+  POSIX_EXECLP("cifuzz", /*argv[0]=*/ "cifuzz", "run", CIFUZZ_TEST_NAME, (char*)0);
+  #endif
```
fmeum, line 46: I think that this should also work, could you give it a try and adjust if so?
```suggestion
  POSIX_EXECLP("cifuzz", /*argv[0]=*/ "cifuzz", "run", CIFUZZ_TEST_NAME, NULL);
```

fmeum (approved):

PR #292, opened 2022-09-12
refactor(namesgenerator): copy the names list from moby/moby

jochil:
... to ensure that we have deterministic names, even if the list gets updated at the moby repository.
Also removing the moby dependency


jochil:
we decided to go with a list on our own to avoid controversial names in the future

PR #295, opened 2022-09-13
cmd/finding: Add short descriptions for argument completions

adombeck:


pkg/finding/finding.go
```diff
@@ -227,7 +227,7 @@ func (f *Finding) ShortDescription() string {
 		errorType = f.Details
 	}
 
-	output := fmt.Sprintf("[%s] %s", f.Name, errorType)
+	output := errorType
```
jochil, line 230: It seems a little bit bloated to remove the name here and add it manually at 3 other locations just to use `ShortDescription` once without name

Whats about a flag or another Method like `ErrorDescription`

jochil: Or just change the `ShortDescription` Format to work everywhere... like `fmt.Sprintf("%s\t%s", f.Name, errorType)`

adombeck: I pushed another commit which also improves the formatting of the output of `cifuzz findings` by using a different format for the short description (it also uses a tab between the error type and the location). I think this makes it harder to find one format which fits all usages. Do you still think we should try to do that?

jochil: IMO even another function for the finding completion would be better than always prepending the name manually to the `ShortDescription()`. What's about something like this:

```
func (f *finding) ErrorDescription() string {
 // return "heap buffer overflow in exploreMe (src/exploreMe.c:10:2)
}

func (f *finding) ShortDescription() string {
  return  fmt.Sprintf("[%s] %s, f.Name, f.ErrorDescription())
}
```
In this case you don't have to change the existing `ShortDescription` usages while avoiding duplication

adombeck: Done, but I used other names for the functions (`ShortDescription` and `ShortDescriptionWithName`)

PR #296, opened 2022-09-13
ci(bundle): fallback to github env vars if `git` is not working

jochil:
... to get commit sha and branch for artifact metadata


pkg/vcs/git.go
```diff
@@ -7,32 +7,64 @@ import (
 	"github.com/pkg/errors"
 
 	"code-intelligence.com/cifuzz/pkg/log"
+	"code-intelligence.com/cifuzz/util/envutil"
 )
 
-// GitCommit returns the full SHA of the current commit if the working directory is contained in a Git repository.
-func GitCommit() (string, error) {
+// GitCommit returns the full SHA of the current commit if the working directory
+// is contained in a Git repository. Otherwise it will check typical CI env vars.
```
fmeum, line 14: Could you explain in this commit why we could end up with a working directory that isn't contained in a Git repo but still have Git-specific env vars set?

Also, "typical" seems to mean "GitHub-provided" here. We could either make an effort to support other commons platforms (GitLab, CircleCI, BuildKite, ...) or document the ones that we support.

pkg/vcs/git.go
```diff
@@ -7,32 +7,64 @@ import (
...
+func GitBranch(env []string) (string, error) {
 	cmd := exec.Command("git", "rev-parse", "--abbrev-ref", "HEAD")
 	commit, err := cmd.Output()
 	if err != nil {
+		log.Debugf("unable to get branch via git command: %+v", err)
+
+		if _, ok := envutil.LookupEnv(env, "CI"); ok {
+			// It is necessary to check both GITHUB_* vars as they depend on how
+			// the github action was triggered
```
fmeum, line 46: How are they related precisely? Do we know that exactly one of them will be set in any GitHub Actions pipeline? The tests indicate that the relation is more subtle, we should document it here.

pkg/vcs/git.go
jochil: I updated the comment and the commit message (describing the specific scenario). It is not clear that the same problem will happen in other platforms. Trying to reproduce it on all these platforms would be a lot of effort I would like to avoid.... same for adding the platform specific vars without testing them

pkg/vcs/git.go
jochil: `GITHUB_HEAD_REF` is set to the source branch if the pipeline is triggered by a pull request
`GITHUB_REF_NAME` is set to the branch (tag) that triggered the pipeline

As far as I got it from https://docs.github.com/en/actions/learn-github-actions/environment-variables it is not guaranteed  that either of them is set for sure. For example when triggered via tag or as the result of another pipeline.

@HenrichN was very specific on the order of the variables... maybe he has more insights?

HenrichN: @fmeum @jochil This was how @simonresch  indicated we are doing it in our current Github Action.

fmeum: @simonresch Could you provide a short comment explaining why we do it this way? It's good to know that this approach is well tested though.

simonresch: We use `GITHUB_HEAD_REF` as the branch name for pull request actions, which contains the name of the branch that is being merged. For pull request actions the `GITHUB_REF_NAME` is `<PR_number>/merge` which might not be intuitive (though you could argue that it's more correct). Since `GITHUB_HEAD_REF` is not set for non PR actions, we use `GITHUB_REF_NAME` as a fallback.

Do we need to include this logic in `cifuzz`? Maybe we could rely on the github action to set a specific environment variable which is read if `git rev-parse` fails. In the github action template we could then add a line like this to handle the fallback logic: https://github.com/CodeIntelligenceTesting/c-cpp-demo/blob/two-vulnerabilities/.github/workflows/main.yml#L31
This could also generalize better to other CI/CD platforms.

jochil: We also need the commit sha (at least we are putting it in the artifact config file)... so it would be 2 specific ENVs for now.
And we should not forget that this is already a fallback for the `git` command. 

In general I like the idea with the specific vars, but it would somehow make a fallback to the default case. And it is just one more thing the user have to handle.
So my vote would be keeping it in cifuzz

@HenrichN @fmeum how should we continue?

HenrichN: I like the idea from @simonresch. This would make it also more compatible with other environments (think Gitlab or Jenkins). This way, we can handle the logic outside of cifuzz and it just checks if the two variables are there and if not, it uses the current default behaviour.

jochil: So we are making the `git` command the fallback? I think I am fine with this, too. 
So everyone ok with `CIFUZZ_BRANCH` and `CIFUZZ_COMMIT_SHA`? @CodeIntelligenceTesting/cifuzz-devs @HenrichN ?

jochil: The idiomatic way for `cifuzz` would be adding flags/params to the `cifuzz bundle` command... as soon as we support to overwrite config params with environment variables this would be pretty similar to the desired behavior. 
An advantage would also be that we the right have position to document it (`cifuzz bundle --help`)

Would this be ok @CodeIntelligenceTesting/cifuzz-devs @HenrichN

fmeum: +1 for implementing the env variable overrides now (should require little more than a viper call) and then implementing the fallback using `bundle` flags that can equivalently be represented as env variables.

jochil:
I created two follow up tasks: CLI-506 and CLI-507 two handle the outcome of the discussion... this PR can be ignored/closed now

PR #297, opened 2022-09-14
cmake: Add error message to remove build folder for unsupported compiler

florianGla:


tools/cmake/cifuzz/cmake/cifuzz-functions.cmake
```diff
@@ -173,7 +173,7 @@ function(add_fuzz_test name)
     elseif(CMAKE_CXX_COMPILER_ID STREQUAL "Clang" OR ((NOT "CXX" IN_LIST _enabled_languages) AND (CMAKE_C_COMPILER_ID STREQUAL "Clang")))
       target_link_options("${name}" PRIVATE -fsanitize=fuzzer)
     else()
-      message(FATAL_ERROR "cifuzz: ${CMAKE_CXX_COMPILER_ID} compiler is not supported with the libfuzzer engine")
+      message(FATAL_ERROR "cifuzz: ${CMAKE_CXX_COMPILER_ID} compiler is not supported with the libfuzzer engine. Please remove .cifuzz-build and try again.")
```
fmeum, line 176: We could be more specific here by printing the particular build directory with `${CMAKE_BINARY_DIR}`. We should also tell the user about what else they need to change to make the retry succeed, e.g. that they should specify the full path to (on macOS: non-Apple) clang/clang++ in CC/CXX or ensure that it comes first in their `PATH`.

fmeum (approved):

PR #299, opened 2022-09-14
Add remote-run command

adombeck:


internal/cmd/remote-run/remote_run.go
```diff
@@ -0,0 +1,299 @@
...
+	url := fmt.Sprintf("%s/v1/projects", c.opts.Server)
+	req, err := http.NewRequest("GET", url, nil)
+	req.Header.Add("Authorization", "Bearer "+token)
+
+	client := &http.Client{}
+	resp, err := client.Do(req)
+	if err != nil {
+		return "", errors.WithStack(err)
+	}
+	defer resp.Body.Close()
```
fmeum, line 178: AFAIK this will keep the HTTP connection open until the end of the scope, which includes a blocking call to `dialog.Select`. It probably won't matter, but I would still prefer to close this eagerly on the happy path. This has the advantage that we can also check for an error in that case.

internal/cmd/remote-run/remote_run.go
```diff
@@ -0,0 +1,299 @@
...
+	err = json.Unmarshal(objmap["projects"], &projects)
+	if err != nil {
+		return "", errors.WithStack(err)
+	}
+
+	// Let the user select a project
+	items := map[string]string{}
+	for _, p := range projects {
+		if p.OwnerOrganizationName != FeaturedProjectsOrganization {
+			// TODO: Maybe align the names
```
fmeum, line 209: Let's do that `fmt.Sprintf`'s padding should work well enough.

internal/cmd/remote-run/remote_run.go
```diff
@@ -0,0 +1,299 @@
...
+	}
+	projectName, err := dialog.Select("Select the project you want to start a fuzzing run for", items)
+	if err != nil {
+		return "", errors.WithStack(err)
+	}
+
+	return projectName, nil
+}
+
+func (c *runRemoteCmd) uploadArtifacts(path string, token string) (*artifact, error) {
```
fmeum, line 221: This may a good time to revisit our choice of compression algorithm: Given that we intend to upload the bundles and have the run start as fast as possible, should we continue to use GZip or use something more efficient such as BZip2?

internal/cmd/remote-run/remote_run.go
```diff
@@ -0,0 +1,299 @@
...
+	}
+	defer f.Close()
+
+	var buf bytes.Buffer
+	w := multipart.NewWriter(&buf)
+	fw, err := w.CreateFormFile("fuzzing-artifacts", path)
+	if err != nil {
+		return nil, errors.WithStack(err)
+	}
+	_, err = io.Copy(fw, f)
```
fmeum, line 234: Since it's not really more effort, can we use a pipe instead of a buffer as suggested in e.g. https://medium.com/@owlwalks/sending-big-file-with-minimal-memory-in-golang-8f3fc280d2c? That way we don't have to read the entire file into memory.

internal/cmd/remote-run/remote_run.go
```diff
@@ -0,0 +1,299 @@
...
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != 200 {
+		err := errors.Errorf("Starting remote fuzzing run failed with %v", resp.Status)
+		log.Error(err)
+		return cmdutils.WrapSilentError(err)
+	}
+
+	// TODO: Would be nice to be able to link to a page which immediately
```
fmeum, line 292: Given that we can't do that yet, could we print the name of the run here so that the user knows which one to look for in the dashboard?

We should also ask for this functionality to be added.

internal/cmd/root/root.go
```diff
@@ -90,8 +91,9 @@ func New() (*cobra.Command, error) {
 	rootCmd.AddCommand(initCmd.New())
 	rootCmd.AddCommand(createCmd.New(cmdConfig))
 	rootCmd.AddCommand(runCmd.New())
+	rootCmd.AddCommand(remoteRunCmd.New(nil))
 	rootCmd.AddCommand(reloadCmd.New(cmdConfig))
-	rootCmd.AddCommand(bundleCmd.New(cmdConfig))
+	rootCmd.AddCommand(bundleCmd.New(nil))
```
fmeum, line 96: Could you explain why we pass `nil` here rather than `cmdConfig` or not having the function take an argument in the first place?

internal/cmd/remote-run/remote_run.go
```diff
@@ -0,0 +1,299 @@
+package remote_run
```
fmeum, line 1: I know it's going to take some effort, but I think we need an automated test for this command that starts a mock server and verifies that the remote-run command interacts correctly.

internal/cmd/remote-run/remote_run.go
```diff
@@ -0,0 +1,299 @@
...
+		return nil, errors.WithStack(err)
+	}
+
+	url := fmt.Sprintf("%s/v2/%s/artifacts/import", c.opts.Server, c.opts.ProjectName)
+	req, err := http.NewRequest("POST", url, &buf)
+	req.Header.Set("Content-Type", w.FormDataContentType())
+	req.Header.Add("Authorization", "Bearer "+token)
+
+	client := &http.Client{}
+	resp, err := client.Do(req)
```
fmeum, line 249: Is there a way to get a progress bar for this upload, which might take quite some time?

internal/cmd/remote-run/remote_run.go
adombeck: After a quick glance at https://tukaani.org/lzma/benchmarks.html, my impression is that when considering the full length of compressing + uploading + decompressing, gzip is probably a better choice than bzip2. Looking at the first sample, a 203 MB OpenOffice.org installation tarball, at compression level 6 (the default in Go's `gzip` package), the gzip compressed size is ~8% bigger than the bzip2 compressed size at compression level 6 - but it took almost 6 times as long to compress.
Looking at the absolute numbers: With gzip the tarball was compressed to 75 MB within 24.5s, with bzip2 to 69 MB in 2m 18s. That means that with this sample, even with a slow uplink of 10 Mbit/s, bzip2 only saves ~5 seconds during uploading but takes 113s longer for compressing.

internal/cmd/remote-run/remote_run.go
adombeck: Great suggestion! Done.

internal/cmd/remote-run/remote_run.go
adombeck: done

internal/cmd/remote-run/remote_run.go
adombeck: Good point. I did some refactoring which solves the issue.

internal/cmd/remote-run/remote_run.go
```diff
@@ -0,0 +1,299 @@
...
+	token := os.Getenv("CIFUZZ_API_TOKEN")
+	if token == "" {
+		if !term.IsTerminal(int(os.Stdin.Fd())) {
+			b, err := io.ReadAll(os.Stdin)
+			if err != nil {
+				return errors.WithStack(err)
+			}
+			token = string(b)
+		} else {
+			fmt.Printf(`Enter an API access token and press Enter. You can generate a token for
```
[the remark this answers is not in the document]
adombeck: I started implementing that. Figuring out how to store such a secrets file on Windows was quite hard, I ended up copying some code from https://github.com/WireGuard/wireguard-windows to create a file with ACLs which, IIUC, only allows the current user and administrators to access the file.

internal/cmd/remote-run/remote_run.go
adombeck: Yes. Done by reusing some code from the core repo.

internal/cmd/remote-run/remote_run.go
fmeum: Thanks for the detailed analysis, sounds like we should stick to GZip.

internal/cmd/remote-run/remote_run.go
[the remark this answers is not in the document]
fmeum: It's definitely going to be useful, but atm we don't support Windows runners, so `remote-run` wouldn't need to support this just yet.

internal/cmd/remote-run/remote_run.go
```diff
@@ -0,0 +1,325 @@
...
+	cmd.Flags().Lookup("build-jobs").NoOptDefVal = "0"
+	// TODO(afl): Also link to https://github.com/AFLplusplus/AFLplusplus/blob/stable/dictionaries/README.md
+	cmd.Flags().String("dict", "", "A `file` containing input language keywords or other interesting byte sequences.\nSee https://llvm.org/docs/LibFuzzer.html#dictionaries.")
+	// TODO(afl): Also link to https://www.mankier.com/8/afl-fuzz
+	cmd.Flags().StringArray("engine-arg", nil, "Command-line `argument` to pass to the fuzzing engine.\nSee https://llvm.org/docs/LibFuzzer.html#options.")
+	cmd.Flags().StringArray("fuzz-test-arg", nil, "Command-line `argument` to pass to the fuzz test.")
+	// TODO(afl): Also link to https://aflplus.plus/docs/fuzzing_in_depth/#a-collecting-inputs
+	cmd.Flags().StringArrayP("seed-corpus", "s", nil, "A `directory` containing sample inputs for the code under test.\nSee https://llvm.org/docs/LibFuzzer.html#corpus.")
+	cmd.Flags().Duration("timeout", 0, "Maximum time to run the fuzz test, e.g. \"30m\", \"1h\". The default is to run indefinitely.")
+	cmd.Flags().StringVarP(&opts.OutputPath, "output", "o", "", "Output path of the artifact (.tar.gz)")
```
jochil, line 117: These flags are the same ones used in the `bundle` command, or?
If so, can we avoid to duplicate that code and make it easier to handle, when updating/renaming the flags or their descriptions

internal/cmd/remote-run/remote_run.go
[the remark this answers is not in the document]
adombeck: That's a good point, I didn't think about that. On systems other than Linux, we should probably just tell the user that they need to build on Linux to use the `remote-run` command, right?

internal/cmd/remote-run/remote_run.go
adombeck: > Given that we can't do that yet, could we print the name of the run here so that the user knows which one to look for in the dashboard?

The dashboard doesn't include a list of fuzzing runs. The user has to click on the fuzz test in the sidebar first to see the "Past runs". In that list the runs also don't have unique names, the names displayed there seem to be  `$BRANCH-$COMMIT`, which of course multiple fuzzing runs can have. I think the only way to then identify a fuzzing run is by looking at the date it was started at. I think describing those steps here would be quite awkward, because it makes it too obvious that the UX is bad. IMO we should just push for the functionality to link to a specific fuzzing run to be added.

internal/cmd/root/root.go
adombeck: I changed that

internal/cmd/remote-run/remote_run.go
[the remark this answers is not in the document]
fmeum: Yes, that would be the most helpful comment we can show. Maybe add a note such as "If you are interested in using this feature on ${OS}, please file an issue at https://github.com/...".

internal/cmd/remote-run/remote_run.go
fmeum: Definitely. I tried out the UI myself and also ran into this issue. This should be fixed in the web app.

adombeck: I created a ticket for that: https://code-intelligence.atlassian.net/browse/APP-948

internal/cmd/remote-run/remote_run.go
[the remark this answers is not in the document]
adombeck: done

internal/cmd/remote-run/remote_run.go
adombeck: done

internal/cmd/remote-run/remote_run.go
adombeck: done

internal/cmd/remote-run/remote_run.go
```diff
@@ -0,0 +1,417 @@
...
+		defer f.Close()
+
+		fmt.Println("Uploading...")
+		progressR := progress.NewReader(f, fileInfo.Size(), "Upload complete")
+		_, err = io.Copy(part, progressR)
+		return errors.WithStack(err)
+	})
+
+	// Send a POST request with what we read from the pipe. The request
+	// gets cancelled with the routines context is cancelled, which
```
fmeum, line 274:
```suggestion
	// gets cancelled when the routines context is cancelled, which
```

fmeum (approved):

PR #301, opened 2022-09-15
fix(finding): short description for memory leak

mlsandnerCI:
I added a fix for the CLI-478 bug and an integration test that produces a memory leak. The fix is not pretty and very specific but the whole function needs a rewrite at some point (TODO already exists) and if we encounter more specific error messages that don't get displayed correctly we can just expand the switch/case for now.


jochil:
> Is it okay if I add a check if the OS is windows to skip the test

Absolutely. I would have suggested the same

jochil (approved):

PR #303, opened 2022-09-16
ci: add pipeline to test compatibility between cifuzz versions

jochil:
... and also replace the soon to be deprecated ubuntu-18.04


.github/workflows/pipeline_e2e_installer.yml
```diff
@@ -17,7 +17,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [macos-11, ubuntu-22.04, ubuntu-18.04, windows-latest]
+        os: [macos-11, ubuntu-22.04, ubuntu-20.04, windows-2022]
```
fmeum, line 20: Could you split off the windows runner change? It's contained in the Ubuntu commit and not explained in the commit message.

fmeum (approved):

.github/workflows/pipeline_e2e_installer.yml
jochil: This is not the windows runner change we talked about, as `windows-latest == windows-2022`... I just wanted to use fixed versions everywhere. 
But this was obviously confusing, so I updated both commit messages

PR #304, opened 2022-09-17
feat: show desktop notifications when a finding was found

jochil:
On my system it looks like this:
image


internal/cmd/run/report_handler/report_handler.go
```diff
@@ -210,6 +211,14 @@ func (h *ReportHandler) handleFinding(f *finding.Finding, print bool) error {
 
 	log.Printf("💥 %s", f.ShortDescription())
 
+	// send a notification if running in a desktop system
+	if value, found := os.LookupEnv("DISPLAY"); found && value != "" {
```
fmeum, line 215: Not entirely sure whether or not we do this, but we should not emit a notification for duplicates.

I also don't know whether `DISPLAY` is commonly set on Windows.

fmeum, line 215: Not entirely sure whether or not we do this, but we should not emit a notification for duplicates.

I also don't know whether `DISPLAY` is commonly set on Windows.

jochil: What do you mean with duplicates? I choose this location as it would will show a notification for every finding output in the terminal.

IMO for Windows we can just assume that we have a display if `CI` is false. I will give it a second thought and adjust the if statement

fmeum (approved):

jochil:
image

jochil:
Ref: CLI-502

PR #308, opened 2022-09-19
Generate crash-resistant coverage reports

fmeum:


fmeum:
Stacked on #311, please review that one first.

internal/build/other/other.go
```diff
@@ -262,6 +262,16 @@ func (b *Builder) setCoverageEnv() error {
 		return err
 	}
 
+	ldflags := []string{
+		// ----- Flags used to link in coverage runtime -----
+		"-fprofile-instr-generate",
+		"-fcoverage-mapping",
+	}
+	b.env, err = envutil.Setenv(b.env, "LDFLAGS", strings.Join(ldflags, " "))
```
jochil, line 253: This overwrites existing `LDFLAGS`, or? Is this ok here?

.github/actions/vm_setup/action.yml
```diff
@@ -19,12 +19,12 @@ runs:
           path: |
             C:/Program Files/LLVM
             ./llvm
-          key: ${{ inputs.os }}-llvm-13.0
+          key: ${{ inputs.os }}-llvm-14.0
```
jochil, line 22: These pipelines are running with llvm 11:
* .github/workflows/pipeline_e2e_installer.yml
* .github/workflows/pipeline_min_requirements.yml

Did you check if this is a problem, too?

internal/cmd/coverage/coverage.go
```diff
@@ -595,3 +606,48 @@ func (c *coverageCmd) checkDependencies() (bool, error) {
...
+	var cifuzzCpu macho.Cpu
+	if runtime.GOARCH == "amd64" {
+		cifuzzCpu = macho.CpuAmd64
+	} else {
+		cifuzzCpu = macho.CpuArm64
+	}
+	if file, err := macho.OpenFat(object); err == nil {
+		defer file.Close()
+		var fallbackCpu macho.Cpu
+		for _, arch := range file.Arches {
```
jochil, line 659: Is there a scenario with arches other than `arm64` and/or `amd64`?

fmeum: Theoretically yes, but in practice these should be the only two architectures that are relevant.

.github/actions/vm_setup/action.yml
fmeum: https://github.com/CodeIntelligenceTesting/cifuzz/actions/runs/3097895038/jobs/5015117510 looks like it should also be failing on `main`.

fmeum: https://github.com/CodeIntelligenceTesting/cifuzz/actions/runs/3097892668/jobs/5015111936 passes.

jochil: Looks like `.github/workflows/pipeline_e2e_installer.yml` is broken for a long time :see_no_evil: 
It is suppose to run after each release, but it doesn't...I will take a look and created a task (CLI-510) for it

pkg/minijail/minijail.go
```diff
@@ -187,6 +187,8 @@ var defaultBindings = []*Binding{
 	// instead doesn't seem to be supported by nginx).
 	{Source: "/etc/passwd"},
 	{Source: "/etc/group"},
+	// Required for system(), which is used by libFuzzer's -merge mode.
+	{Source: "/bin/sh"},
```
adombeck, line 191: I think this will break if `/bin/sh` is a symlink (which is usually the case at least on Debian-based systems)

internal/cmd/coverage/coverage.go
```diff
@@ -547,12 +592,11 @@ func (c *coverageCmd) rawProfilePattern() string {
 	// file, all of which we have to merge in the end. By using "%m",
 	// the profile is written to a unique file for each executable and
 	// shared library.
-	//
-	// TODO: According to the documentation [1], "%c" should be useful
-	//       here, but for an unclear reason that results in no .profraw
-	//       files being generated.
```
adombeck, line 553: On my system, this change still causes the command to fail with
```
did not generate .profraw files at /tmp/minijail-out/coverage-3103363690/%c%m.profraw
```

internal/cmd/coverage/coverage.go
fmeum: Added explicit error handling.

internal/cmd/coverage/coverage.go
```diff
@@ -595,3 +606,48 @@ func (c *coverageCmd) checkDependencies() (bool, error) {
...
+				fallbackCpu = arch.Cpu
+			}
+		}
+		return cpuToArchFlag(fallbackCpu), nil
+	}
+	if file, err := macho.Open(object); err == nil {
+		defer file.Close()
+		return cpuToArchFlag(file.Cpu), nil
+	}
+	return "", errors.Errorf("failed to parse Mach-O file: %s", object)
```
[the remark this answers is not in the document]
fmeum: Good point, I am now printing both in case of an error.

pkg/minijail/minijail.go
fmeum: It's a symlink also on my system (`/bin/sh` --> `/bin/dash` --> `/usr/bin/dash`), but this works. Would I have to add bindings for all intermediate targets to make this fully reliable? How did we solve this for the unit test runner?

internal/cmd/coverage/coverage.go
fmeum: Do you have an idea why? Does it work with sandboxing disabled?

adombeck: I have no idea why. Disabling sandboxing doesn't fix it.

fmeum: We debugged this locally and found out that it is caused by a lingering outdated cifuzz CMake package installed as root.

pkg/minijail/minijail.go
adombeck: I was sure that I had issues before because minijail didn't follow symlinks in the bind-mount paths, but apparently it does. Note that I found [this commit](https://github.com/google/minijail/commit/a8eef8bfba354e94e8be2d9814c9e6ece14910eb), which changes that behavior. It wasn't released yet, but once it is, we probably have to add `/bin` to the `BINDMOUNT_ALLOWED_PREFIXES` variable when building minijail.

adombeck (approved):

PR #311, opened 2022-09-20
Various fixes in preparation for crash-resistant coverage

fmeum:


internal/build/other/other.go
```diff
@@ -262,6 +262,16 @@ func (b *Builder) setCoverageEnv() error {
 		return err
 	}
 
+	ldflags := []string{
+		// ----- Flags used to link in coverage runtime -----
+		"-fprofile-instr-generate",
+		"-fcoverage-mapping",
+	}
+	b.env, err = envutil.Setenv(b.env, "LDFLAGS", strings.Join(ldflags, " "))
```
jochil, line 270: This overwrites existing LDFLAGS, or? Is this ok here?

jochil (approved):

internal/build/other/other.go
fmeum: That's a good point. I think we are already overriding `CFLAGS` in the same way, so I'm inclined to answer `yes` for now. We should revisit that, but I'm not sure we can reliably work with someone else injecting their `CFLAGS`/`LDFLAGS`.

PR #312, opened 2022-09-21
installer: Support CIFUZZ_INSTALLER_NO_CMAKE env var

adombeck:


cmd/installer/installer.go
```diff
@@ -145,6 +145,13 @@ func ExtractEmbeddedFiles(targetDir string, files *embed.FS) error {
 		return err
 	}
 
+	// Support not copying and registering the CMake package. One use
+	// case for this is to install cifuzz inside a sandbox which doesn't
+	// allow access to the CMake package directories.
+	if os.Getenv("CIFUZZ_INSTALLER_NO_CMAKE") != "" {
+		return nil
```
jochil, line 152: IMO it would be more reliable to put the CMake stuff inside this if block (or move it to a function) instead of just returning here.... as this is basically a "skip everything after this line" (which can change if someone add stuff below) and not a "skip cmake"

adombeck: done

jochil (approved):

PR #313, opened 2022-09-21
feat: "cifuzz init" for maven & gradle

mlsandnerCI:
Init command for maven & gradle with tests + refactoring of integration tests.


internal/cmd/init/instructions/gradle
```diff
@@ -0,0 +1,12 @@
+
+Please make sure to add the following dependencies to your
+build.gradle to enable fuzz testing:
+
+   implementation 'com.code-intelligence:jazzer-api:0.12.0'
```
fmeum, line 5: This is a dependency of `jazzer-junit`, so we might be able to just leave it out.

internal/cmd/init/instructions/gradle
```diff
@@ -0,0 +1,12 @@
...
+   implementation 'com.code-intelligence:jazzer-junit:0.12.0'
```
fmeum, line 6: This could be `testImplementation`, which excludes it from the production classpath.

internal/cmd/init/instructions/maven
```diff
@@ -0,0 +1,27 @@
...
+        <groupId>com.code-intelligence</groupId>
+        <artifactId>jazzer-junit</artifactId>
+        <version>0.12.0</version>
+        <scope>test</scope>
+    </dependency>
+    <dependency>
+        <groupId>com.code-intelligence</groupId>
+        <artifactId>jazzer-api</artifactId>
+        <version>0.12.0</version>
+        <scope>compile</scope>
```
fmeum, line 15: Same here, this dependency might not be necessary for this to work.

internal/cmd/init/instructions/gradle
mlsandnerCI: Okay, i will remove it. We added it in the example `build.gradle` because we couldn't use `com.code_intelligence.jazzer.api.FuzzerSecurityIssueMedium` otherwise since it couldn't be resolved and we thought everyone would need it.

internal/testutil/example_project.go
```diff
@@ -10,20 +10,20 @@ import (
 	"code-intelligence.com/cifuzz/util/testutil"
 )
 
-// Creates a copy of the CMake example project in a temp dir and
-// change the working dir to the new path
-func ChdirToClonedCmakeExampleProject(prefix string) (string, error) {
+// CopyExampleProjectAndChDir copies the given example project to a temporary folder
+// and changes into that directory.
+func CopyExampleProjectAndChDir(prefix, exampleName string) (string, error) {
```
jochil, line 15: `CopyExampleProjectAndChDir` implies that the example project is copied to an existing directory (I would expect the first param to be the target). What's about `BootstrapExampleProjectForTest`

integration-tests/gradle/gradle_test.go
```diff
@@ -0,0 +1,41 @@
...
+	projectDir, err := testutil.CopyExampleProjectAndChDir("cifuzz-gradle-example", config.BuildSystemGradle)
+	require.NoError(t, err)
+
+	// Remove cifuzz.yaml from example to test init command
+	err = os.RemoveAll(filepath.Join(projectDir, "cifuzz.yaml"))
+	require.NoError(t, err)
+	defer fileutil.Cleanup(projectDir)
+
+	// Execute the init command
+	cmd := executil.Command(cifuzz, "init", "-C", projectDir)
```
jochil, line 32: Is the `-C` necessary after changing the working directroy in  `CopyExampleProjectAndChDir`?

integration-tests/gradle/gradle_test.go
```diff
@@ -0,0 +1,41 @@
...
+	installDir := testutil.InstallCifuzzInTemp(t)
+	cifuzz := builderPkg.CIFuzzExecutablePath(filepath.Join(installDir, "bin"))
+
+	projectDir, err := testutil.CopyExampleProjectAndChDir("cifuzz-gradle-example", config.BuildSystemGradle)
+	require.NoError(t, err)
+
+	// Remove cifuzz.yaml from example to test init command
+	err = os.RemoveAll(filepath.Join(projectDir, "cifuzz.yaml"))
+	require.NoError(t, err)
+	defer fileutil.Cleanup(projectDir)
```
jochil, line 29: I would move this up to L25... so the cleanup of the project dir is close to its creation

integration-tests/gradle/gradle_test.go
```diff
@@ -0,0 +1,41 @@
...
+	}
+
+	installDir := testutil.InstallCifuzzInTemp(t)
+	cifuzz := builderPkg.CIFuzzExecutablePath(filepath.Join(installDir, "bin"))
+
+	projectDir, err := testutil.CopyExampleProjectAndChDir("cifuzz-gradle-example", config.BuildSystemGradle)
+	require.NoError(t, err)
+
+	// Remove cifuzz.yaml from example to test init command
+	err = os.RemoveAll(filepath.Join(projectDir, "cifuzz.yaml"))
```
jochil, line 27: I think `os.Remove()` is enough for a single file

integration-tests/maven/maven_test.go
```diff
@@ -0,0 +1,43 @@
...
+	// Execute the init command
+	cmd := executil.Command(cifuzz, "init", "-C", projectDir)
+	cmd.Dir = projectDir
+	stderrPipe, err := cmd.StderrTeePipe(os.Stderr)
+	require.NoError(t, err)
+	t.Logf("Command: %s", cmd.String())
+	err = cmd.Run()
+	require.NoError(t, err)
+	err = stderrPipe.Close()
+	require.NoError(t, err)
```
jochil, line 42: Maybe we can move these "command execution" blocks to a single parameterized `t.Helper` function... especially for the cmake/make integration tests. What do you think?

integration-tests/gradle/gradle_test.go
```diff
@@ -0,0 +1,41 @@
...
+	cmd := executil.Command(cifuzz, "init", "-C", projectDir)
+	cmd.Dir = projectDir
+	stderrPipe, err := cmd.StderrTeePipe(os.Stderr)
+	require.NoError(t, err)
+	t.Logf("Command: %s", cmd.String())
+	err = cmd.Run()
+	require.NoError(t, err)
+	err = stderrPipe.Close()
+	require.NoError(t, err)
+}
```
jochil, line 41: So we are testing for errors but not for the actual output of the command... the creation of the `cifuzz.yaml`.
I think `assert.FileExists(t, filepath.Join(projectDir, "cifuzz.yaml")` would do the job

jochil: The cmake integration tests are going even one step further by adding the "code" printed out by `cifuzz init` to the `CMakeLists.txt`... maybe we should consider that too

internal/testutil/example_project.go
mlsandnerCI: Good point!

integration-tests/gradle/gradle_test.go
mlsandnerCI: Test works without 👍

integration-tests/maven/maven_test.go
mlsandnerCI: We thought about this but the cmake test includes a `followStepsPrintedByInitCommand()` function which uses the `stderrPipe` from the `init` command and since the steps are different for the build systems so it didn't seem universal. 

I could let the function return the pipe but then it isn't sure that we won't forget to close it when calling the function elsewhere. Any ideas how to solve this elegantly?

internal/cmd/init/instructions/gradle
fmeum: I forgot about this class - at least for now you are right to depend on `jazzer-api` for it. Just make it a `testImplementation` dependency.

mlsandnerCI: So we should keep it in the description or only in the example project?

mlsandnerCI: And the gradle build fails if I set the `jazzer-api `as `testImplementation` instead of `implementation` because it can't resolve the `FuzzerSecurityIssueMedium()` function. 
Could we maybe look over this together late? I seem to have trouble to understand the difference and how you would expect it to work.

integration-tests/gradle/gradle_test.go
mlsandnerCI: Right now we copy the example project that already has these lines in the pom.xml. Would it be better to create a different example for testing like the cmake integration test?

integration-tests/maven/maven_test.go
jochil: The helper function could read the pipe and return a `string` instead. Something like
```go
func RunCommand(args []string, t *testing.T) string {
}
```

jochil:
> @florianGla and I are unsure of the last commit `test: maven & gradle integration` is okay because it also refactors the existing integration tests. Should we leave it or split it up? Opinions?

A single commit would be nicer... especially if you refactor more (the discussion above)

jochil (approved):

PR #317, opened 2022-09-22
fix(dependency): loosen the filename check for clang in CC/CXX

jochil:
... to cover cases like `clang++-13' or with prefixes.


pkg/dependencies/version.go
```diff
@@ -46,7 +47,7 @@ func clangVersion(dep *Dependency, clangCheck execCheck) (*semver.Version, error
 	checkPath := false
 
 	if cc, found := os.LookupEnv("CC"); found {
-		if path.Base(cc) == "clang" {
+		if strings.Contains(path.Base(cc), "clang") {
```
fmeum, line 50: I think we could start with `strings.HasPrefix`, which should be loose enough.

pkg/dependencies/version.go
```diff
@@ -61,7 +62,7 @@ func clangVersion(dep *Dependency, clangCheck execCheck) (*semver.Version, error
 	}
 
 	if cxx, found := os.LookupEnv("CXX"); found {
-		if path.Base(cxx) == "clang++" {
+		if strings.Contains(path.Base(cxx), "clang") {
```
fmeum, line 65: Should we still match `clang++` here?

fmeum (approved):

fmeum:
I added https://github.com/CodeIntelligenceTesting/cifuzz/pull/315/commits/164685e05e8d2c515e47b43ad2caa322d69d89df to my PR

pkg/dependencies/version.go
jochil: It is the second time I adjust this check, so I wanted to make sure it is the last time :see_no_evil: 
I can easily image people having self compiled clang version with prefixes

pkg/dependencies/version.go
jochil: see above... the check was too strict, again. Now I would prefer to have it a little bit more tolerant :)

PR #324, opened 2022-09-26
Fix usage message and reuse flags in multiple commands

adombeck:


florianGla (approved):

internal/bundler/bundler.go
```diff
@@ -67,21 +67,22 @@ func versionedLibraryRegexp(unversionedBasename string) *regexp.Regexp {
 }
 
 type Opts struct {
+	Branch         string        `mapstructure:"branch"`
 	BuildSystem    string        `mapstructure:"build-system"`
 	NumBuildJobs   uint          `mapstructure:"build-jobs"`
+	Commit         string        `mapstructure:"commit"`
```
jochil, line 73: I left `Branch` and `Commit` out of the config file on purpose. I think it does not make any sense to define a fixed commit and/or branch in the config file. 
So maybe using `mapstructure:"-"` ? Or do you think there is a valid scenario for this?

internal/cmd/finding/finding_cmd.go
```diff
@@ -29,47 +30,52 @@ type findingCmd struct {
...
+				// printing a stack trace
+				log.Error(err, fmt.Sprintf("%s\nUse 'cifuzz init' to set up a project for use with cifuzz.", err.Error()))
+				return cmdutils.WrapSilentError(err)
+			}
+			if err != nil {
+				log.Errorf(err, "Failed to parse cifuzz.yaml: %v", err.Error())
+				return cmdutils.WrapSilentError(err)
+			}
+			opts.ProjectDir = projectDir
+			return nil
```
jochil, line 59: Is this `projectDir` change related to the rest of this commit... if yes, I do not get how  :thinking:

internal/bundler/bundler.go
adombeck: This allows the user to set the branch and commit via environment variables, i.e. `CIFUZZ_BRANCH` or `CIFUZZ_COMMIT`. Since we use viper for both the config file parsing and the environment variables, we can't support one without also supporting the other.

jochil: Are you sure? I tested it and it was working fine with environment variables... you can try `CIFUZZ_COMMIT=123 go run cmd/cifuzz/main.go -C ./examples/cmake bundle my_fuzz_test --branch my-branch` on `main` and check the artifact config... it will contain `123` as commit

internal/cmd/finding/finding_cmd.go
adombeck: It is related, because `ParseProjectConfig(opts)` sets the fields of `opts` to the viper values, and with this commit the `--json` flag is not bound to a field of the `opts` struct directly but only to the viper value. And since `ParseProjectConfig` already returns the project dir, it makes sense to use that and not call `FindProjectDir()` again in the `run()` function,

internal/bundler/bundler.go
adombeck: That did work before this commit, because the `branch` and `commit` flags were bound to the `opts.Branch` and `opts.Commit` fields directly via `cmd.Flags().StringVar`. To be able to share these flags with other commands, we now use `cmd.Flags().String` instead and the `opts.Branch`/`opts.Commit` fields are set by `ParseProjectConfig`.

I realize that this isn't a perfect solution, because before we could support environment variables without also supporting config files. We might be able to solve this with reflection, by changing `AddBranchFlag()` to receive an opts struct (`AddBranchFlag(opts interface{})`) and use reflection to bind the flag directly to the `Branch` field.

adombeck: I checked again, `Branch` and `Commit` are actually configurable via the config file on main, because you did not use `mapstructure:"-"` but left the fields untagged instead. Try:
```
echo "Branch: foo" > examples/cmake/cifuzz.yaml
cifuzz -C examples/cmake bundle my_fuzz_test
```

When setting `mapstructure:"-"` on main, setting the field via the environment variable also doesn't work anymore.

adombeck: > We might be able to solve this with reflection, by changing AddBranchFlag() to receive an opts struct (AddBranchFlag(opts interface{})) and use reflection to bind the flag directly to the Branch field.

I tried that and it didn't work. As long as `mapstructure:"-"` is set, the field is not set via environment variables or the config file. If the tag is omitted, it can be set via both environment variables and the config file.

adombeck: This is blocking other work of mine and since it's not a regression, I'll merge this as is now. We can later revisit the use case of options which should be configurable via environment variables and not via the config file.

fmeum: Sounds good! Having `branch` configurable in the YAML may even make sense.

PR #326, opened 2022-09-26
feat: maven & gradle integration in 'create' command

mlsandnerCI:
Integrating maven & gradle build systems into the create command with their own template in .java and associated tests.

Plus 2x refactoring:
- Extracting a method out of the cmake integration test to use it in the maven & gradle integration tests.
- I changed if-statements that check for the build system or file type to switch/cases so that we can expand them easily if we need to add instructions or dependencies for other build systems at some point


pkg/stubs/stubs.go
```diff
@@ -15,6 +15,9 @@ import (
 //go:embed fuzz-test.cpp.tmpl
 var cppStub []byte
 
+//go:embed FuzzTestCase.java.tmpl
+var javaStub []byte
+
 // Create creates a stub based for the given test type
 func Create(path string, testType config.FuzzTestType) error {
```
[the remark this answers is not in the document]
mlsandnerCI: Good catch, thank you!

fmeum:
> Never seen this before. Is a new line at the end of the file (here the template for the test) important?

It's not very important, but still a good practice as it keeps git diffs cleaner (otherwise adding a line in the end would also add a line break to the previous line).

pkg/stubs/FuzzTestCase.java.tmpl
```diff
@@ -0,0 +1,15 @@
+import com.code_intelligence.jazzer.api.FuzzedDataProvider;
+import com.code_intelligence.jazzer.junit.FuzzTest;
+
+class FuzzTestCase {
```
fmeum, line 4: The Java test class for the regular class `MyClass` is usually called `MyClassTest`, wo how about calling the stub class `MyClassFuzzTest`?

fmeum (approved):

internal/cmd/create/create.go
```diff
@@ -152,11 +153,16 @@ and $FUZZ_TEST_LDFLAGS to the linker.`)
...
 		deps = append(deps, dependencies.CLANG)
 	}
-	if c.config.BuildSystem == config.BuildSystemCMake {
+
+	switch c.config.BuildSystem {
+	case config.BuildSystemCMake:
 		deps = append(deps, dependencies.CMAKE)
 	}
+
 	return dependencies.Check(deps, dependencies.Default, runfiles.Finder)
```
florianGla, line 167: Maybe we could check for maven/gradle as a dependency

fmeum: I don't think we should do this as we don't need a particular version (yet, but also potentially never). I think that we can assume that users installed *some version* of the build system they are using.

pkg/stubs/FuzzTestCase.java.tmpl
mlsandnerCI: Should the file name then also be `MyClassFuzzTest.java`?

fmeum: If the filename is user-provided, then we should ensure that the class name in the template matches the filename. For `public` classes this is a hard requirement, for source files with a single class it's still common practice.

mlsandnerCI: Okay, so I added extra logic that if an output path was given, we change the template class name to the filename but I'm unsure if this is a good way. And the question is, what would happen if the given output path is only to a directory and not a class file. I don't think we check that right now.

pkg/stubs/stubs.go
```diff
@@ -30,6 +34,11 @@ func Create(path string, testType config.FuzzTestType) error {
 	switch testType {
 	case config.CPP:
 		content = cppStub
+	case config.JAVA:
+		content = javaStub
+		if customFileName {
+			strings.Replace(string(javaStub), "MyClassFuzzTest", filepath.Base(path), 1)
```
fmeum, line 40: I would prefer a placeholder that stands out (e.g. `__CLASS_NAME__`) that we always replace instead of only in the case of a custom file name.

mlsandnerCI: Good point. I also found an error, I forgot to assign the result of strings.Replace... I wrote a test to ensure this works correctly.

pkg/stubs/stubs.go
```diff
@@ -30,6 +34,9 @@ func Create(path string, testType config.FuzzTestType) error {
 	switch testType {
 	case config.CPP:
 		content = cppStub
+	case config.JAVA:
+		fileName := strings.TrimSuffix(filepath.Base(path), ".java")
+		content = []byte(strings.Replace(string(javaStub), "__CLASS_NAME__", fileName, 1))
```
fmeum, line 39: I would suggest `basename` instead of `fileName`:
```suggestion
		basename := strings.TrimSuffix(filepath.Base(path), ".java")
		content = []byte(strings.Replace(string(javaStub), "__CLASS_NAME__", basename, 1))
```

fmeum (approved):

PR #327, opened 2022-09-26
cmd/remote-run: Support existing artifacts

adombeck:


internal/cmd/remote-run/remote_run.go
```diff
@@ -123,6 +131,7 @@ https://github.com/CodeIntelligenceTesting/cifuzz/issues`, system)
 	cmd.Flags().StringVar(&opts.Branch, "branch", "", "Branch name to use in the artifacts config. By default, the currently checked out git branch is used.")
 	cmd.Flags().StringVar(&opts.Commit, "commit", "", "Commit to use in the artifacts config. By default, the head of the currently checked out git branch is used.")
 
+	cmd.Flags().StringVar(&opts.ArtifactsPath, "artifacts", "", "Path of an existing artifacts archive to start a remote run with.")
```
fmeum, line 134: I find the flag name confusing as it's plural, but we are asking the user to pass in a single thing. How about calling it `--archive` or `--archive-path` or `--existing-archive`?

internal/cmd/remote-run/remote_run.go
```diff
@@ -67,7 +70,12 @@ func newWithOptions(opts *remoteRunOpts) *cobra.Command {
 		Short: "Build fuzz tests and run them on a remote fuzzing server",
 		Long: `This command builds fuzz tests, bundles all runtime artifacts into a
 self-contained archive and uploads that to a remote fuzzing server to
-start a remote fuzzing run.`,
+start a remote fuzzing run.
+
+If the --artifacts flag is used, building and bundling is skipped and
+instead the specified artifacts archive is uploaded to start a remote
+fuzzing run.
+`,
```
fmeum, line 78: This sentence flows better if `instead` is moved to the end.

fmeum (approved):

internal/cmd/remote-run/remote_run.go
adombeck: done

internal/cmd/remote-run/remote_run.go
adombeck: Since we called the command which creates this artifacts archive `bundle`, maybe we should commit to that name and call the produced file an "artifacts bundle" or "fuzzing bundle". WDYT? Probably something we should discuss in the team.

adombeck: I would call the flag `--bundle` if we decide that we want to use that name for the file.

fmeum: `--bundle` is fine with me, I just don't like `--artifacts`.

adombeck: done

PR #330, opened 2022-09-27
Support specifying the project dir

adombeck:


jochil:
This has some really nice cleanups & refactorings :+1:

internal/config/config.go
```diff
@@ -64,9 +64,18 @@ func CreateProjectConfig(configDir string) (string, error) {
 }
 
 func FindAndParseProjectConfig(opts interface{}) error {
```
jochil, line 66: I am not a big fan of `interace{}`... what's about having a Config struct with the minimal necessary fields (I guess `ConfigDir` and `ProjectDir`) and embedding it in the command option structs. Then we could use a "real" type for this function and don't have to rely on `reflect`

adombeck: We can't use a struct type here, because this function is supposed to receive various different structs (e.g. `runOptions`, `bundler.Opts` etc.). The fields of those structs are set by the call to `viper.Unmarshal`. I admit this is a bit obscure but I'm afraid that's the price for having viper take care of all the config stuff for us.

jochil: I understand... that's a pity, but ok

jochil (approved):

PR #332, opened 2022-09-27
cmd/finding: Remove the --all flag

adombeck:
The name of the flag did not fit well, because `cifuzz findings` already lists all findings (but only the summary) and so `cifuzz findings --json` already prints all findings with all details (as JSON).

We suspect that `cifuzz findings --json` actually already covers the use cases of this flag. Users can pipe the output to jq and filter for the fields they are actually interested in.


adombeck:
@pnoltof Since you requested the functionality of the `--all` flag, do you think you can use `--json` instead - possibly piped through a `jq` command?

jochil (approved):

adombeck:
> @pnoltof Since you requested the functionality of the --all flag, do you think you can use --json instead - possibly piped through a jq command?

@pnoltof ping

pnoltof:
@adombeck Yes, that should be fine. Thanks!

PR #336, opened 2022-09-29
feat: print jazzer findings correctly

mlsandnerCI:
Added a new regex pattern to correctly read java stacktraces and display crash findings.

When creating the short description I deliberately wrote one case for jazzer and one for other "special" findings eventhough they execute the same code to make it more readable. As I understand this is only temporary anyway because this is a naive approach and has to refactored at some point.
```
switch {
   case f.Details == "detected memory leaks":
	// Special vulnerabilities
	errorType = f.Details
   case strings.Contains(f.Details, "Security Issue:"):
	// Jazzer findings
	errorType = f.Details
   default:
	errorType = strings.ReplaceAll(strings.Split(f.Details, " ")[0], "-", " ")
}
```


internal/cmd/run/report_handler/stacktrace/stacktrace.go
```diff
@@ -108,7 +112,10 @@ func (p *parser) parseSourceLocation(logs []string) ([]*StackFrame, error) {
 func (p *parser) stackFrameFromLine(line string) (*StackFrame, error) {
 	matches, found := regexutil.FindNamedGroupsMatch(framePattern, line)
 	if !found {
-		return nil, nil
+		matches, found = regexutil.FindNamedGroupsMatch(framePatternJava, line)
+		if !found {
+			return nil, nil
+		}
```
adombeck, line 118: I would like to avoid matching regexes for which we know that they shouldn't match - in this case trying to match a Java stack trace when the output comes from libFuzzer and not Jazzer.

I think we should refactor this stack trace parsing a bit. It should probably be done by the parser and not the report handler - and the parser already knows if it should support Jazzer or not.

adombeck: > I think we should refactor this stack trace parsing a bit. It should probably be done by the parser and not the report handler - and the parser already knows if it should support Jazzer or not.

FTR, we did that in a pairing session

pkg/parser/libfuzzer/stacktrace/stacktrace.go
```diff
@@ -106,7 +115,13 @@ func (p *parser) parseSourceLocation(logs []string) ([]*StackFrame, error) {
...
 
 func (p *parser) stackFrameFromLine(line string) (*StackFrame, error) {
-	matches, found := regexutil.FindNamedGroupsMatch(framePattern, line)
+	var matches map[string]string
+	var found bool
+	if p.SupportJazzer {
+		matches, found = regexutil.FindNamedGroupsMatch(framePatternJava, line)
+	} else {
+		matches, found = regexutil.FindNamedGroupsMatch(framePattern, line)
+	}
```
adombeck, line 124: @fmeum: Should we also check the non-Java frame pattern here if the logs come from Jazzer? I.e. is it possible that Jazzer findings contain non-Java stack traces?

fmeum: Jazzer findings only contain Java stack frames, but Jazzer can produce e.g. ASan findings.

adombeck: @mlsandnerCI I pushed a fixup commit (because I didn't want to force-push to your branch without you knowing). Please squash that one into the `stacktrace: Only try to parse as Java stack trace if logs come from Jazzer` commit.

jochil (approved):

pkg/parser/libfuzzer/stacktrace/stacktrace.go
```diff
@@ -14,6 +14,10 @@ import (
 var framePattern = regexp.MustCompile(
 	`#(?P<frame_number>\d+)\s+0x[a-fA-F0-9]+\s+in\s+(?P<function>(\(anonymous namespace\))?[^(\s]+).*\s(?P<source_file>\S+?):(?P<line>\d+):?(?P<column>\d*)`)
 
+// Special pattern for Java stack traces which does not filter out java base classes.
+// This has to checked elsewhere since go regex does not allow lookups to filter out specific words.
```
adombeck, line 18: I suggest we move the explanation about Java base classes to the code where we actually exclude stack frames of Java base classes (see below)
```suggestion
// Special pattern for Java stack traces
```

pkg/parser/libfuzzer/stacktrace/stacktrace.go
adombeck, line 124: It looks like you accidentally removed my fixup commit instead of squashing it. Please squash this change into the `refactor: Parse stack trace in libFuzzer parser` commit.

```suggestion
	matches, found := regexutil.FindNamedGroupsMatch(framePattern, line)
        if !found && p.SupportJazzer {
                matches, found = regexutil.FindNamedGroupsMatch(framePatternJava, line)
                if !found {
                        return nil, nil
                }
        }
```

pkg/parser/libfuzzer/stacktrace/stacktrace.go
```diff
@@ -186,6 +204,13 @@ func (p *parser) validateSourceFile(path string) (string, error) {
 		return "", nil
 	}
 
+	if p.SupportJazzer {
+		// Ignore files from the java standard library
+		if strings.Contains(path, "java.base") {
+			return "", nil
+		}
```
adombeck, line 211:
```suggestion
		// Ignore files from the Java standard library. We can't filter
		// these out via the regex because go regex doesn't support
		// lookups to filter out specific words.
		if strings.Contains(path, "java.base") {
			return "", nil
		}
```

pkg/parser/libfuzzer/stacktrace/stacktrace.go
mlsandnerCI: I'm sorry, I was so sure I squashed it in, will fix it right away!

adombeck:
> There is a problem that two libfuzzer parser tests (`metric line in the middle of a report` and `segfault at the end`) print stacktraces in windows but not on the other OS'. So if I add in the stacktraces in the expected reports it fails on macOS and Linux but if I don't add them in, it fails on Windows. Any idea how to elegantly solve this?

I pushed a commit which should fix that the parser creates a stack trace from the provided logs on Windows.

PR #337, opened 2022-09-29
examples: Use same project structure for maven and gradle

florianGla:


examples/gradle/src/main/java/com/github/CodeIntelligenceTesting/cifuzz/App.java
```diff
@@ -1,7 +1,4 @@
-/*
- * This Java source file was generated by the Gradle 'init' task.
- */
-package gradle;
+package com.github.CodeIntelligenceTesting.cifuzz;
```
jochil, line 1: The new package name for the `App` class has to be updated in the build.gradle file at
```
application {
    // Define the main class for the application.
    mainClass = 'gradle.App'
}
```

jochil (approved):

PR #343, opened 2022-09-30
all: Only build and run with continuous mode if supported

fmeum:
LLVM's continuous mode requires compile-time instrumentation on Linux, which unfortunately isn't all that stable in Clang 13 and below. Since continuous mode is not crucial, we fall back to disabling it with lower Clang versions.


internal/cmd/coverage/coverage.go
```diff
@@ -691,3 +699,41 @@ func cpuToArchFlag(cpu macho.Cpu) (string, error) {
...
+	for _, symbol := range symbols {
+		if symbol.Name == "__llvm_profile_counter_bias" {
+			biasVarAddress = symbol.Value
+		} else if symbol.Name == "__llvm_profile_counter_bias_default" {
+			biasDefaultVarAddress = symbol.Value
+		}
+	}
+	// Check taken from:
+	// https://github.com/llvm/llvm-project/blob/846709b287abe541fcad42e5a54d37a41dae3f67/compiler-rt/lib/profile/InstrProfilingFile.c#L574
+	return biasVarAddress != biasDefaultVarAddress
```
adombeck, line 738: Is it expected that this returns true if there are no symbols with names `__llvm_profile_counter_bias` and `__llvm_profile_counter_bias_default` and therefore both `biasVarAddress` and `biasDefaultVarAddress` are 0?

adombeck (approved):

internal/cmd/coverage/coverage.go
fmeum: This can never happen with coverage builds as the profile runtime always contains these two symbols. But that may be confusing, I will add an explicit check.

PR #347, opened 2022-10-04
Add run command for maven/gradle

florianGla:


internal/builder/builder.go
```diff
@@ -184,12 +184,7 @@ func (i *CIFuzzBuilder) BuildCIFuzzAndDeps() error {
 		return err
 	}
 
-	err = i.CopyVSCodeTasks()
-	if err != nil {
-		return err
-	}
-
-	err = i.CopyLogo()
```
jochil, line 192: I would guess we can delete the `CopyLogo` function too, or?

florianGla: Yes, you are right

pkg/dependencies/definitions_test.go
```diff
@@ -9,12 +9,12 @@ import (
 )
 
 func TestResetDefault(t *testing.T) {
-	oldDefault := Default
-	Default = Dependencies{}
+	oldDefault := CMakeDeps
```
jochil, line 12: can you also add some assertions for the new `MavenDeps` and `GradleDeps` to this test?

pkg/dependencies/deps.go
```diff
@@ -17,6 +17,9 @@ const (
 	LLVM_SYMBOLIZER Key = "llvm-symbolizer"
 	LLVM_PROFDATA   Key = "llvm-profdata"
 
+	MAVEN  Key = "maven"
```
jochil, line 20: for all other cases the key is also the name of the executable... what do you think about using `mvn` here?

jochil:
@florianGla @fmeum could you please outline the changes that are necessary after the jazzer refactoring is done? Maybe we can mark them in the code with some `TODO`s. IMO it would help us later

internal/build/gradle/gradle.go
```diff
@@ -0,0 +1,178 @@
...
+	"code-intelligence.com/cifuzz/internal/build"
+	"code-intelligence.com/cifuzz/internal/cmdutils"
+	"code-intelligence.com/cifuzz/pkg/log"
+	"code-intelligence.com/cifuzz/pkg/runfiles"
+	"code-intelligence.com/cifuzz/util/fileutil"
+)
+
+var classpathRegex = regexp.MustCompile("(?m)^Test classpath = (?P<classpath>.*)$")
+
+func FindGradleWrapper(projectDir string) (string, error) {
```
jochil, line 23: In the gradle docs (https://docs.gradle.org/current/userguide/gradle_wrapper.html) it sounds like the Wrapper is recommended but not necessary? Do you think we need a fallback to the `gradle` command if the wrapper is missing?

jochil:
@florianGla can you also update the READMEs of the maven/gradle example projects? At least the neccessary `cifuzz run` command would be helpful

internal/build/gradle/gradle.go
```diff
@@ -0,0 +1,178 @@
...
+	output, err := cmd.Output()
+	if err != nil {
+		// It's expected that gradle might fail due to user configuration,
+		// so we print the error without the stack trace.
+		err = cmdutils.WrapExecError(err, cmd)
+		log.Error(err)
+		return nil, cmdutils.ErrSilent
+	}
+	classpath := classpathRegex.FindStringSubmatch(string(output))
+	deps := strings.Split(string(classpath[1]), ":")
```
fmeum, line 153: The class path separator is `os.PathListSeparator`, which is `;` on Windows.

internal/build/gradle/gradle.go
```diff
@@ -0,0 +1,178 @@
...
+
+	return deps, nil
+}
+
+func (b *Builder) getFuzzTestSeedCorpus(targetClass string) string {
+	path := strings.Split(targetClass, ".")
```
fmeum, line 159: We currently store the seed corpus in `<ClassName>SeedCorpus`, so the last segment has to be modified.

internal/build/gradle/gradle.go
```diff
@@ -0,0 +1,178 @@
...
+func (b *Builder) getGradleCommand() (string, error) {
+	wrapper, err := FindGradleWrapper(b.ProjectDir)
+	if err != nil {
+		return "", err
+	}
+	if wrapper != "" {
+		return wrapper, nil
+	}
+
+	return "gradle", nil
```
fmeum, line 177: Shouldn't we use the runfiles finder here?

internal/build/maven/maven.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+
+	bytes, err := os.ReadFile(outputPath)
+	if err != nil {
+		return nil, errors.WithStack(err)
+	}
+
+	deps := strings.Split(string(bytes), ":")
+	// Append dependencies which are not listed by "mvn dependency:build-classpath"
+	localDeps := []string{"target/classes", "target/test-classes", "resource", "test-resource"}
```
fmeum, line 140: These should use `filepath.Join`. Are the third and fourth entries correct as they are or do they also lie under `target`?

internal/build/maven/maven.go
```diff
@@ -0,0 +1,153 @@
...
+	for _, dep := range localDeps {
+		deps = append(deps, filepath.Join(b.ProjectDir, dep))
+	}
+
+	return deps, nil
+}
+
+func (b *Builder) getFuzzTestSeedCorpus(targetClass string) string {
+	path := strings.Split(targetClass, ".")
```
fmeum, line 149: See the commend on Gradle. Could we extract this method?

internal/build/maven/maven.go
florianGla: You are right, I think they also lie under target.

fmeum:
> @florianGla @fmeum could you please outline the changes that are necessary after the jazzer refactoring is done? Maybe we can mark them in the code with some `TODO`s. IMO it would help us later

There will be smaller changes required in multiple places, but nothing that changes the overall architecture. I would rather not add comments describing a version that hasn't been released yet - we may make further changes to it and the current approach will continue to work.

integration-tests/gradle/gradle_test.go
```diff
@@ -31,7 +38,18 @@ func TestIntegration_Gradle(t *testing.T) {
...
 
 	// Execute the create command
-	outputPath := filepath.Join("src", "MyClassFuzzTest.java")
+	testDir := filepath.Join(
+		"src",
+		"test",
+		"java",
+		"com",
+		"github",
+		"CodeIntelligenceTesting",
```
fmeum, line 47: Package names should always be in lower case. You can use `com/example` or `com/github/code_intelligence`.

integration-tests/maven/maven_test.go
```diff
@@ -1,7 +1,14 @@
 package maven
 
 import (
+	"bufio"
+	"code-intelligence.com/cifuzz/pkg/parser/libfuzzer/stacktrace"
```
fmeum, line 5: This should be reformatted with `make fmt`. You can install the push hook in `githooks` to fix this automatically.

integration-tests/gradle/gradle_test.go
florianGla: Currently our gradle fuzz test class is located under cifuzz/examples/gradle/src/main/java/com/github/CodeIntelligenceTesting/cifuzz. Shall we change this path and the package name accordingly?

fmeum: Yes, we should move it to a more common package.

jochil:
> > @florianGla @fmeum could you please outline the changes that are necessary after the jazzer refactoring is done? Maybe we can mark them in the code with some `TODO`s. IMO it would help us later
> 
> There will be smaller changes required in multiple places, but nothing that changes the overall architecture. I would rather not add comments describing a version that hasn't been released yet - we may make further changes to it and the current approach will continue to work.

IIRC the current approach requires the jazzer driver to be installed manually. Or Am I wrong @florianGla?
It will not be shipped with this change... so from user perspective this is not a working change

florianGla:
> IIRC the current approach requires the jazzer driver to be installed manually. Or Am I wrong @florianGla? It will not be shipped with this change... so from user perspective this is not a working change

You are right, I think we should wait for the jazzer standalone package before merging.

internal/build/build.go
```diff
@@ -67,3 +68,11 @@ func CommonBuildEnv() ([]string, error) {
 
 	return env, nil
 }
+
+func JazzerSeedCorpus(targetClass string, projectDir string) string {
+	seedCorpus := targetClass + "SeedCorpus"
```
jochil, line 73: I think the Suffix is `Inputs` instead of `SeedCorpus`... at least after the jazzer changes from @fmeum

MarkusZoppelt: Changed!

fmeum:
Based on the log output it looks like our fancy printer may have erased some parts of Jazzer's exception:
```
2022-10-27T13:07:48.2765717Z 
2022-10-27T13:07:48.2765961Z remote code execution depending on available classes on the classpath.eHigh: Remote Code Execution
2022-10-27T13:07:48.2766406Z Unrestricted class loading based on externally controlled data may allow
2022-10-27T13:07:48.2766788Z remote code execution depending on available classes on the classpath.
2022-10-27T13:07:48.2767118Z 	at jaz.Zer.<clinit>(Zer.java:54)
2022-10-27T13:07:48.2767447Z 	at java.base/java.lang.Class.forName0(Native Method)
2022-10-27T13:07:48.2767811Z 	at java.base/java.lang.Class.forName(Class.java:315)
2022-10-27T13:07:48.2768191Z 	at com.example.ExploreMe.exploreMe(ExploreMe.java:13)
2022-10-27T13:07:48.2768606Z 	at com.example.FuzzTestCase.myFuzzTest(FuzzTestCase.java:19)
```

@adombeck Do you see how this could happen?

pkg/runfiles/finder.go
```diff
@@ -55,6 +48,12 @@ func (f RunfilesFinderImpl) LLVMSymbolizerPath() (string, error) {
 	return path, errors.WithStack(err)
 }
 
+func (f RunfilesFinderImpl) JavaPath() (string, error) {
```
fmeum, line 51: Is this even used?

pkg/runner/jazzer/jazzer_runner.go
```diff
@@ -63,16 +63,18 @@ func (r *Runner) Run(ctx context.Context) error {
...
-	jazzerPath, err := runfiles.Finder.JazzerPath()
-	if err != nil {
-		return err
-	}
-	agentPath, err := runfiles.Finder.JazzerAgentDeployJarPath()
+	javaHome, err := runfiles.Finder.JavaHomePath()
 	if err != nil {
 		return err
 	}
+	javaBin := filepath.Join(javaHome, "bin", "java")
```
fmeum, line 70: We have to append `.exe` on Windows.

pkg/runfiles/finder.go
florianGla: No, I think were are looking at an old version

florianGla: You are right, I forgot to delete old functions

internal/build/maven/maven.go
```diff
@@ -0,0 +1,148 @@
...
+		} else {
+			// Use one thread per cpu core
+			flags = append(flags, "1C")
+		}
+	}
+
+	args := append(flags, "test-compile")
+
+	cmd := exec.Command("mvn", args...)
+	// Redirect the command's stdout to stderr to only have
```
fmeum, line 74: See above

internal/build/maven/maven.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+
+	bytes, err := os.ReadFile(outputPath)
+	if err != nil {
+		return nil, errors.WithStack(err)
+	}
+
+	deps := strings.Split(string(bytes), string(os.PathListSeparator))
+	// Append local dependencies which are not listed by "mvn dependency:build-classpath"
+	localDeps := []string{"classes", "test-classes", "resource", "test-resource"}
```
fmeum, line 142: We should add a comment stating that these directories are configurable and we should really read them from the resolved POM.

fmeum (approved):
Looks pretty good! The remaining comments are all just minor things, so I'm going to approve right away.

integration-tests/gradle/gradle_test.go
```diff
@@ -49,5 +66,108 @@ func TestIntegration_Gradle(t *testing.T) {
...
+	cifuzzRunner.Run(t, &shared.RunOptions{
+		ExpectedOutputs: []*regexp.Regexp{expectedOutputExp},
+	})
+
+	// Check that the findings command lists the finding
+	findings = shared.GetFindings(t, cifuzz, projectDir)
+	require.Len(t, findings, 1)
+	require.Contains(t, findings[0].Details, "Remote Code Execution")
+
+	//// TODO: This check currently fails on macOS because there
```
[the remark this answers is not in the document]
MarkusZoppelt: Fixed, thanks!

internal/build/maven/maven.go
MarkusZoppelt: Added a comment and opened a ticket for a different PR.

PR #349, opened 2022-10-04
ci(e2e): Fix `cifuzz bundle` failing on Windows and macOS

adombeck:
Since commit f5d5446f59bf2028d8363d6c9fdc12116df43011, the bundle command fails if the platform is not supported by our remote fuzzing servers (because the bundle is of no use in that case).

To still be able to test that the bundle can be created on those platforms, we set the CIFUZZ_BUNDLE_ON_UNSUPPORTED_PLATFORMS environment variable in our tests.


adombeck:
This should fix the failing compatibility check: https://github.com/CodeIntelligenceTesting/cifuzz/actions/runs/3182159093/jobs/5187756816

jochil (approved):

PR #352, opened 2022-10-05
integrate: Update gitignore integration for multiple build systems

florianGla:


internal/cmd/integrate/integrate_test.go
```diff
@@ -54,27 +55,83 @@ func TestMissingCIFuzzProject(t *testing.T) {
...
+	require.NoError(t, err)
+	defer fileutil.Cleanup(projectDir)
+
+	// Check that CMake specific entries are added to gitignore
+	err = fileutil.Touch(filepath.Join(projectDir, "CMakeLists.txt"))
+	require.NoError(t, err)
+	err = setupGitIgnore(projectDir)
+	assert.NoError(t, err)
+	content, err := os.ReadFile(filepath.Join(projectDir, ".gitignore"))
+	assert.NoError(t, err)
```
jochil, line 91: IMO some of these `assert.NoError`can be changed to `require.NoError`... if there is some unexpected error from a function as basic as `os.WriteFile` there is no need to continue the test

internal/cmd/integrate/integrate_test.go
```diff
@@ -54,27 +55,83 @@ func TestMissingCIFuzzProject(t *testing.T) {
...
 	// Check that only nonexisting entries are added
-	filesToIgnore := []string{
-		"/.cifuzz-build/",
-		"/.cifuzz-corpus/",
-		"test/.cifuzz-findings/",
-	}
-	err = os.WriteFile(gitIgnorePath, []byte(strings.Join(filesToIgnore, "\n")), 0644)
-	err = setupGitIgnore(testDir)
+	fileToIgnore := "/.cifuzz-corpus/"
+	err = os.WriteFile(gitIgnorePath, []byte(fileToIgnore), 0644)
```
jochil, line 72: this error is not checked

internal/cmd/integrate/integrate.go
```diff
@@ -105,11 +105,29 @@ func (c *integrateCmd) run() error {
...
+	buildSystem, err := config.DetermineBuildSystem(projectDir)
+	if err != nil {
+		return err
+	}
+	switch buildSystem {
+	case config.BuildSystemCMake:
+		filesToIgnore = append(filesToIgnore, []string{
+			"/.cifuzz-build/",
+			"/CMakeUserPresets.json",
+		}...)
```
adombeck, line 123: You don't have to create a slice to pass multiple arguments to `append`:
```suggestion
		filesToIgnore = append(filesToIgnore,
			"/.cifuzz-build/",
			"/CMakeUserPresets.json",
		)
```

internal/cmd/integrate/integrate.go
```diff
@@ -105,11 +105,29 @@ func (c *integrateCmd) run() error {
...
+	case config.BuildSystemMaven:
+		filesToIgnore = append(filesToIgnore, "/target/")
+	case config.BuildSystemGradle:
+		filesToIgnore = append(filesToIgnore, []string{
+			"/build/",
+			".gradle",
+		}...)
```
adombeck, line 130:
```suggestion
		filesToIgnore = append(filesToIgnore,
			"/build/",
			".gradle",
		)
```

internal/cmd/integrate/integrate.go
```diff
@@ -105,11 +105,29 @@ func (c *integrateCmd) run() error {
...
+		filesToIgnore = append(filesToIgnore, []string{
+			"/.cifuzz-build/",
+			"/CMakeUserPresets.json",
+		}...)
+	case config.BuildSystemMaven:
+		filesToIgnore = append(filesToIgnore, "/target/")
+	case config.BuildSystemGradle:
+		filesToIgnore = append(filesToIgnore, []string{
+			"/build/",
+			".gradle",
```
adombeck, line 129: What does `cifuzz` have to do with these directories, that we add those to `.gitignore`? To me it looks like those are directories used by Gradle by default - which means that if the user didn't change them in the Gradle config, they have to deal with ignoring them anyway, whether they use `cifuzz` or not. Is that correct?

internal/cmd/integrate/integrate.go
```diff
@@ -105,11 +105,29 @@ func (c *integrateCmd) run() error {
...
+		return err
+	}
+	switch buildSystem {
+	case config.BuildSystemCMake:
+		filesToIgnore = append(filesToIgnore, []string{
+			"/.cifuzz-build/",
+			"/CMakeUserPresets.json",
+		}...)
+	case config.BuildSystemMaven:
+		filesToIgnore = append(filesToIgnore, "/target/")
```
adombeck, line 125: Same question as with the Gradle ignored files: Is there anything special that `cifuzz` does with this directory that users don't have to deal with anyway when they use Maven?

internal/cmd/integrate/integrate.go
florianGla: You are right

adombeck (approved):

PR #354, opened 2022-10-06
chore: remove previous install from HOME when using make install

MarkusZoppelt:


jochil (approved):

adombeck (commented):
>  Merge branch 'main' into mz/remove-prev-install 

We try to avoid merge commits by rebasing our feature branches onto main instead

Makefile
```diff
@@ -52,6 +52,7 @@ deps/dev: deps
 
 .PHONY: install
 install:
+	-rm -rI ~/cifuzz
```
adombeck, line 55: Sorry for the long wait. I'm not completely happy with this change, because I have my cifuzz git repo checked out at `~/cifuzz` and install cifuzz to `~/cifuzz/build` instead, so I'm a bit afraid that this change might cause me to accidentally delete my cifuzz repo (with all the git-ignored files) at some point. I was thinking about alternatives, for example installing our files to `~/.local/{bin,lib,share}` instead, but didn't find enough time to think this through.

Now that it's already merged, I'm fine with keeping it this way and I'll try to be cautious not to use `make install` on my system.

MarkusZoppelt: I would also be happy with the `~/.local/{bin,lib,share}` locations.
This should be more conventional.

Shall we revert?

jochil: In the previous version `make install` already used `~/cifuzz` as the installation target, so we assumed that users having the repo at `~/cifuzz` would not use `make install`

jochil: > Shall we revert?

I would say it can stay... we usually avoid reverting stuff on `main` as it would affect all PRs and branches out there to some point.

In general I like the idea with `~/.local` too... maybe we can create a task/story in the backlog for it, as it would require adjusting the Installer/bundler/finder

jochil: @adombeck And I just remember... because of the `-I` flag on `rm` you have to confirm the deletion

PR #355, opened 2022-10-07
Don't recover from UBSan findings by default

adombeck:


tools/cmake/cifuzz/cmake/cifuzz-functions.cmake
```diff
@@ -85,14 +85,15 @@ function(enable_fuzz_testing)
...
-            -fsanitize=undefined
-            -fno-sanitize-recover=undefined
-        )
-        add_link_options(
-            -fsanitize=undefined
-            -fno-sanitize-recover=undefined
-        )
+        add_compile_options(-fsanitize=undefined)
+        add_link_options(-fsanitize=undefined)
+        if(CIFUZZ_RECOVER_UBSAN)
```
fmeum, line 90: Do we really need both a compile-time and a runtime switch? The compile-time switch is problematic as CMake variables are cached per build directory, meaning that the new bit would need to be "hashed" into the build directory name.

integration-tests/cmake/testdata/src/parser/parser.cpp
```diff
@@ -12,6 +12,10 @@ int parse(const std::string &input) {
   }
   if (input[0] == 'a' && input[1] == 'b' && input[2] == 'c') {
     if (input.find(SECRET_VALUE) != std::string::npos) {
+      // Trigger the undefined behavior sanitizer
+      int n = 23;
+      n <<= 32;
```
fmeum, line 17: This may not be very reliable as the compile can easily tell it's UB and optimize accordingly. I have seen `__attribute__((optnone))` in this context, we could add it here as well.

internal/build/other/other.go
```diff
@@ -174,6 +174,9 @@ func (b *Builder) setLibFuzzerEnv() error {
 		"-fsanitize=address,undefined",
 		// To support recovering from ASan findings
 		"-fsanitize-recover=address",
+		// Don't recover from UBSan findings, else no crashing input is
```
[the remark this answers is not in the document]
adombeck: Done in this commit, but same as this comment here it's removed again by the commit which introduces the `--recover-ubsan` option, which already explains the behavior in the flag's usage message.

internal/build/other/other.go
```diff
@@ -167,22 +168,33 @@ func (b *Builder) setLibFuzzerEnv() error {
...
-		"-fsanitize-address-use-after-scope",
 	}...)
+
+	if stringutil.Contains(b.Sanitizers, "address") {
+		cflags = append(cflags,
+			// Build with instrumentation for ASan and link its runtime
+			"-fsanitize=address",
+			// By default, we only fuzz until the first finding, so we
+			// don't recover from ASan findings
+			"-fno-sanitize-recover=address",
```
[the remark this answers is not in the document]
adombeck: Yes, good catch! But now that I'm thinking about it again, I wonder if this change could break current use cases of our users. IIRC, @pnoltof tried to set `ASAN_OPTIONS=halt_on_error=0` to recover from ASan findings - which is not something we officially support and I expect the UX to be pretty bad (with the same findings being reported over and over again), but maybe there is still some value in it? @pnoltof: Is this something you're currently using?

internal/cmdutils/pflag.go
```diff
@@ -141,6 +141,15 @@ func AddProjectDirFlag(cmd *cobra.Command) func() {
 	}
 }
 
+func AddRecoverUBSanFlag(cmd *cobra.Command) func() {
+	cmd.Flags().Bool("recover-ubsan", false,
+		"Recover from findings of the undefined behavior sanitizer and continue fuzzing.\n"+
```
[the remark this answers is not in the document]
adombeck: fixed

tools/cmake/cifuzz/cmake/cifuzz-functions.cmake
adombeck: I don't know if there is any difference in the behavior of ASan with (1) `-fsanitize-recover=address` and `ASAN_OPTIONS=halt_on_error=1` and (2) `-fno-sanitize-recover=address`. We could go with (1) by default and only set the runtime option, `ASAN_OPTIONS=halt_on_error=0` when the `--recover-ubsan` flag is used.

internal/build/other/other.go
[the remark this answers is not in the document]
fmeum: We could emit a warning if we see `halt_on_error=0` in options, but I would say that using such an option is pretty clearly beyond any "stable API contract" users could reasonably expect.

tools/cmake/cifuzz/cmake/cifuzz-functions.cmake
fmeum: Me neither. It's possible that `-fsanitize-recover=address` introduces runtime overhead, so we might want to check the LLVM sources for that. 

Not having conditional instrumentation flags would make this part of the code much simpler. In case we really do need different flags, I would rather introduce a new `asan-recoverable` sanitizer - this will play nicely with the cifuzz-CMake API contract we have settled on and lead to distinct build directories without further logic.

jochil (changes requested):
If we change the default behavior we also have to check the tests in the pipelines.
These pipelines explicitly expect two findings on mac and linux for the example projects:
```
.github/workflows/pipeline_min_requirements.yml
.github/workflows/pipeline_e2e_installer.yml
.github/workflows/pipeline_e2e_source.yml
```
And the compatibility check here `.github/actions/compatibility_check/action.yml` expect that a new release produces the same findings as the previous one

I think I would prefer using the `recover-ubsan` in the examples projects instead of adjusting the tests.

We should also update the READMEs of the example projects so the user is aware of this... as, with this change, the "interessting" heap buffer overflow in the last branch of the example source code is not covered anymore

tools/cmake/cifuzz/cmake/cifuzz-functions.cmake
adombeck: I spent some time looking through the LLVM sources, but I was not able to determine whether the `-fsanitize-recover` flag introduces significant runtime overhead. It does introduce checks in the instrumented code, for example [here](https://github.com/llvm/llvm-project/blob/df487bc425101c2550261db705f24f446967c6a7/llvm/lib/Transforms/Instrumentation/AddressSanitizer.cpp#L1632), but I can't tell if the runtime overhead of that is significant.

To get some rough idea, I ran our libfuzzer-runner tests a few times with `-fsanitize-recover=address` and a few times with `-fno-sanitize-recover=address`. The results are similar:
* recover: average CPU time: `13.27s` (min: `13.18`, max: `13.33`)
* no-recover: average CPU time: `13.39s` (min: `13.06`, max: `13.75`)

So I'm leaning towards always using `-fsanitize-recover=address` and only switching the runtime option. WDYT?

fmeum: I don't understand the details of the codegen, but there doesn't seem to be anything that should be unreasonably slow in this case. Backed up by the measurement, I would say go for it.

adombeck: done

integration-tests/cmake/testdata/src/parser/parser.cpp
adombeck: done

adombeck:
> I think I would prefer using the recover-ubsan in the examples projects instead of adjusting the tests.

done

adombeck:
> We should also update the READMEs of the example projects so the user is aware of this... as, with this change, the "interessting" heap buffer overflow in the last branch of the example source code is not covered anymore

done

internal/build/other/other.go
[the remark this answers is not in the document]
adombeck: The outcome of [this discussion](https://github.com/CodeIntelligenceTesting/cifuzz/pull/355#discussion_r989920133) was that we don't want to use the compile-time switch, so the behavior doesn't change - setting `ASAN_OPTIONS=halt_on_error=0` will work as before (i.e. still unsupported and untested but might still do something useful for some use cases). I agree that users can't have any reasonable expectations for options they set via `ASAN_OPTIONS` to work, so I don't think we need print a warning.

integration-tests/cmake/testdata/src/parser/parser.cpp
adombeck: that's causing errors on Windows: https://github.com/CodeIntelligenceTesting/cifuzz/actions/runs/3230350656/jobs/5288700407#step:5:167

I'll revert it

adombeck: Seems like MSVC does actually optimize this, because the undefined behavior is not reported. I'm trying to disable optimization for MSVC as well..

examples/other/cifuzz.yaml
```diff
@@ -1,16 +1,51 @@
...
+## Generated on 2022-10-11
+
+## The build system used to build this project. If not set, cifuzz tries
+## to detect the build system automatically.
+## Valid values: "cmake", "maven", "gradle", "other".
+#build-system: cmake
+
+## If the build system type is "other", this command is used by
+## `cifuzz run` to build the fuzz test.
+#build-command: "make my_fuzz_test"
```
jochil, line 11: my commenting out `build-command`? Shouldn't it be `build-command: make clean && make $FUZZ_TEST`?

integration-tests/cmake/testdata/src/parser/parser.cpp
adombeck: nevermind, forgot that MSVC doesn't support UBSan

examples/other/cifuzz.yaml
adombeck: thanks, fixed

examples/cmake/README.md
```diff
@@ -11,6 +11,11 @@ You can start the fuzzing with
 cifuzz run my_fuzz_test
 ```
 
+You should see two findings, an undefined behavior and a heap buffer
```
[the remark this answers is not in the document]
adombeck: Right. In that case, I'm not sure if it makes sense to mention the `recover-ubsan` in the README at all, it seems to me like we would have to explain too many things that are not really relevant for the user at the point they want to try out the example project.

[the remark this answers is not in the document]
jochil: I agree... it might be to specific/advanced for users having their first contact with fuzzing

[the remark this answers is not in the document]
adombeck: I removed the commit which updated the READMEs

jochil (approved):

PR #356, opened 2022-10-11
replayer: Fix UAF on internal failures

fmeum:
`current_path` has to be nulled right after the fuzz target has returned so that the exit handler doesn't try to print it if e.g. the next readdir call in the replayer fails - at this point, the path it points to has already been freed.


fmeum:
With https://github.com/CodeIntelligenceTesting/cifuzz/pull/358, this is not strictly necessary, but it may serve to prevent bugs in the future.

adombeck (approved):

PR #357, opened 2022-10-11
replayer: Fix failure due to fuzz target errno leak

fmeum:
The common idiom of calling `readdir` in a while loop fails if the body can end up setting `errno`. In our case, the fuzz target itself may end up setting `errno`, which we would then attribute to a `readdir` failure.


adombeck (approved):

adombeck:
This fix is also what the man page suggests (https://man7.org/linux/man-pages/man3/readdir.3.html#RETURN_VALUE):
> If the end of the directory stream is reached, NULL is returned
       and [errno](https://man7.org/linux/man-pages/man3/errno.3.html) is not changed.  If an error occurs, NULL is returned
       and [errno](https://man7.org/linux/man-pages/man3/errno.3.html) is set to indicate the error.  To distinguish end of
       stream from an error, set [errno](https://man7.org/linux/man-pages/man3/errno.3.html) to zero before calling readdir()
       and then check the value of [errno](https://man7.org/linux/man-pages/man3/errno.3.html) if NULL is returned.

fmeum:
> stream from an error, set [errno](https://man7.org/linux/man-pages/man3/errno.3.html) to zero before calling readdir()

Yep, now that I read it again, this is quite obvious. It seems impossible to find a correct snippet on SO though ;-)

PR #359, opened 2022-10-11
refactor: extract llvm specific coverage handling from coverage cmd

florianGla:
- Create own package for coverage generating
- Add test


internal/cmd/coverage/coverage.go
```diff
@@ -168,471 +151,71 @@ func (c *coverageCmd) run() error {
...
+			OutputFormat:   c.opts.OutputFormat,
+			OutputPath:     c.opts.OutputPath,
+			BuildSystem:    c.opts.BuildSystem,
+			BuildCommand:   c.opts.BuildCommand,
+			NumBuildJobs:   c.opts.NumBuildJobs,
+			SeedCorpusDirs: c.opts.SeedCorpusDirs,
+			FuzzTestArgs:   c.opts.FuzzTestArgs,
+			UseSandbox:     c.opts.UseSandbox,
+			FuzzTest:       c.opts.fuzzTest,
+			ProjectDir:     c.opts.ProjectDir,
```
adombeck, line 171: Instead of duplicating all these options, I would prefer to define them in an options struct in the `generator` package and re-use that in the options of the the coverage command, similar to how we do it with the [`bundler` options](https://github.com/CodeIntelligenceTesting/cifuzz/blob/refactor%2FCLI-549-coverage-cmd/internal/bundler/bundler.go#L73-L86) and the [bundle command options](https://github.com/CodeIntelligenceTesting/cifuzz/blob/refactor%2FCLI-549-coverage-cmd/internal/cmd/bundle/bundle.go#L20-L20).

internal/cmd/coverage/generator/generator.go
```diff
@@ -0,0 +1,5 @@
+package generator
+
+type CoverageGenerator interface {
+	Generate() (string, error)
+}
```
adombeck, line 5: What do we gain from defining this interface? IMO it's an unnecessary level of indirection, making it harder to read and understand the code. I would prefer this instead:
```golang
var reportPath string
switch c.opts.BuildSystem {
case config.BuildSystemCMake,
  config.BuildSystemOther:

  cov = &generator.LLVMCoverageGenerator{
    [...]
  }
  reportPath, err = cov.Generate()
  if err != nil {
    return err
  }
  
default:
  return errors.Errorf("Unsupported build system \"%s\"", c.opts.BuildSystem)
}
```

internal/cmd/coverage/generator/llvm_test.go
```diff
@@ -0,0 +1,107 @@
...
+			defer func() {
+				runfiles.Finder = oldFinder
+			}()
+
+			// mock finder to use include dir from repository
+			finder := &mocks.RunfilesFinderMock{}
+			finder.On("CIFuzzIncludePath").Return(includePath, nil)
+			finder.On("LLVMProfDataPath").Return("llvm-profdata", nil)
+			finder.On("LLVMCovPath").Return("llvm-cov", nil)
+			runfiles.Finder = finder
```
adombeck, line 80: Instead of setting the global `runfiles.Finder` here, would it work to add a `finder` field to the `LLVMCoverageGenerator` struct and let it use that? That could then default to `runfiles.Finder` and in the test we could set it to the mock finder instead.

internal/cmd/coverage/generator/generator.go
jochil: The interface is for the upcoming Gradle and maven integrations, which then can have their specific structs implementations

internal/cmd/coverage/coverage.go
jochil: We can think about this but most probably the structs for Gradle and Maven will have a smaller/different option set... Maybe we can wait until we have implemented these and can then think about a solution again

internal/cmd/coverage/generator/llvm_test.go
jochil: IMO we should not spread the usage of the `Finder` and instead think about a better solution than this global variable
Until then I think the Mock is a good solution

internal/cmd/coverage/generator/generator.go
jochil: To visualize it a little bit
```go
var cov CoverageGenerator

switch c.opts.BuildSystem {
case config.BuildSystemCMake,
  config.BuildSystemOther:
  cov = &generator.LLVMCoverageGenerator{[...]}
  
case config.BuildSystemMaven:
  cov = &generator.MavenCoverageGenerator{[...]}

case config.BuildSystemGradle:
  cov = &generator.GradleCoverageGenerator{[...]}

default:
  return errors.Errorf("Unsupported build system \"%s\"", c.opts.BuildSystem)
}

reportPath, err := cov.Generate()
if err != nil {
  return err
}
```

internal/cmd/coverage/generator/llvm_test.go
jochil: I thought about it again and I think this is a good idea, we are going to do it
But to keep the test in the `generator_test` package I guess we will make it a public field or adding a constructor

internal/cmd/coverage/generator/generator.go
adombeck: The same without an interface:
```
var reportPath string
switch c.opts.BuildSystem {
case config.BuildSystemCMake, config.BuildSystemOther:
  reportPath, err = &generator.LLVMCoverageGenerator{[...]}.Generate()
case config.BuildSystemMaven:
  reportPath, err = &generator.MavenCoverageGenerator{[...]}.Generate()
case config.BuildSystemGradle:
 reportPath, err = &generator.GradleCoverageGenerator{[...]}.Generate()
  reportPath, err = cov.Generate()
} 
default:
  return errors.Errorf("Unsupported build system \"%s\"", c.opts.BuildSystem)
}
if err != nil {
  return err
}
```

internal/cmd/coverage/generator/llvm_test.go
adombeck: Is there any advantage in using the `generator_test` package instead of `generator`? The latter would allow us to use a private field.

internal/cmd/coverage/coverage.go
adombeck: Ok, lets refactor when we have the implementations for Gradle and Maven

internal/cmd/coverage/generator/generator.go
jochil: of course it would work without an interface, no doubt... but my thoughts on that:
- the error handling is separated from the actual call. This reduces readability if the switch statement grows more eventually. And these 8 lines (or more) introduces the risk that when code gets updated the handling got overseen or something similar 
- This works only as long as we have a just a `Generate` method. If we have to add another method we are back to the interface (eg. checking Dependencies, dry-run, ...)
- using an interface here make it clear for the developer what we wanted to achieve with this separation: "No matter how the coverage generation internally works for the specific build system, just provide a `Generate` method and we are fine here".
- With the different generators (and coming more and more in the feature) this is like a book example for polymorphism and interfaces. I do not see that this add indirection as the actually program flow is still going straight to the specific struct.

> Note that this is also not the way interfaces should be used. Quoiting the [Go Code Review Comments](https://github.com/golang/go/wiki/CodeReviewComments#interfaces):
> 
> > Go interfaces generally belong in the package that uses values of the interface type, not the package that implements those values. The implementing package should return concrete (usually pointer or struct) types: that way, new methods can be added to implementations without requiring extensive refactoring.

What's your point here? This is about the location of the interface...of course we can move it to the `coverage` package, but IMO it still makes sense to use one

internal/cmd/coverage/generator/llvm_test.go
jochil: technically It would be no blackbox test anymore... and personally I prefer to do it like that when using an interface, but I think it would be ok to change it here. Would you prefer a constructor in this case?

internal/cmd/coverage/generator/generator.go
adombeck: > This works only as long as we have a just a Generate method. If we have to add another method we are back to the interface (eg. checking Dependencies, dry-run, ...)

That's premature optimization. If later we add more methods to the generators which the consumer calls in the same way, which means that an interface would actually make the code simpler, we could still add an interface.

> What's your point here? This is about the location of the interface...of course we can move it to the coverage package, but IMO it still makes sense to use one

That doesn't work with the way the interace is used here. In Go, structs implementing an interface should be passed to the consumer, which then uses the interface without knowing the actual type. Here it's the other way around, the coverage command does know which type the struct has (because it's instantiating it itself).

jochil: for the record: we agreed on removing the interface

internal/cmd/coverage/generator/llvm_test.go
jochil: I changed it, but as the generator uses the `other` Builder I had to change the `finder` usage there too... I created a new commit for this change

MarkusZoppelt: Can we mark this as resolved?
This seems to be the only unresolved review change for this PR to be merged.

adombeck: I pushed two refactor commits to pass the `runfiles.Finder` as a field in the options struct and use consistent naming.

adombeck (approved):

PR #360, opened 2022-10-11
cmake: Fix coverage build failure when C/CXX compiler isn't available

fmeum:
`DEFINED ${foo}` checks whether the contents of `foo` name a defined variable, which is not what we wanted.

Also remove some unused variables.


fmeum:
@adombeck This scenario is pretty difficult to cover with tests as we would need to mess with `PATH` to have it contain only a C++ but not a C compiler. I will rely on @wojtekCodeInt to manually verify that this actually fixes his issue.

fmeum:
That's a good base for a test, but we would also have to include `cmake` and all the other tools (e.g. the linker) it needs - except for a C compiler - onto the `PATH`. That would be rather fragile.

That said, I can't even reproduce the original failure without first failing the dependency check. This PR is still fixing a bug, but I don't know how @wojtekCodeInt reached it.

wojtekCodeInt:
> @adombeck This scenario is pretty difficult to cover with tests as we would need to mess with `PATH` to have it contain only a C++ but not a C compiler. I will rely on @wojtekCodeInt to manually verify that this actually fixes his issue.

It fixes the issue, thank you.

PR #363, opened 2022-10-12
Add completion for java fuzz tests

florianGla:


internal/completion/fuzz_test_completion.go
```diff
@@ -60,3 +67,42 @@ func validCMakeFuzzTests(projectDir string) ([]string, cobra.ShellCompDirective)
 	}
 	return res, cobra.ShellCompDirectiveNoFileComp
 }
+
+func validJavaFuzzTests(projectDir string) ([]string, cobra.ShellCompDirective) {
+	var res []string
+
+	testDir := filepath.Join(projectDir, "src", "test", "java")
+	err := filepath.WalkDir(testDir, func(path string, d fs.DirEntry, err error) error {
```
jochil, line 75: can we leave the function early when the file is not ending with `*java`?

fmeum, line 75: Instead of doing a full recursive walk of `src/test/java`, we could use `toComplete` to only walk and grep for the next segment.

florianGla: Good idea.

florianGla: How do you know if the next segment is valid, without checking the contents of all files?

fmeum: What we should definitely do is only walk directories that have `toComplete` as a prefix.

We *could* further optimize this by completing with any directory (corresponding to a Java package), but only those classes that contain the magic substrings. This would probably be vastly more efficient at the cost of letting the user complete their way into packages without fuzz tests.

Thinking more about it, doing the former is probably good enough for now.

internal/completion/fuzz_test_completion.go
```diff
@@ -60,3 +67,55 @@ func validCMakeFuzzTests(projectDir string) ([]string, cobra.ShellCompDirective)
...
+
+	testDir := filepath.Join(projectDir, "src", "test", "java")
+	completionPrefix := filepath.Join(
+		testDir,
+		strings.ReplaceAll(toComplete, ".", string(os.PathSeparator)),
+	)
+
+	err := filepath.WalkDir(testDir, func(path string, d fs.DirEntry, err error) error {
+		if !strings.HasPrefix(path, completionPrefix) {
+			return nil
```
fmeum, line 82: If `!strings.HasPrefix(path, filepath.Dirname(completionPrefix))`, we could return `filepath.SkipDir` as an optimization.

fmeum (approved):
Looks good! Just left one suggestion.

PR #364, opened 2022-10-12
Fix installation of command completion scripts

adombeck:


MarkusZoppelt (approved):

MarkusZoppelt:
> installer: Fix installation of bash completion script on macOS


This PR works for zsh completions but bash completions don't seem to work for me (macOS)

adombeck:
> This PR works for zsh completions but bash completions don't seem to work for me on macOS

Thanks for reporting! Maybe this is something you and/or @mlsandnerCI could investigate further, because you have access to a Mac.

cmd/installer/installer.go
```diff
@@ -205,37 +244,92 @@ func installBashCompletionScript(cifuzzPath string) error {
...
 	case "darwin":
-		cmd = exec.Command("bash", "-c", "'"+cifuzzPath+"' completion bash > \"$(brew --prefix)/etc/bash_completion.d/cifuzz\"")
+		// There are no bash completoin directories on macOS by default,
+		// so we need user action to source our installation directory
+		notes = append(notes, fmt.Sprintf(`To enable command completion:
+
+    echo source '%s' >> ~/.bash_profile
+
+`, completionScriptPath))
 	}
```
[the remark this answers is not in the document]
mlsandnerCI: Checked this and it works for me too 👍

cmd/installer/installer.go
```diff
@@ -214,28 +232,74 @@ func installBashCompletionScript(cifuzzPath string) error {
...
+		notes = append(notes, fmt.Sprintf(`To enable command completion:
+
+    echo fpath=(%s $fpath) >> ~/.zshrc
+    echo "autoload -U compinit; compinit" >> ~/.zshrc
+
+`, completionsDir))
+	} else {
+		notes = append(notes, `To enable command completion (if not already enabled):
+
+    echo "autoload -U compinit; compinit" >> ~/.zshrc
```
jochil, line 329: Is there a way to check if completions are already available? It would be nice to print this only when needed

jochil: I guess something like `autoload | grep -E "^compinit "` could work

adombeck: > I guess something like autoload | grep -E "^compinit " could work

I think that only tells us if `autoload compinit` was executed, not if the `compinit` function itself was executed.

> Is there a way to check if completions are already available? 

I don't know a reliable way to check if completions are available.

cmd/installer/installer.go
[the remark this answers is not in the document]
adombeck: > (This implies that users need to have bash-completion installed via homebrew.)

So we should tell them that they have to install bash-completion if it's not installed, right?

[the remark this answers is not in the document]
MarkusZoppelt: > So we should tell them that they have to install bash-completion if it's not installed, right?

Yes. Though, this is needed for any bash-completion on macOS. So we can assume that "seasoned" bash users already have this installed and enabled, imho.

The default login shell on macOS is zsh anyway...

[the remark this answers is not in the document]
adombeck: > The default login shell on macOS is zsh anyway...

Ah, I didn't know that

cmd/installer/installer.go
```diff
@@ -205,37 +244,94 @@ func installBashCompletionScript(cifuzzPath string) error {
...
+			return errors.WithStack(err)
+		}
 	case "darwin":
-		cmd = exec.Command("bash", "-c", "'"+cifuzzPath+"' completion bash > \"$(brew --prefix)/etc/bash_completion.d/cifuzz\"")
+		// There are no bash completoin directories on macOS by default,
+		// so we need user action to source our installation directory
+		notes = append(notes, fmt.Sprintf(`To enable command completion:
+
+    # enable bash completion (if not already enabled):
+    echo "if [ -f $(brew --prefix)/etc/bash_completion ]; then . $(brew --prefix)/etc/bash_completion; fi" >> ~/.bash_profile
```
adombeck, line 258: Suggestion for minor improvements:
* Use the more descriptive `source` instead of `.`
* Use the shorter `&&` instead of `if`, `then` (since this is a one-liner - otherwise I also prefer `if`)
```suggestion
    echo "[ -f $(brew --prefix)/etc/bash_completion ] && source $(brew --prefix)/etc/bash_completion" >> ~/.bash_profile
```

MarkusZoppelt: Great suggestion, will change. Thanks!

cmd/installer/installer.go
jochil: ok, I see

PR #365, opened 2022-10-13
integrate: Add interactive multiselect dialog

florianGla:


internal/cmd/integrate/integrate.go
```diff
@@ -104,6 +114,20 @@ func (c *integrateCmd) run() error {
...
+	printer.KeyConfirm = keys.Enter
+	printer.KeySelect = keys.Space
+	selectedOptions, err := printer.Show()
+	if err != nil {
+		return nil, errors.WithStack(err)
+	}
+
+	return selectedOptions, nil
+}
+
```
jochil, line 130: What do you think about moving it to the `dialog` package... there is already another `Select` method we use in interactive commands

internal/cmd/integrate/integrate.go
```diff
@@ -104,6 +116,16 @@ func (c *integrateCmd) run() error {
 	return nil
 }
 
+// selectTools lets the user select the desired tools via an interactive multiselect dialog
+func selectTools() ([]string, error) {
+	selectedTools, err := dialog.MultiSelect("Select tools for integration", supportedTools)
+	if err != nil {
+		fmt.Printf("%+v \n", err)
```
adombeck, line 123: The error shouldn't be printed to stdout. If we want to print it here, it should be to stderr - but since it's an unexpected error, IMO we should just return it with a stack trace.

jochil: It is the same way we handle the error in the `create` command (see `func (c *createCmd) getTestType() ` L125)
I would agree to change it... but then also in this location, too.

or maybe still returning it as an silent error and using `log.Debug`?

jochil (approved):

PR #367, opened 2022-10-17
all: Remove the --fuzz-test-args flag

adombeck:
It was broken and no one noticed, so there doesn't seem to be any demand for this flag.


integration-tests/shared/bundle.go
```diff
@@ -117,9 +115,8 @@ func TestBundle(t *testing.T, dir string, cifuzz string, args ...string) {
 	assert.Equal(t, "my-branch", metadata.CodeRevision.Git.Branch)
 	assert.Equal(t, "123456abcdef", metadata.CodeRevision.Git.Commit)
 
-	// Verify that the metadata contain the engine args and fuzz test args
+	// Verify that the metadata contain the engine options
```
fmeum, line 118:
```suggestion
	// Verify that the metadata contain the engine flags
```

fmeum (approved):

integration-tests/shared/bundle.go
adombeck: applied

PR #368, opened 2022-10-17
pkg/artifact: Rename the YAML file in the bundle

adombeck:
It used to be named cifuzz.yaml, same as the project config file. To avoid confusion, it's now named bundle.yaml.


integration-tests/shared/bundle.go
```diff
@@ -102,7 +102,7 @@ func TestBundle(t *testing.T, dir string, cifuzz string, args ...string) {
 	require.NoError(t, err)
 
 	// Read the fuzzer path from the YAML.
-	metadataPath := filepath.Join(archiveDir, "cifuzz.yaml")
+	metadataPath := filepath.Join(archiveDir, artifact.MetadataFileName)
```
fmeum, line 105: I would slightly prefer inlining the constant. It is part of our API contract, so some test should break if we change it.

fmeum (approved):

integration-tests/shared/bundle.go
adombeck: done

PR #374, opened 2022-10-24
feat: change install directory

mlsandnerCI:
- Installation directory is now either `$XDG_DATA_HOME/cifuzz` if set, `/opt/code-intelligence/cifuzz` if run as root or as a default `~/.local/share/cifuzz`. 
- The `bin` directory is now always `~/.local/bin`
- `-i` flag for installation was removed 
- Tests were adjusted to new executable paths
- Makefile was adjusted to accommodate the new installation directories when running `make install`


Makefile
```diff
@@ -56,7 +56,9 @@ deps/dev: deps
 
 .PHONY: install
 install:
-	-rm -rI ~/cifuzz
+	-rm -rI $(XDG_DATA_HOME)/cifuzz
+	-rm -rI ~/.local/share/cifuzz
+	-rm -rI opt/code-intelligence/cifuzz
```
MarkusZoppelt, line 61: pretty sure this is missing a leading slash:
    
    -rm -rI /opt/code-intelligence/cifuzz

internal/installer/installer.go
```diff
@@ -1,5 +1,40 @@
...
+func GetInstallDir() (string, error) {
+	var installDir string
+
+	xdgPath, xdgSet := os.LookupEnv("XDG_DATA_HOME")
+
+	switch {
+	case xdgSet:
+		installDir = filepath.Join(xdgPath, "cifuzz")
+	case os.Getuid() == 0:
+		installDir = filepath.Join("opt", "code-intelligence", "cifuzz")
```
MarkusZoppelt, line 25: I would define the /opt directory as an absolute path, including the forward slash

Makefile
mlsandnerCI: Thanks, didn't see that 👍

internal/installer/installer.go
mlsandnerCI: Done 👍

mlsandnerCI: Wait, shouldn't this be resolved because I call filepath.Abs() in the end?

MarkusZoppelt: if that ultimately returns `/opt/code-intelligence/cifuzz`, then yes.

mlsandnerCI: As I understand the [function](https://www.geeksforgeeks.org/filepath-abs-function-in-golang-with-examples/), yes it should 👍

MarkusZoppelt: As I read it, it prefixes installdir with the current working directory and returns the absolute path to that location, e.g.:

```go
// this returns "opt/code-intelligence/cifuzz":
installDir = filepath.Join("opt", "code-intelligence", "cifuzz")

// if $PWD is ~/foo/bar, then
installDir = filepath.Abs(installDir)
// returns "/home/$user/foo/bar/opt/code-intelligence/cifuzz"
```

mlsandnerCI: Oh okay

mlsandnerCI: I added a path separator before the `filepath.Join()`

MarkusZoppelt: Looks good! I am curious how this will behave on Windows tho...
Is there maybe another "official" directory equivalent for /opt?

adombeck:
> * Installation directory is now either `$XDG_DATA_HOME/cifuzz` if set, `/opt/code-intelligence/cifuzz` if run as root or as a default `~/.local/share/cifuzz`.
> 
>     * The `bin` directory is now always `~/.local/bin`
> 
>     * `-i` flag for installation was removed
> 
>     * Tests were adjusted to new executable paths
> 
>     * Makefile was adjusted to accommodate the new installation directories when running `make install`

Sounds good! Could you add this summary to the commit message as well (preferably in present tense)?

Makefile
```diff
@@ -56,7 +56,10 @@ deps/dev: deps
 
 .PHONY: install
 install:
-	-rm -rI ~/cifuzz
+	-rm -rI $(XDG_DATA_HOME)/cifuzz
+	-rm -rI ~/.local/share/cifuzz
+	-rm -rI /opt/code-intelligence/cifuzz
+	-rm -i ~/.local/bin/cifuzz
```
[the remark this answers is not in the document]
adombeck: >  I think it's fine to just remove any previous installation without asking for confirmation

In the exact locations we are about to install to

internal/installer/installer.go
```diff
@@ -1,5 +1,56 @@
...
+func GetInstallDir() (string, error) {
+	var installDir string
+
+	xdgPath, xdgSet := os.LookupEnv("XDG_DATA_HOME")
+
+	switch {
+	case xdgSet:
+		installDir = filepath.Join(xdgPath, "cifuzz")
+	case os.Getuid() == 0:
+		installDir = string(os.PathSeparator) + filepath.Join("opt", "code-intelligence", "cifuzz")
```
adombeck, line 26: Lets switch these two cases: If we're run as root, we should always do a system-wide installation and therefore not use `$XDG_DATA_HOME`, which is for user-specific files.

```suggestion
	case os.Getuid() == 0:
		installDir = string(os.PathSeparator) + filepath.Join("opt", "code-intelligence", "cifuzz")
	case xdgSet:
		installDir = filepath.Join(xdgPath, "cifuzz")
```

internal/installer/installer.go
```diff
@@ -1,5 +1,56 @@
...
+)
+
 var Deps = []string{
 	"cmd", "internal", "pkg", "util", "third-party/minijail",
 }
+
+// GetInstallDir returns the absolute path to the installation
+// directory which is chosen according if $XDG_DATA_HOME is set
+// or the user is running the command as root.
+func GetInstallDir() (string, error) {
```
adombeck, line 17: This implementation is now doing the right thing for Unix, but on Windows the new installation directory is worse than the old one: On Windows, `os.UserHomeDir()` returns `%USERPROFILE%` (i.e. `c:\Users\username`), so the old installation directory was `c:\Users\username\cifuzz`. The new one is `c:\Users\username\.local\share\cifuzz`, which is very non-standard on Windows. We should look up what the most appropriate installation directory is on Windows (I suspect something like `C:\Program Files\cifuzz`) and use that. Doesn't have to be part of this PR, in which case we should create a follow-up task and add a check here to return the old installation directory (`os.UserHomeDir()\cifuzz`) on Windows.

Makefile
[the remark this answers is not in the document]
mlsandnerCI: Okay, to check if I understood it correctly: The installer should remove the directories/executable on it's own instead of returning an error if they already exist?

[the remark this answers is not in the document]
adombeck: yes, any directories / files which are in the location the installer is about to install to

internal/installer/installer.go
mlsandnerCI: I think it makes sense to do it in this PR. I will research the topic but I expect `C:\Program Files\cifuzz` will probably be the way to go because of what I have read before when I looked up whether `/opt` is applicable for Windows.

adombeck:
> > @mlsandnerCI It's kind of odd how deep the share directories point now, e.g., ~/.local/share/cifuzz/share/cifuzz/share
> > 
> > Can't we put everything from ~/.local/share/cifuzz/share/cifuzz/* to ~/.local/share/cifuzz/.?
> > 
> > I don't see why we need
> > 
> > ~/.local/share/cifuzz/share/cifuzz/share
> >                       ~~~~~~~~~~~~
> >                          this  
> > 
> Good point. @adombeck What do you think?

Absolutely, our installation directory contains too many nested `cifuzz/share` directories. But that's already the case currently (`~/cifuzz/share/cifuzz/share/*`), this PR only makes it slightly worse by adding another `share` to the path, so I suggest we create a follow-up task to improve that. I thought I had already created one for this issue a few weeks ago but I can't find it, so lets create a new one.

cmd/installer/installer.go
```diff
@@ -114,7 +130,13 @@ func ExtractEmbeddedFiles(targetDir string, files *embed.FS) error {
 		}
 
 		if !d.IsDir() {
-			targetDir := filepath.Dir(filepath.Join(targetDir, path))
+			var targetDir string
+			if path == "bin/cifuzz" {
+				targetDir = binDir
+			} else {
+				targetDir = filepath.Dir(filepath.Join(installDir, path))
+			}
```
adombeck, line 138: We don't want to use a special directory for binaries on Windows, so we should add a check here. It's funny that this code probably does the correct thing, because you're using a `/` as the path separator here instead of using `filepath.Join` and on Windows the path will never match `bin/cifuzz` - but I would still prefer an explicit platform check :)

cmd/installer/installer.go
```diff
@@ -150,12 +172,12 @@ func ExtractEmbeddedFiles(targetDir string, files *embed.FS) error {
 
 	// Install the autocompletion script for the current shell (if the
 	// shell is supported)
-	cifuzzPath := filepath.Join(targetDir, "bin", "cifuzz")
+	cifuzzPath := filepath.Join(binDir, "cifuzz")
```
adombeck, line 175: This will also break on Windows - although I'm not sure whether `SHELL` is even set on Windows. We should check if this code is relevant for Windows at all.

internal/builder/builder.go
```diff
@@ -436,12 +441,17 @@ func FindProjectDir() (string, error) {
...
 
-func CIFuzzExecutablePath(binDir string) string {
-	path := filepath.Join(binDir, "cifuzz")
+func CIFuzzExecutablePath() (string, error) {
+	path, err := installer.GetBinDir()
+	if err != nil {
+		return "", err
+	}
+
+	path = filepath.Join(path, "cifuzz")
```
adombeck, line 450: The builder shouldn't create the binary in the installation directory for binaries, it's the job of the installer to put it there.

To clear up confusions, this is our procedure for building and installing:
1. Run the builder, which builds cifuzz and its dependencies and puts the binaries in a build directory
2. Build the installer, which embeds the binaries from the build directory
3. Run the installer, which installs the binaries to the installation directory (directories now)

mlsandnerCI: I definitely had problems understanding the difference between the building process and the installation. Thanks for clearing it up, I will fix this.

internal/installer/installer.go
adombeck: > I think the correct directory is %AppData% which points to C:\Users\USERNAME\AppData\Roaming on Windows.

Any authoritative source for that? Then we can use that and refer to the source in a comment.

cmd/installer/installer.go
mlsandnerCI: Could you explain why it will break on Windows?

adombeck: Sorry, should have been clearer: It will break when we don't put cifuzz in the `binDir` anymore on Windows, as I suggested [here](https://github.com/CodeIntelligenceTesting/cifuzz/pull/374#discussion_r1004531444)

internal/installer/installer.go
MarkusZoppelt: Not much online. Maybe this? (german only)
https://www.heise.de/select/ct/2018/11/1526770940977508#:~:text=Appdata%3A%20Local%2C%20LocalLow,ct.de

I just noticed that NeoVim put some config files and plugin data there when I was on Windows some years ago...

cmd/installer/installer.go
mlsandnerCI: Got it, thank you 👍

internal/installer/installer.go
mlsandnerCI: I did some research and could not find any official document, only some stackoverflow threads talking about either `C:\Program Files` or `%AppData%`.

[Microsoft defines](https://learn.microsoft.com/en-us/dotnet/api/system.environment.specialfolder?view=net-6.0&viewFallbackFrom=windowsdesktop-7.0) `ProgramFiles` as:

> The program files directory.

(duh, thanks for that)

And `AppData` as:

> The directory that serves as a common repository for application-specific data for the current roaming user. A roaming user works on more than one computer on a network. A roaming user's profile is kept on a server on the network and is loaded onto a system when the user logs on.

I also found [an interesting (but not official) thread](https://learn.microsoft.com/en-us/answers/questions/584279/what39s-the-idea-behind-the-name-of-the-appdata-fo.html) where `ProgramFiles` is explained as:

> To store your application's read-only files, create a subfolder within the Program Files folder.
> Include your application's executable (.exe) files as well as any .dll files, Help files, clip art or
> other fixed library files, and other support files used by your application. 

And `AppData`:

> Create other miscellaneous files that the user doesn't directly access in the Application Data
> folder. For example, include custom dictionaries, mail stores, user history and other caches,
> index files, and log files. If you provide multiple applications, within the Application Data folder
> you can create a folder with your company's name that contains subfolders for each product.

And I like the answer about advantages and disadvantages:

> One of the benefits of installing to Program Files is that is exactly where the user expects it to go. Usually users go here first to look where the program is installed. The biggest downside is that the user needs administrative privileges, which isn't always required by the program.
> 
> I'd advise against installing to AppData unless you have really good reason to do so. Program Files is where the user will expect it, and the permissions are set up like that as well. AppData has a few flaws, namely that the program is only installed for that one user, and could actually be repeatedly installed because of that. The other thing is AppData roaming follows users on a server network, and so making that take more space could make logging in slower.

So yeah, both would work. Any opinions? I think I personally would go with `ProgramFiles`.

cmd/installer/installer.go
mlsandnerCI: Done.

internal/installer/installer.go
adombeck: Thanks for the summary!

> The biggest downside is that the user needs administrative privileges, which isn't always required by the program.

That is indeed a downside. Since cifuzz itself doesn't need administrator privileges, and we also support installing it without root privileges on Unix, I'm leaning towards installing to `AppData`. I guess most similar to the behavior on Unix would be to install to `ProgramFiles` when the user does have administrative privileges and else install to `AppData`. But we shouldn't complicate things for our limited Windows support (and on Windows we don't have the issue that there is no user-specific installation directory for the root user), so IMO it would be fine to always install to `AppData`.

mlsandnerCI: Alright 👍

Makefile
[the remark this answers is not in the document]
mlsandnerCI: Done.

internal/installer/installer.go
mlsandnerCI: Done.

internal/builder/builder.go
mlsandnerCI: Done.

internal/installer/installer.go
```diff
@@ -1,5 +1,75 @@
...
+			return "", err
+		}
+		binDir = filepath.Join(installDir, "bin")
+	} else {
+		home, err := os.UserHomeDir()
+		if err != nil {
+			return "", errors.WithStack(err)
+		}
+		binDir = filepath.Join(home, ".local", "bin")
+	}
```
adombeck, line 67: I see that I didn't put that in the ticket description: The executable should only be installed to `~/.local/bin` if the installer is run as non-root, because `/root/.local/bin` is not in the `PATH` of any users and not even readable by any user but root. I think we should install the executable to `/usr/local/bin` instead. I created a follow-up ticket for that: https://code-intelligence.atlassian.net/browse/CLI-612

integration-tests/shared/helpers.go
```diff
@@ -135,20 +136,30 @@ func InstallCIFuzzInTemp(t *testing.T) string {
...
-		// Install cifuzz
-		tempDir, err := os.MkdirTemp("", "cifuzz-")
+		// Create directory for installation files
+		installDir, err := os.MkdirTemp("", "cifuzz-")
+		require.NoError(t, err)
+		err = os.Setenv("XDG_DATA_HOME", installDir)
+		require.NoError(t, err)
+		err = os.Setenv("HOME", installDir)
+		require.NoError(t, err)
+		err = os.Setenv("APPDATA", installDir)
```
adombeck, line 146: We should avoid setting environment variables in tests, because those are global state and can therefore cause race conditions when multiple tests are run in parallel.

Lets do this instead:
```golang
		env := os.Environ()
		env, err = envutil.Setenv(env, "XDG_DATA_HOME", installDir)
		require.NoError(t, err)
		env, err = envutil.Setenv(env, "HOME", installDir)
		require.NoError(t, err)
		env, err = envutil.Setenv(env, "APPDATA", installDir)
		require.NoError(t, err)
[...]
		installCmd.Env = env
```

This will have the side-effect that `installer.GetBinDir()` doesn't return the correct binary directory anymore. I think we can avoid the call to `installer.GetBinDir()` here and just return `filepath.Join(installDir, "bin", "cifuzz")`, because that's the path where `cifuzz` will be installed to by this function in all cases.

internal/installer/installer.go
adombeck: If you want to, you can do it as part of this PR - I just didn't want to block on this.

mlsandnerCI: Okay 👍

integration-tests/shared/helpers.go
mlsandnerCI: I implemented this but now there is the problem that the generation of a coverage report fails because the builder uses the `RunfilesFinder`which uses the `installer.GetInstallDir()` function which then returns the wrong path to the executable because the environment variables are not set in that scope.

mlsandnerCI: Fixed it by just doing the same and setting the environment variables specific for the coverage and run command

adombeck: That's a good solution. Since we always have to set the test environment when `CIFuzzRunner.Run()` is called, we can do it inside that function like this:

```golang
cmd.Env = append(opts.Env, CreateTestEnvironment(t, installDir)...)
```

Some other minor points:
* Maybe rename `CreateTestEnvironment` to something shorter like `TestEnv` (that's just a matter of taste)
* Since `installDir` is a global variable in package `shared`, and `installOnce` ensures that it is only set once, we don't even have to pass it to `TestEnv`
* If we use `cmd.Env = append(opts.Env, TestEnv(t)...)` like I suggested above, `TestEnv` doesn't have to include `os.Environ`, because `opts.Env` defaults to `os.Environ`. This allows us to still execute commands in an environment which does not include `os.Environ` (by setting `opts.Env`).

adombeck:
> Maybe someone with a Linux OS could look over it tomorrow real quick with me to check what is actually needed to get the tests to pass?

Yes, we can do that together

integration-tests/shared/helpers.go
mlsandnerCI: Thank you for the feedback! I will implement these changes tomorrow morning 👍

cmd/installer/installer.go
```diff
@@ -363,36 +386,20 @@ func installFishCompletionScript(cifuzzPath string) error {
...
-
-	if strings.HasPrefix(installDir, "~") {
-		home, err := os.UserHomeDir()
-		if err != nil {
-			return "", errors.WithStack(err)
-		}
-		installDir = home + strings.TrimPrefix(installDir, "~")
+// checkAndRemoveDirectory checks if the given directory already exists
+// and removes it in that case.
+func checkAndRemoveDirectory(dir string) error {
```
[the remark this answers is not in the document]
mlsandnerCI: Done.

cmd/installer/installer.go
```diff
@@ -114,7 +130,14 @@ func ExtractEmbeddedFiles(targetDir string, files *embed.FS) error {
 		}
 
 		if !d.IsDir() {
-			targetDir := filepath.Dir(filepath.Join(targetDir, path))
+			var targetDir string
+			// cifuzz executable should be in extra bin directory
+			if path == executableFiles[0] {
```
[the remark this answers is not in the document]
mlsandnerCI: Done.

internal/installer/installer.go
```diff
@@ -1,5 +1,75 @@
...
+
+	switch {
+	case runtime.GOOS == "windows":
+		appdata, err := os.UserConfigDir()
+		if err != nil {
+			return "", errors.WithStack(err)
+		}
+		installDir = filepath.Join(appdata, "cifuzz")
+	case os.Getuid() == 0:
+		installDir = string(os.PathSeparator) + filepath.Join("opt", "code-intelligence", "cifuzz")
```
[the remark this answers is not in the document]
mlsandnerCI: Done.

integration-tests/shared/helpers.go
mlsandnerCI: I implemented the changes 👍

adombeck:
> @adombeck Just checking in, are these outputs okay?

Looks good to me

README.md
```diff
@@ -161,29 +175,71 @@ choco install microsoft-openjdk maven
...
+If you installed cifuzz as **root**:
+
+```bash
+sudo rm -rf /opt/code-intelligence/cifuzz /usr/local/bin/cifuzz /usr/local/share/cifuzz
+```
+
+If you installed cifuzz as **non-root**:
+
+```bash
+rm -rf "${XDG_HOME_DATA:-$HOME/.local/share}/cifuzz" ~/.local/bin/cifuzz ~/.cmake/packages/cifuzz
```
adombeck, line 233:
```suggestion
rm -rf "${XDG_DATA_HOME:-$HOME/.local/share}/cifuzz" ~/.local/bin/cifuzz ~/.cmake/packages/cifuzz
```

README.md
```diff
@@ -161,29 +175,71 @@ choco install microsoft-openjdk maven
...
+</details>
+
+## Uninstall
+
+<details>
+<summary>Version <= 0.5.0</summary>
+If you installed cifuzz into the default directory:
+
+```bash
+sudo rm -rf ~/cifuzz
```
adombeck, line 213: I see two issues here:
* When executed via `sudo`, `~/cifuzz` will resolve to `/root/cifuzz`, which is only the installation directory when the installer was also executed as root. So we also need to handle the case that cifuzz was installed as root differently here.
* It's missing the files installed for the CMake integration, `/usr/local/share/cifuzz` in the case cifuzz was installed as root, otherwise `~/.cmake/packages/cifuzz`

README.md
```diff
@@ -161,29 +175,71 @@ choco install microsoft-openjdk maven
...
+```
+
+If you installed into a custom set installation directory you will have to remove that one instead.
```
adombeck, line 216:
```suggestion
If you installed into a custom installation directory you have to remove that one instead.
```

README.md
```diff
@@ -161,29 +175,71 @@ choco install microsoft-openjdk maven
...
+If you installed cifuzz as **non-root**:
+
+```bash
+rm -rf "${XDG_DATA_HOME:-$HOME/.local/share}/cifuzz" ~/.local/bin/cifuzz ~/.cmake/packages/cifuzz
+```
+
+If you installed cifuzz on **Windows**:
+
+```bash
+rd /s %APPDATA%/cifuzz
```
adombeck, line 239: To be extra clean, we might want to remove the registry entry as well, see https://github.com/CodeIntelligenceTesting/cifuzz/blob/530c81f6d066dc35c53e05cfe2b36c3d28587967/internal/installer/installer_windows.go#L24-L24.

Here is how registry entries can be deleted from the command-line: https://superuser.com/a/404856/551429

mlsandnerCI: How do I know if I need `HKCU` or `HKLM` as prefix?

adombeck: I don't have a Windows machine to test this, but maybe we can just try both and ignore errors

mlsandnerCI: Did that and added `> nul` to ignore errors (found it [here](https://stackoverflow.com/questions/20298682/suppress-error-messages-in-windows-commandline))

cmd/installer/installer.go
```diff
@@ -46,13 +46,26 @@ func main() {
 		os.Exit(0)
 	}
 
-	err = ExtractEmbeddedFiles(*installDir, &buildFiles)
+	if installDir, exists := oldInstallationExists(); exists && !*ignoreCheck {
+		log.Warnf(
+			`Old cifuzz installation exists in %s. 
+To prevent version issues please remove the files. 
+See https://github.com/CodeIntelligenceTesting/cifuzz#uninstall `,
```
adombeck, line 53:
```suggestion
			`Old cifuzz installation exists in %s.
To prevent version issues please remove the files.
See https://github.com/CodeIntelligenceTesting/cifuzz#uninstall`,
```

README.md
mlsandnerCI: Done.

adombeck (approved):

PR #375, opened 2022-10-25
trigger coverage run in 'cifuzz coverage' for maven/gradle

MarkusZoppelt:
Open issues:

- We create both html and xml output. The xml output has to be created anyway, because it is needed for the summary table. That means that the output format flag is not used. Shall we change this behavior?
- The default output directory is "build/reports/cifuzz" for gradle and ".cifuzz-build/report" for maven. I am not sure why we decided to do this. Shall we change that to a common output directory or is that the expected behavior for maven/gradle projects respectively?


examples/gradle/src/test/java/com/example/FuzzTestCase.java
```diff
@@ -10,6 +10,7 @@ void myFuzzTest(FuzzedDataProvider data) {
         int b = data.consumeInt();
         String c = data.consumeRemainingAsString();
 
-        ExploreMe.exploreMe(a, b, c);
+        ExploreMe ex = new ExploreMe();
```
fmeum, line 13: Passing one of the parameters into the constructor and the other two into the method call would make this example more realistic.

examples/maven/src/main/java/com/example/ExploreMe.java
```diff
@@ -2,7 +2,7 @@
 
 public class ExploreMe {
     // Function with multiple paths that can be discovered by a fuzzer.
-    public static void exploreMe(int a, int b, String c) {
+    public void exploreMe(int a, int b, String c) {
```
jochil, line 5: What's the reason for this change? I would expect it working also with static methods?

share/init.gradle
```diff
@@ -9,4 +9,48 @@ allprojects {
...
+        // `generate()`
+        executionData execFile
+        classDirectories.from files(project.sourceSets.main.output)
+        sourceDirectories.from files(project.sourceSets.main.java.srcDirs)
+
+        def output = project.findProperty("cifuzz.report.output") ?: layout.buildDirectory.dir("reports/cifuzz").get().toString()
+        reports {
+          xml.required = true
+          xml.outputLocation = file("$output/jacoco.xml")
+          html.outputLocation = new File(output)
```
fmeum, line 51: Do we emit the XML right next to the HTML reports index.html this way?
O would prefer having the HTML in a separate subdirectory instead.

examples/maven/src/main/java/com/example/ExploreMe.java
MarkusZoppelt: It does!
But @fmeum suggested to change the examples to be more Java idiomatic.
It also shows more interesting results in coverage reports.

share/init.gradle
```diff
@@ -9,4 +9,48 @@ allprojects {
...
+
+    // we need to set the exec file output path explicitly to make sure we find
+    // the file in the `cifuzzReport` tasks
+    def execFile = layout.buildDirectory.file("jacaco/cifuzz.exec").get().asFile
+    def fuzzTest = project.findProperty("cifuzz.fuzztest") ?: "*"
+
+    tasks.register("cifuzzTest", Test) {
+        useJUnitPlatform()
+        ignoreFailures = true
+        jvmArgs '-Djazzer.hooks=false'
```
jochil, line 28: Maybe we can shortly explain the reason why this is necessary in a comment?

pkg/coverage/jacoco.go
```diff
@@ -0,0 +1,113 @@
...
+		log.Debugf("unable to parse jacoco xml report: %s", reportPath)
+	}
+
+	var currentFile *FileCoverage
+	for _, xmlPackage := range report.Packages {
+		for _, sourcefile := range xmlPackage.Sourcefiles {
+			currentFile = &FileCoverage{
+				Filename: fmt.Sprintf("%s/%s", xmlPackage.Name, sourcefile.Name),
+				Coverage: &Coverage{
+					FunctionsFound: 0,
```
fmeum, line 94: Are these explicit initializers necessary?

pkg/coverage/jacoco.go
```diff
@@ -0,0 +1,113 @@
...
+		c.FunctionsHit += counter.Covered
+	}
+}
+
+func ParseJacocoXML(reportPath string) *CoverageSummary {
+	summary := &CoverageSummary{
+		Total: &Coverage{},
+	}
+
+	data, _ := os.ReadFile(reportPath)
```
fmeum, line 81: Shouldn't we return this error rather than ignoring it?

pkg/coverage/jacoco.go
```diff
@@ -0,0 +1,113 @@
...
+	report := &jacocoReport{}
+	err := xml.Unmarshal([]byte(data), report)
+	if err != nil {
+		log.Debugf("unable to parse jacoco xml report: %s", reportPath)
```
fmeum, line 85: Continuing after this error probably doesn't make sense. We could just return it here.

share/init.gradle
```diff
@@ -9,4 +9,48 @@ allprojects {
...
+        println "cifuzz.test.buildDir=${layout.buildDirectory.get()}";
+    }
+
+    // we register own tasks for the report to avoid side effects on existing user tasks
+    // when overwriting config values (like output path)
+
+    // we need to set the exec file output path explicitly to make sure we find
+    // the file in the `cifuzzReport` tasks
+    def execFile = layout.buildDirectory.file("jacaco/cifuzz.exec").get().asFile
+    def fuzzTest = project.findProperty("cifuzz.fuzztest") ?: "*"
```
jochil, line 23: wouldn't this "*" execute all junit tests including non-fuzz tests?
My expectation would be, that even if I do not specify a specific fuzz test it will only execute fuzz tests in general

pkg/coverage/jacoco.go
```diff
@@ -0,0 +1,113 @@
...
+	case "BRANCH":
+		c.BranchesFound += counter.Covered + counter.Missed
+		c.BranchesHit += counter.Covered
+	case "METHOD":
+		c.FunctionsFound += counter.Covered + counter.Missed
+		c.FunctionsHit += counter.Covered
+	}
+}
+
+func ParseJacocoXML(reportPath string) *CoverageSummary {
```
fmeum, line 76: For functions that parse a file, it's usually better to have them take in an `io.Reader`. The caller is usually in a better position to handle non-existing files and tests can just pass in byte buffers backed by strings rather than paths to temporary files.

examples/maven/pom.xml
```diff
@@ -32,6 +32,25 @@
...
+         <build>
+           <directory>.cifuzz-build</directory>
+            <plugins>
+              <plugin>
+                <groupId>org.jacoco</groupId>
+                <artifactId>jacoco-maven-plugin</artifactId>
+               <configuration>
+                  <formats>HTML,XML</formats>
+                  <outputDirectory>${cifuzz.report.output}</outputDirectory>
+                </configuration>
```
jochil, line 47: This block needs more indentation

internal/build/gradle/gradle.go
```diff
@@ -174,3 +151,36 @@ func (b *Builder) getGradleCommand() (string, error) {
...
+
+func GetBuildDirectory(projectDir string) (string, error) {
+	cmd, err := buildGradleCommand(projectDir, []string{"printBuildDir"})
+	if err != nil {
+		return "", nil
+	}
+
+	log.Debugf("Command: %s", cmd.String())
+	output, err := cmd.Output()
+	result := buildDirRegex.FindStringSubmatch(string(output))
```
fmeum, line 182: Even though it would be unexpected here, we should handle the no match case and produce an easy to debug error.

fmeum (approved):

examples/maven/src/main/java/com/example/ExploreMe.java
fmeum: I left a comment below suggesting to pass one of the parameters in as a constructor argument, which would make this even more interesting.

share/init.gradle
florianGla: we moved the html report to a subfolder

pkg/coverage/jacoco.go
florianGla: We solve it in another ticket together with lcov parsing.

jochil (approved):

PR #384, opened 2022-10-26
Minor improvements to pkg/dependencies

adombeck:
* Print which version was found
* Print the path of the binary that was found
* Start log messages with uppercase


pkg/dependencies/version.go
```diff
@@ -92,7 +92,7 @@ func clangVersion(dep *Dependency, clangCheck execCheck) (*semver.Version, error
 			return nil, err
 		}
 		pathVersion, err := clangCheck(path, dep.Key)
-		log.Debugf("Found clang version %s in PATH", pathVersion.String())
```
fmeum, line 95: `$PATH` is Unix-specific, whereas `PATH` also makes sense on Windows. I like that we are now printing the value of the variable, but could we revert adding the `$`?

adombeck: Good point, I'll revert that

fmeum (approved):

PR #388, opened 2022-10-27
feat: uninstall command

mlsandnerCI:


adombeck:
I'm not sure if adding a subcommand to cifuzz is the best way to solve this. Lets discuss how we best want to solve the issue (also given that we now have CLI-575 it implemented, which changes the situation a bit).

PR #389, opened 2022-10-27
pkg/binary: Fix check for LLVM continuous mode support

adombeck:
Includes some minor improvements I found useful while trying to reproduce the issue.


internal/build/cmake/cmake.go
```diff
@@ -117,6 +118,9 @@ func (b *Builder) Configure() error {
 		"-DCIFUZZ_SANITIZERS=" + strings.Join(b.Sanitizers, ";"),
 		"-DCIFUZZ_TESTING:BOOL=ON",
 	}
+	if viper.GetBool("verbose") {
```
fmeum, line 121: Just a note, not a request to change anything: This setting will persist with the build directory, even if users leave out the `--verbose` flag later.

adombeck: Good point. I'll remove it, setting it via the CMake config is probably the better way to make CMake print verbose output.

fmeum (approved):

PR #393, opened 2022-10-28
build: Always set -UNDEBUG

fmeum:
Ensures that `assert`s are always enabled with both the "other" and "bazel" build system.


internal/build/bazel/bazel.go
```diff
@@ -133,6 +133,8 @@ func (b *Builder) BuildForRun(fuzzTests []string) ([]*build.Result, error) {
 	runFlags := []string{
 		// Build with debug symbols
 		"-c", "opt", "--copt", "-g",
+		// Enable asserts (disabled by --compilation_mode=opt.
```
adombeck, line 136:
```suggestion
		// Enable asserts (disabled by --compilation_mode=opt).
```

adombeck (approved):

PR #394, opened 2022-10-28
build/bazel: Fix debug symbols stripped when building bundle

adombeck:


internal/build/bazel/bazel.go
```diff
@@ -132,7 +132,10 @@ func (b *Builder) BuildForRun(fuzzTests []string) ([]*build.Result, error) {
 	// not supported by the other bazel commands we use
 	runFlags := []string{
 		// Build with debug symbols
-		"-c", "opt", "--copt", "-g",
+		"--copt", "-g",
+		// Tell bazel to keep debug symbols (the default compilation
+		// mode is "fastbuild" which strips debug symbols)
```
fmeum, line 137: The main reason is that `opt` leads to an optimized build, which conveniently already has debug symbols (similar to CMake's RelWithDebInfo). We don't have to mention fastbuild.

internal/build/bazel/bazel.go
```diff
@@ -239,14 +239,17 @@ func (b *Builder) BuildForBundle(engine string, sanitizers []string, fuzzTests [
 	}
 
 	// Flags which should only be used for bazel build
-	buildFlags := []string{
+	buildAndCQueryFlags := []string{
+		// Tell bazel to keep debug symbols (the default compilation
```
fmeum, line 243: Same comment as on the previous commit.

internal/build/bazel/bazel.go
adombeck: > We don't have to mention fastbuild.

Isn't the part that by default debug symbols are stripped quite essential, because we already specify `g` and the exact optimization level via `--copt`, so one could expect that the result already has debug symbols without setting the compilation mode. I think this comment might have helped me realize quickly that the `-c opt` was missing in the flags used to build fur bundle.

fmeum: That's a good point. Maybe just list both reasons? `-c opt` can be configured to do more than just `-O2`, so it's important to set here for "optimized" semantics.

adombeck: Like this?
```diff
-               // Tell bazel to keep debug symbols (the default compilation
-               // mode is "fastbuild" which strips debug symbols)
+               // Tell bazel to do an optimized build, which includes debug
+               // symbols (in contrast to the default "fastbuild" compilation
+               // mode which strips debug symbols).
```

fmeum (approved):

PR #395, opened 2022-10-31
fix: remove endless loop from Cleanup

MarkusZoppelt:


florianGla (approved):

fmeum:
It's quite surprising that none of the linters flagged this. IntelliJ does label it as a recursive call, but doesn't show the same clear warning it would show for Java.

PR #401, opened 2022-11-02
added error checking for ftell result

weichslgartner:
will resolve compiler `-Wsign-conversion` warning and prevent `malloc(-1)` if `run_file` is called on a non-seekable file like `/dev/stdin`.

### DIFF
--- a/tools/replayer/replayer_test.go
+++ b/tools/replayer/replayer_test.go
@@ -79,6 +79,7 @@ var clang = compilerCase{
 		"-Wall",
 		"-Wextra",
 		"-Werror",
+		"-Wconversion",
 		// Add debug info to make ASan and UBSan findings more useful.
 		"-g",
 		"-fsanitize=address,undefined",
@@ -100,6 +101,7 @@ var mingw = compilerCase{
 	[]string{
 		"-Wall",
 		"-Wextra",
+		"-Wconversion",
 		"-pedantic",
 		"-pedantic-errors",
 		"-Werror",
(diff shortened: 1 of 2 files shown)


fmeum:
Thanks for the fix! 

We are testing with `-Wall -Wextra`, but as we see from this issue that doesn't include `-Wsign-conversion`. Could you add that flag [here](https://github.com/CodeIntelligenceTesting/cifuzz/blob/94f6181aa6c37dad635186530765d79f1d2381bd/tools/replayer/replayer_test.go#L82) and [here](https://github.com/CodeIntelligenceTesting/cifuzz/blob/94f6181aa6c37dad635186530765d79f1d2381bd/tools/replayer/replayer_test.go#L103) so that we don't regress with that warning enabled?

weichslgartner:
@fmeum 

I added the flag to the linked locations.

fmeum:
@weichslgartner Thanks!

fmeum (approved):

weichslgartner:
I just checked I use -Wconversion in my project. Wconversion includes -Wsign-conversion (https://clang.llvm.org/docs/DiagnosticsReference.html#wsign-conversion). Can I changed to Wconversion @fmeum?

fmeum:
@weichslgartner Sure, we want the strictest (reasonable) warnings anyway.

weichslgartner:
@fmeum  Good, I updated the commit.

fmeum (approved):

PR #403, opened 2022-11-03
cmd/coverage: Support bazel

adombeck:
TODO:
* [x] lcov reports
* [ ] ~~Try to run the coverage command in minijail~~ Running bazel in minijail can cause all kinds of failure, for example:
```
2022/11/11 10:23:11 could not resolve the version 'latest' to an actual version number: unable to determine latest version: could not list Bazel versions in GCS bucket: could not list GCS objects at https://www.googleapis.com/storage/v1/b/bazel/o?delimiter=/: could not fetch https://www.googleapis.com/storage/v1/b/bazel/o?delimiter=/: Get "https://www.googleapis.com/storage/v1/b/bazel/o?delimiter=/": dial tcp: lookup www.googleapis.com on [::1]:53: read udp [::1]:49777->[::1]:53: read: connection refused
```


internal/build/bazel/bazel.go
```diff
@@ -208,6 +208,136 @@ func (b *Builder) BuildForRun(fuzzTests []string) ([]*build.Result, error) {
...
+		}
+		path, err := pathFromLabel(fuzzTest, commonFlags)
+		if err != nil {
+			return "", err
+		}
+		outputPath = filepath.Join(outputDir, path)
+	}
+
+	// Create an HTML report via genhtml
+	cmd = exec.Command("genhtml", "--output", outputPath, lcovReport)
```
fmeum: The coverage reports contain relative paths, which AFAIK `genhtml` resolves relative to the current working directory. Did you check whether this works if cifuzz is executed from a subdirectory of the project directory?

examples/bazel/.gitignore
```diff
@@ -2,3 +2,4 @@
 *_inputs
 /.cifuzz-corpus/
 /.cifuzz-findings/
+/.ijwb/
```
fmeum, line 5:
```suggestion
/.ijwb/
/.clwb/
```

internal/build/bazel/bazel.go
adombeck: Good catch, it did indeed fail when executed from a subdirectory! I pushed a fix.

fmeum (approved):

PR #404, opened 2022-11-03
fix replayer on arm64 macOS

MarkusZoppelt:
* added quick fix in replayer_test.go for setting `halt_on_error=1` UBSan default option
==> this is a bug in replayer.c and needs to be fixed in the future
* Replaced `sprintf` with `snprintf` for macOS in replayer.c

```
sprintf is deprecated. This function is provided for compatibility reasons
only.  Due to security concerns inherent in the design of sprintf(3), it is
highly recommended that you use snprintf(3) instead.
[-Werror,-Wdeprecated-declarations]
```


tools/replayer/src/replayer.c
```diff
@@ -302,7 +302,7 @@ static void run_dir_entry(const char *dir, const char *file) {
   sprintf_s(path, path_size, "%s\\%s", dir, file);
 #else
   /* sprintf_s is not available in Unix C90 system headers. */
-  sprintf(path, "%s/%s", dir, file);
+  snprintf(path, path_size, "%s/%s", dir, file);
```
fmeum, line 305: We have so far adhered to strict C90 compliance for the replayer. `snprintf` is only in C99 (https://en.cppreference.com/w/c/io/fprintf, (4)).

We can either decide that C99 is good enough or guard `snprintf` behind a check similar to the one we have for `_WIN32`, just for macOS. I would prefer the latter if it isn't too much work.

MarkusZoppelt: Added a guard as you suggested, thanks!

PR #405, opened 2022-11-03
cmd/remote-run: Support --json flag

adombeck:


internal/cmd/create/create.go
```diff
@@ -20,11 +22,25 @@ import (
...
 	BuildSystem string `mapstructure:"build-system"`
+	Interactive bool   `mapstructure:"interactive"`
 
 	outputPath string
 	testType   config.FuzzTestType
 }
 
+func (opts *createOpts) Validate() error {
+	if opts.Interactive {
+		opts.Interactive = term.IsTerminal(int(os.Stdin.Fd())) && term.IsTerminal(int(os.Stdout.Fd()))
```
fmeum, line 33: If I understand this correctly, there is jo way to force interactivity? Do you think that there could be situations where someone would need that?

internal/cmd/remote-run/remote_run.go
```diff
@@ -132,11 +138,17 @@ https://github.com/CodeIntelligenceTesting/cifuzz/issues`, system)
 				opts.ProjectName = "projects/" + opts.ProjectName
 			}
 
+			opts.Stdout = cmd.OutOrStdout()
+			opts.Stderr = cmd.ErrOrStderr()
+
+			// If --json was specified, print all build output to stderr
+			if opts.PrintJSON {
+				opts.Stdout = cmd.ErrOrStderr()
```
fmeum, line 146: Can we turn this into an if-else so that `opt.Stdout` is only set once in both cases?

fmeum (approved):

internal/cmd/create/create.go
adombeck: > If I understand this correctly, there is jo way to force interactivity? 

correct

> Do you think that there could be situations where someone would need that?

I can't imagine a situation where someone would need that. We're copying the behavior of `gh` here, so I think that should be fine.

internal/cmd/remote-run/remote_run.go
adombeck: done

internal/cmd/create/create.go
fmeum: Sounds good, just wanted to make sure we made this choice consciously.

fmeum (approved):

PR #407, opened 2022-11-07
minijail: Allow read-only access to the whole filesystem

adombeck:


pkg/minijail/minijail.go
```diff
@@ -155,37 +156,43 @@ var fixedMinijailArgs = []string{
...
 	"-k", "tmpfs,/dev/shm,tmpfs," + strconv.Itoa(MS_NOSUID|MS_NODEV|MS_STRICTATIME) + ",mode=1777",
+	// Applications generally assume that /tmp is writable, so we allow
+	// write access to /tmp.
+	// We could set $TMPDIR to a different directory instead and allow
+	// write access to that, but not all applications use $TMPDIR (but
+	// /tmp instead).
+	// We could also mount a different directory to /tmp, but that would
+	// lead to paths printed by the application not being accessible
+	// from the host.
+	// It should safe to assume that there are no critical files stored
```
fmeum, line 174: Not using a dedicated directory for `/tmp` means that we may see PID file collisions between PID namespaced sandboxes. For example, running multiple sandboxed Java fuzz tests may fail in the same way as https://github.com/bazelbuild/bazel/issues/3236.

Do we need the PID namespace? Bazel is using it to simplify killing the sandboxed process. Is that relevant for us?

adombeck: Do you think it would it work to mount a tmpfs on /tmp instead? On current main, we also don't mount /tmp from the host.

fmeum: That's exactly what I was just thinking. :-)

fmeum: I can see that work well for fuzzing, even though it doesn't for building where it's easy to exhaust your RAM that way.

adombeck: If we find that memory exhaustion is a problem, we can also mount an empty directory to /tmp, for example `/tmp/sandbox-tmp-123 -> <path to chroot>/tmp`

fmeum: Yes, that would be an alternative. I'm open for both. With modern I/O performance, the difference in performance may not even be that big.

adombeck: I'm now mounting a tmpfs on /tmp - and on /run and /var/run as well, because those actually contain access to sensitive files, for example the ssh keyring.

fmeum:
@adombeck tmpfs on `/tmp` makes it harder to get our integration tests to run. Maybe that's a reason to mount a dedicated subdirectory of `/tmp` instead?

adombeck:
> @adombeck tmpfs on /tmp makes it harder to get our integration tests to run

Why is that?

fmeum:
> > @adombeck tmpfs on /tmp makes it harder to get our integration tests to run
> 
> Why is that?

If everything passes that's fine. I just saw test failures due to our integration test data being copied under `/tmp` on the host and then not being accessible in the sandbox, but looks like you fixed that.

fmeum (approved):

PR #408, opened 2022-11-07
fix: use new cifuzz install path in pipelines 

mlsandnerCI:
- make cifuzz calls more robust by using $PATH instead of direct path
- min_requirements pipeline runs after every release is published


.github/workflows/pipeline_e2e_installer.yml
```diff
@@ -88,7 +88,7 @@ jobs:
         if: contains(matrix.os, 'windows')
         run: |
           cd examples/cmake
-          ~/cifuzz/bin/cifuzz run my_fuzz_test 2>&1 | tee fuzzing.log
+          ~/.local/bin/cifuzz run my_fuzz_test 2>&1 | tee fuzzing.log
           cat fuzzing.log | grep "heap buffer overflow"
           # we are not testing for the ubsan finding 
           # as we currently do not support ubsan on windows
```
mlsandnerCI, line 94: Just realized I have to change this to the windows one, will add this quickly

adombeck (approved):

jochil (approved):
lgtm

PR #413, opened 2022-11-08
feat(bazel): command completion creates absolute labels

mlsandnerCI:
Command completion now creates absolute labels if we can't find any other fuzz tests in our current directory.

The screenshot shows the command completion results in different directories in the bazel project.
Screenshot 07 11 2022 um 08 06 30 AM


adombeck:
> Command completion now creates absolute labels if we can't find any other fuzz tests in our current directory.

That's not what bazel commands like `bazel run` do though. I would prefer it if cifuzz handles command completion for bazel targets similar to how bazel does it, so that it's not confusing for our users. As far as I can tell, bazel handles command completion like this:
If the user already entered "//", it completes absolute targets, else it completes relative targets.

adombeck:
@mlsandnerCI: I pushed a commit which refactors the code a bit. Please take a look.

adombeck (approved):

PR #414, opened 2022-11-08
docs: expand cifuzz installation path section in readme

mlsandnerCI:
I thought including the executable path would be useful in the `Installation` section since we also remind to add the `bin` directory to the `PATH`.


MarkusZoppelt (approved):

jochil:
IMHO these uninstall and directory sections can be moved to another document under `/docs/` as they are additional information

jochil:
> > IMHO these uninstall and directory sections can be moved to another document under `/docs/` as they are additional information
> 
> I would have thought that `Install` and `Uninstall` would be the most basic information and therefore should be in the main ReadMe thinking You could argue that we don't want people to uninstall cifuzz so we shouldn't even give them the idea and only make the information available to the ones who actively seek it out.
> 
> What kind of title for the document would you propose?

I think it is important to keep the README concise and therefore only mention the most relevant use cases... we definitely need to explain how to uninstall cifuzz, but maybe not directly on the main entry point of our documentation ;)
Whats about a linking to a "additional installation info" document, containing the directory listing, uninstall methods and maybe additional installation ways (like from source?)

docs/Installation-Information.md
```diff
@@ -0,0 +1,64 @@
+# Install Locations
```
jochil, line 1:
```suggestion
# Installation Guide
```

docs/Installation-Information.md
```diff
@@ -0,0 +1,64 @@
...
+
```
jochil, line 2: What do you think about adding another headline `## Installation Directories` and move the Windows/Linux section one level down? If doing so the uninstall section would also be on h level 2 which seems more consistent to me

README.md
```diff
@@ -80,17 +80,8 @@ If you are using Windows you can download
...
-  * `$XDG_DATA_HOME/cifuzz` if set
-* *Linux / macOS* **root** installation:
-  * `/opt/code-intelligence/cifuzz`
-* *Windows* installation:
-  * `%APPDATA%/cifuzz`
-
-
-Do not forget to add the installation's `bin` directory to your `PATH`.
+Do not forget to add the installation's `bin` directory to your `PATH`. 
+You can find the installation locations in the [installation information guide](docs/Installation-Information.md).
```
jochil, line 84:
```suggestion
You can find additional information in our [Installation Guide](link).
```

jochil (approved):
IMO "Installation Guide" would be a good name

docs/Installation-Information.md
mlsandnerCI: Good idea, I included this 👍

PR #415, opened 2022-11-08
bundle: Ensure that the fuzz tests contain no duplicates

adombeck:


Makefile
```diff
@@ -54,7 +54,7 @@ deps/integration-tests:
 .PHONY: deps/dev
 deps/dev: deps
 	go install github.com/incu6us/goimports-reviser/v2@latest
-	go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.46.2
+	go install github.com/golangci/golangci-lint/cmd/golangci-lint@latest
```
fmeum, line 57: I would prefer pinning this to a more recent release instead to prevent "lints on my machine".

internal/bundler/bundler.go
```diff
@@ -97,6 +97,9 @@ type Opts struct {
 func (opts *Opts) Validate() error {
 	var err error
 
+	// Ensure that the fuzz tests contain no duplicates
+	opts.FuzzTests = sliceutil.RemoveDuplicates(opts.FuzzTests)
```
fmeum, line 101: Please also add a comment to the bundle methods that actually rely on this property.

fmeum (approved):

Makefile
adombeck: done

internal/bundler/bundler.go
adombeck: done

fmeum (approved):

PR #419, opened 2022-11-09
tests: include bazel example in pipelines

mlsandnerCI:


jochil (approved):

adombeck:
Is this ready for review?

jochil:
> Not really, the PR on bazels side is not merged yet, the `e2e_installer` test still fails.

But the actual tests in the pipeline should look like this, or?
Are there any changes we expect in the pipeline code?

jochil:
> I just thought that it wouldn't be good to merge something into main that causes a pipeline to fail

Absolutely... I was just asking to see if my PR approval was premature

adombeck:
> a) that is okay to add

Sure, we can and should add all dependencies to the container.

> b) we need to document that and

Yes, we should document all dependencies.

>  c) why does bazel need python for a java example?

We don't support using bazel with Java projects yet, so I assume it's using a C/C++ example instead. And I'm not sure why bazel needs Python, probably something we're using is written in Python. What did the error message say?

I also noticed yesterday that a JDK is required for the `bazel coverage` command. I created [a ticket](https://code-intelligence.atlassian.net/browse/CLI-665) to document that. We should also create a ticket to document the Python dependency (preferably, we figure out why Python is required and mention that in the commit message).

adombeck:
> There is still a bug in bazel that prevents the e2e_installer to pass but @fmeum will take care of that (https://github.com/bazelbuild/bazel/pull/16730).

If that only affects the e2e-installer  pipeline, can we split this PR into one part which adjusts the other pipelines (which we can then already merge) and one which only adjusts the e2e-installer pipeline (which is blocked on https://github.com/bazelbuild/bazel/pull/16730).

adombeck:
> > And I'm not sure why bazel needs Python, probably something we're using is written in Python. What did the error message say?
> 
> ```
> Error in fail: python interpreter `None` not found in PATH
> ERROR: /__w/cifuzz/cifuzz/examples/bazel/src/BUILD.bazel:10:13: //src:explore_me_fuzz_test_corpus depends on @rules_fuzzing//fuzzing/tools:make_corpus_dir in repository @rules_fuzzing which failed to fetch. no such package '@fuzzing_py_deps//': python interpreter `None` not found in PATH
> ```

So it's a dependency of the `rules_fuzzing` package. Let's note that and the error message in the commit message.

.github/workflows/pipeline_min_requirements.yml
```diff
@@ -19,7 +19,7 @@ jobs:
       - name: Install dependencies
         run: |
           apt update
-          apt install -y cmake llvm-11 clang-11 openjdk-8-jdk maven subversion curl unzip
+          DEBIAN_FRONTEND=noninteractive apt install -y cmake llvm-11 clang-11 openjdk-8-jdk maven subversion curl unzip git
```
adombeck, line 22: Why is git needed?

.github/workflows/pipeline_e2e_source.yml
```diff
@@ -56,6 +56,12 @@ jobs:
           unzip -d /opt/gradle /tmp/gradle-7.5.1-bin.zip
           echo "/opt/gradle/gradle-7.5.1/bin" >> $GITHUB_PATH
 
+      - name: Install Bazel
+        if: contains(matrix.container, 'archlinux') || contains(matrix.container, 'ubuntu')
```
adombeck, line 60: We don't need this condition because we only run this job on Ubuntu and Arch Linux - and if we would run it on other platforms as well, we would also need to install bazel.

.github/workflows/pipeline_min_requirements.yml
mlsandnerCI: To install bazel in the "Install Bazel" section. Would it work to only add it to the `apt` command (i.e. `apt install bazel`)?

adombeck (approved):

.github/workflows/pipeline_min_requirements.yml
adombeck: I don't see git being used in the "Install bazel" section. Is it a dependency of `bazelisk`?

mlsandnerCI: Not quite sure tbh. I added this line with Fabian and not including git caused an error if I remember correctly. It could be the dependency of bazelisk.

adombeck: > Would it work to only add it to the apt command (i.e. apt install bazel)?

I don't think there are Debian packages for bazel, so no, we can't install it via `apt`.

adombeck: > Not quite sure tbh. I added this line with Fabian and not including git caused an error if I remember correctly. It could be a dependency of bazelisk.

Ok, then I'm fine with keeping it

fmeum: We need git because we bring in the `cifuzz` repo in the Bazel example as a `git_repository`. Bazel doesn't implement git itself, it simply uses the host git.

adombeck: That makes sense, thanks!

PR #423, opened 2022-11-10
installer: Refactor and create symlinks to the cifuzz executable

adombeck:


cmd/installer/installer.go
```diff
@@ -102,11 +101,22 @@ func installCIFuzz(installDir string) error {
 			// https://gitlab.kitware.com/cmake/cmake/-/blob/5ed9232d781ccfa3a9fae709e12999c6649aca2f/Modules/Platform/UnixPaths.cmake#L30)
 			cmakeSrc := filepath.Join(installDir, "share", "cifuzz")
 			cmakeDest := "/usr/local/share/cifuzz"
-			log.Debugf("Creating symlink %s -> %s", cmakeSrc, cmakeDest)
-			err = copy.Copy(cmakeSrc, cmakeDest)
+			log.Printf("Creating symlink %s -> %s", cmakeSrc, cmakeDest)
```
jochil, line 104: IMO this can stay a`log.Debugf` as I would guess in normal use cases the user does not care

util/fileutil/fileutil.go
```diff
@@ -118,3 +118,22 @@ func CanonicalPath(path string) (string, error) {
 	}
 	return canonicalPath, nil
 }
+
+// ForceSymlink creates newname as a symbolic link to oldname. If
+// newname already exists, it will be overwritten.
+func ForceSymlink(oldname, newname string) error {
```
jochil, line 124: Maybe it makes sense to cover this util function with a unit test... what do you think?

cmd/installer/installer.go
adombeck: I think it's nice to tell the user which files and directories are created during installation. I pushed two more commits which make sure that's the case for all files and that no other output is printed (unless `--verbose` is used). That's around 4 or 5 lines, so it's not an overwhelming amount of output.

util/fileutil/fileutil.go
adombeck: done

mlsandnerCI (approved):

PR #427, opened 2022-11-10
implement 'cifuzz bundle' for Java

MarkusZoppelt:
This PR enables running `cifuzz bundle` on Java projects using Maven / Gradle.

**NOTE:** We'll display a warning message since Java remote-runs are currently not compatible and we cannot verify the correctness of the resulting archive bundles.


pkg/artifact/metadata.go
```diff
@@ -18,6 +18,7 @@ type Metadata struct {
 // Fuzzer specifies the type and locations of fuzzers contained in the archive.
 type Fuzzer struct {
 	Target    string `yaml:"target"`
+	Name      string `yaml:"name,omitempty"`
```
fmeum, line 21: `Name` is missing from the commit message

internal/build/gradle/gradle.go
```diff
@@ -188,6 +188,13 @@ func GetBuildDirectory(projectDir string) (string, error) {
 
 	log.Debugf("Command: %s", cmd.String())
 	output, err := cmd.Output()
+	if err != nil {
```
fmeum, line 191: If we didn't handle this error before, that appears to be more of a bugfix. Could you move this into a separate commit?

internal/bundler/jazzer.go
```diff
@@ -8,17 +24,168 @@ func newJazzerBundler(opts *Opts) *jazzerBundler {
...
+				// Hence, there will be a single entry for the runtime path but
+				// multiple entries for the archive manifest.
+				relPath, err := filepath.Rel(buildResult.BuildDir, runtimeDep)
+				if err != nil {
+					return nil, nil, errors.WithStack(err)
+				}
+				runtimePaths = append(runtimePaths, filepath.Join(workDirPath, relPath))
+
+				err = filepath.WalkDir(runtimeDep, func(path string, d os.DirEntry, err error) error {
+					if err != nil {
```
fmeum, line 81: Can't you use `artifact.AddDirToManifest` here?

internal/bundler/jazzer.go
```diff
@@ -8,17 +24,168 @@ func newJazzerBundler(opts *Opts) *jazzerBundler {
...
+						return err
+					}
+					if d.IsDir() {
+						return nil
+					}
+					relPath, err := filepath.Rel(buildResult.BuildDir, path)
+					if err != nil {
+						return err
+					}
+					manifest[filepath.Join(workDirPath, relPath)] = path
```
fmeum, line 91: Why are you adding the files to `work_dir`? My understanding of the spec is that it's meant for files the fuzz test itself creates at runtime. Since the build dir differs per `buildResult`, I'm also worried about collisions between fuzz tests.

internal/cmd/bundle/bundle.go
```diff
@@ -113,21 +113,25 @@ system configured for the project.
...
 interested in using this feature on %s, please file an issue at
-https://github.com/CodeIntelligenceTesting/cifuzz/issues`, system)
+https://github.com/CodeIntelligenceTesting/cifuzz/issues`, opts.BuildSystem, system)
 				log.Print(err.Error())
 				return cmdutils.WrapSilentError(err)
 			}
 
 			opts.FuzzTests = args
+			if isOSIndependent {
+				log.Warnf("EXPERIMENTAL WARNING: Bundling for %s is experimental and is currently not supported for remote runs.",
```
fmeum, line 132: I would suggest dropping the `EXPERIMENTAL ` prefix, it's not the warning that is experimental ;⁠-⁠)

internal/bundler/jazzer_test.go
```diff
@@ -0,0 +1,86 @@
...
+	buildResults := []*build.Result{}
+	buildResult := &build.Result{
+		Name:        fuzzTest,
+		BuildDir:    buildDir,
+		SeedCorpus:  seedCorpus,
+		Engine:      "JAVA_LIBFUZZER",
+		RuntimeDeps: []string{filepath.Join(libraryPath, "lib", "mylib.jar")},
+		ProjectDir:  projectDir,
+	}
+	buildResults = append(buildResults, buildResult)
```
fmeum, line 55: Could you add another buildResult here to verify that relative lath collisions are handled correctly?

internal/bundler/jazzer_test.go
```diff
@@ -0,0 +1,86 @@
...
+	libraryPath, err := os.MkdirTemp(tempDir, "cifuzz-lib-*")
+	require.NoError(t, err)
+	defer fileutil.Cleanup(libraryPath)
+	err = os.MkdirAll(filepath.Join(libraryPath, "lib"), 0o755)
+	require.NoError(t, err)
+	err = fileutil.Touch(filepath.Join(libraryPath, "lib", "mylib.jar"))
+	require.NoError(t, err)
+	runtimeDeps := []string{
+		// A library in the project's build directory.
+		filepath.Join(libraryPath, "lib", "mylib.jar"),
```
fmeum, line 43: Please also test a directory structure of `.class` files here..

internal/bundler/jazzer_test.go
```diff
@@ -0,0 +1,86 @@
...
+		Seeds:        filepath.Join("work_dir", "seeds"),
+		RuntimePaths: expectedDeps,
+		EngineOptions: artifact.EngineOptions{
+			Env:   b.opts.Env,
+			Flags: b.opts.EngineArgs,
+		},
+	}
+	require.Equal(t, *fuzzer, *fuzzers[0])
+
+	// TODO: Check manifest
```
fmeum, line 84: Please do ;⁠)

internal/cmd/bundle/bundle.go
MarkusZoppelt: Done

pkg/artifact/metadata.go
MarkusZoppelt: Done

internal/build/gradle/gradle.go
MarkusZoppelt: Done

internal/bundler/jazzer_test.go
MarkusZoppelt: Added some

internal/bundler/jazzer.go
MarkusZoppelt: I removed the work_dir prefix. Now the new bundle structure looks like this (example):
```
├── mylib.jar
├── bundle.yaml
├── classes
│  └── com
│     └── example
│        ├── App.class
│        └── ExploreMe.class
├── jazzer-0.13.0.jar
├── jazzer-api-0.13.0.jar
├── jazzer-junit-0.13.0.jar
├── ...
├── test-classes
│  └── com
│     └── example
│        └── FuzzTestCase.class
└── work_dir
```

internal/bundler/jazzer_test.go
MarkusZoppelt: done

internal/bundler/jazzer_test.go
MarkusZoppelt: done

internal/bundler/jazzer.go
MarkusZoppelt: I don't think so iirc...

As we try to explain in the comment a few lines above:
```
// if the current runtime dep is a directory, add all files to
// the archive but add just the directory path to the runtime
// paths. Hence, there will be a single entry for the runtime
// path but multiple entries for the archive manifest.
```

Using `artifact.AddDirToManifest` would add multiple entries.

To visualise the difference:

We want the manifest to look like this:
```
[...]
  runtime_paths:
    - [...]
    - jazzer-0.13.0.jar
    - [...]
    - classes
    - test-classes
```

instead of this:
```
[...]
  runtime_paths:
    - [...]
    - jazzer-0.13.0.jar
    - [...]
    - classes
    - classes/com
    - classes/com/example
    - classes/com/example/MyClass.class
    - test-classes
    - [...]
```

@jochil @florianGla Is that correct?

jochil: > We want the manifest to look like this

not quite as I remembered... the `runtime_paths` should look like this (only the directories), but the archive manifest should contain all subfiles/-directories (to make sure they got copied). IMO that was the reason not to use `AddDirToManifest`

florianGla:
The seeds are not yet specified in the bundle.yaml file.

internal/bundler/jazzer.go
fmeum: Yes, the runtime paths should only contain the top-level directories, but that's covered by line 78. The `WalkDir` call has to collect all files recursively and that's what I would expect is covered by the existing manifest function.

MarkusZoppelt: @fmeum Actually, you were right about using the function instead.

I replaced the `WalkDir` block with `artifact.AddDirToFileMap` function (we renamed it in the meantime) and the artifact structure looks correct afaik.

internal/bundler/jazzer.go
```diff
@@ -0,0 +1,241 @@
...
+			MaxRunTime: uint(b.opts.Timeout.Seconds()),
+		}
+		fuzzers = append(fuzzers, fuzzer)
+	}
+	return fuzzers, archiveFileMap, nil
+}
+
+func (b *jazzerBundler) copySeeds(archiveFileMap artifact.FileMap) (string, error) {
+	archiveSeedDir := "seeds"
+	err := prepareSeeds(b.opts.SeedCorpusDirs, archiveSeedDir, archiveFileMap)
```
fmeum, line 130: Does this distinguish seeds for different fuzz tests?

internal/build/build.go
```diff
@@ -153,3 +157,46 @@ func JazzerSeedCorpus(targetClass string, projectDir string) string {
...
+			}
+			if match == true {
+				classPath, err := filepath.Rel(testDir, path)
+				if err != nil {
+					return errors.WithStack(err)
+				}
+
+				className := strings.TrimSuffix(filepath.Base(path), ".java")
+				classPath = filepath.Join(filepath.Dir(classPath), className)
+				classPath = strings.ReplaceAll(classPath, string(os.PathSeparator), ".")
```
fmeum, line 188: `classPath` is a somewhat confusing name here in the context of Java. Maybe inline the intermediate result and call it `className` instead?

internal/build/build.go
```diff
@@ -153,3 +157,46 @@ func JazzerSeedCorpus(targetClass string, projectDir string) string {
...
+			if filepath.Ext(path) != ".java" {
+				return nil
+			}
+
+			bytes, err := os.ReadFile(path)
+			if err != nil {
+				return errors.WithStack(err)
+			}
+
+			match, err := regexp.MatchString(`@FuzzTest|fuzzerTestOneInput\(`, string(bytes))
```
fmeum, line 176: Can we share this regex with other parts of the codebase?

It would also make sense to precompile it.

While we are at it, `@FuzzTest|\sfuzzerTestOneInput\(` would be slightly more correct.

internal/bundler/jazzer.go
```diff
@@ -220,7 +249,7 @@ func (b *jazzerBundler) createManifestJar(targetClass string) (string, error) {
 	jarWriter := zip.NewWriter(jarFile)
 
 	// create explicit parent directory in zip file
-	_, err = jarWriter.Create("META-INF")
+	_, err = jarWriter.Create("META-INF" + string(os.PathSeparator))
```
fmeum, line 252: I am not 100% sure, but I have only every seen `/` as a separator in ZIP files. If it works, it's preferable here as it makes the generated archive independent of the host system.

fmeum:
From the Windows integration test:
```
remove C:\Users\RUNNER~1\AppData\Local\Temp\cifuzz-bundle-4233131296\com.example.FuzzTestCase\manifest.jar: The process cannot access the file because it is being used by another process.
```
We could be cleaning up the bundle temp dir too early.

internal/bundler/jazzer.go
```diff
@@ -0,0 +1,278 @@
...
+		archiveManifestPath := filepath.Join(buildResult.Name, "manifest.jar")
+		archiveFileMap[archiveManifestPath] = manifestJar
+		// making sure the manifest jar is the first entry in the class path
+		runtimePaths := []string{
+			archiveManifestPath,
+		}
+
+		for _, runtimeDep := range buildResult.RuntimeDeps {
+			// check if the file exists
+			entry, err := os.Stat(runtimeDep)
```
fmeum, line 85: Could you print these paths with quotes to debug the CI failure? I suspect that it is an escaping problem or a space snuck in somewhere.

internal/build/build.go
florianGla: yes, that makes sense, we use a really similar code block for autocompletion of fuzz tests

internal/bundler/jazzer.go
florianGla: Good catch, that is indeed the case

internal/bundler/jazzer.go
florianGla: It works on my local windows machine

florianGla:
Unzipping the manifest.jar causes problems on windows workers. We should ignore this test for now on windows. Maybe it works again after archiveutils.Unzip is fixed. I have tested the generated manifest.jar on windows locally.

internal/bundler/jazzer.go
fmeum: I think this is not correct according to the ZIP spec, see https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT (4.4.17.1).

florianGla: Thank you, I changed it to a forward slash.

fmeum (approved):

internal/bundler/jazzer.go
```diff
@@ -244,15 +244,22 @@ func (b *jazzerBundler) createManifestJar(targetClass string) (string, error) {
 	}
 	defer jarFile.Close()
 	jarWriter := zip.NewWriter(jarFile)
+	defer jarWriter.Close()
 
 	// create explicit parent directory in zip file
-	_, err = jarWriter.Create("META-INF/")
+	fh := &zip.FileHeader{}
+	fh.Name = "META-INF/"
```
fmeum, line 251: Please use a struct literal instead.

fmeum (approved):

PR #429, opened 2022-11-11
Fix installation guide

adombeck:


docs/Installation-Guide.md
```diff
@@ -4,61 +4,71 @@
...
-All files will be installed to `%APPDATA%/cifuzz` with the executable located in `%APPDATA%/cifuzz/bin`.
+All files are installed to `%APPDATA%/cifuzz` with the executable located
+in `%APPDATA%/cifuzz/bin`.
 
 ## How to uninstall cifuzz
 
-### Version <= 0.6.0
+### Linux / macOS
+
+#### Version < 0.7.0
```
mlsandnerCI, line 27: Jochen mentioned in a previous PR review that it would be more obvious to write <=0.6.0 instead of < 0.7.0. We could add the patch number to be absolutely correct.

docs/Installation-Guide.md
```diff
@@ -4,61 +4,71 @@
...
+### Windows
+
+To uninstall cifuzz on Windows:
 
 ```bash
 rd /s %APPDATA%/cifuzz
 ```
 
-To delete the registry entries on **Windows**:
+To delete the registry entries:
```
mlsandnerCI, line 71: I already debated before that this could just be merged with the `rd` command in one block, because Windows users would probably know that that command will delete registry entries. What do you think?

mlsandnerCI (approved):

docs/Installation-Guide.md
adombeck: I don't think many Windows users know the `rd` command, so I prefer keeping the description.

docs/Installation-Guide.md
adombeck: I disagree that `<= 0.6.0` and `>= 0.7.0` is more obvious than `< 0.7.0` and `>= 0.7.0`. IMO, when you look at any version number, it's very clear which of the two cases applies.

docs/Installation-Guide.md
mlsandnerCI:
```suggestion
To uninstall cifuzz on Windows and to delete the corresponding registry entries:
```bash
rd /s %APPDATA%/cifuzz

reg delete "HKLM\Software\Kitware\CMake\Packages\cifuzz" /f 2> nul
reg delete "HKCU\Software\Kitware\CMake\Packages\cifuzz" /f 2> nul
```
I would like the keep the description too, I just meant to fuse it into one code block.

docs/Installation-Guide.md
jochil: I have no strong opinion on this. My thought was that you can easily spot the section that applies for your specific version if you have 0.6.0 for example.
But we can also make >= 0.7.0 the default section without mentioning the version number explicitly and move <= 0.6.0 to `#### Old versions` or `#### Version prior v0.7.0

adombeck: > But we can also make >= 0.7.0 the default section without mentioning the version number explicitly and move <= 0.6.0 to #### Old versions or `#### Version prior v0.7.0

How about we do that the next time we update the installation guide? For now, all current installations are < 0.7.0, so that section is currently very relevant. After a few releases, we can make the >= 0.7.0 section the default.

jochil: sounds good

PR #440, opened 2022-11-15
fix(installer): --no-symlink flag

mlsandnerCI:
Due to the unwanted action that the symlink will be changed in tests, I added a `--no-symlink` flag to prevent this. 
I also expanded the symlink logs to not confuse the user because we still create some symlinks even with the flag (cmake and completion scripts) just not of the cifuzz executable (which is specified in the description of the flag).


adombeck:
> Due to the unwanted action that the symlink will be changed in tests

Can you elaborate on that? What exactly is the issue with the symlink in the tests?

adombeck:
I see, that makes sense. But we have the same issue with all the symlinks created by the installer, not only the one created in the `bin` directory: They all install some component of cifuzz on the host system, which we don't want when running the tests. So IMO the flag should avoid creating any symlinks (and be named `--no-symlinks`).

adombeck:
Lets try to better explain the problem we're solving in the commit message. Here is my proposal:

Running the installer in tests creates symlinks to cifuzz and various components on the host system (e.g. `~/.local/bin` and `.cmake/packages/cifuzz`). We don't want those to point to the temporary cifuzz installation directory created for the tests. To avoid that, the installer now supports a `--no-symlinks` flag.

adombeck:
> I wasn't sure if the other symlinks could be removed

You have a point here, in the case of the CMake package symlink, we probably need a workaround which allows CMake to still find the package. IIUC, setting a [`CMAKE_PREFIX_PATH`](https://cmake.org/cmake/help/latest/variable/CMAKE_PREFIX_PATH.html#variable:CMAKE_PREFIX_PATH) or `CIFUZZ_ROOT` CMake variable to the correct path below the installation directory should work, see https://cmake.org/cmake/help/latest/command/find_package.html#id9

adombeck:
> in the case of the CMake package symlink, we probably need a workaround which allows CMake to still find the package

Alternatively, we could run the integration tests in Docker when they're run locally (in the CI they're run in a fresh VM, so we don't care about the system being modified there).

PR #442, opened 2022-11-16
ci(pr): Check that there are no WIP or XXX commits on the branch

adombeck:


adombeck:
> I didn’t manage to get the actual commit messages, the current branch only contains a single commit with message “Merge X into Y”.

Got it to work now

> Even when working as expected, this job would cause the PR checks to fail, making it impossible to merge. Sometimes, it’s legitimate for a commit message to include “WIP” or “XXX”, for example the message of the commit introducing this check would read “Check that there are no WIP or XXX commits on the branch” which causes the job to fail. What we want instead is a clear indication on the PR that there are WIP or XXX commits but allow the developer to still merge the PR. Unfortunately, GitHub doesn’t support something like that currently.

The test now only fails if the commit message contains any lines starting with WIP, fixup or XXX.

jochil:
> What we want instead is a clear indication on the PR that there are WIP or XXX commits but allow the developer to still merge the PR. Unfortunately, GitHub doesn’t support something like that currently.

As long you do not make a certain pipeline check required in the branch protection rules this is the default behavior. You could have added the check to the `pipeline_pr.yaml` and just let it fail.

IMHO another solution could have been using an already existing action like this: https://github.com/marketplace/actions/gs-commit-message-checker
With that it could have been a simple regex, without fetching the specific commit range and grep it by hand

adombeck:
> As long you do not make a certain pipeline check required in the branch protection rules this is the default behavior. You could have added the check to the pipeline_pr.yaml and just let it fail.

Yes, I already learned that yesterday from @fmeum - we briefly discussed that but found that we do actually want the test to be required, because else PRs will be merged when auto-merge was enabled but the job fails.

> IMHO another solution could have been using an already existing action like this: https://github.com/marketplace/actions/gs-commit-message-checker
With that it could have been a simple regex, without fetching the specific commit range and grep it by hand

I saw that action but decided not to use it because:
1. I don't like drawing in dependencies for such trivial tasks
2. That action uses the GitHub API to fetch the current commits on the PR, which could be different than the commits the pipeline runs on, if another push happened between the job being triggered and the API request being sent. My solution always uses the commit range which the job was triggered for.

PR #445, opened 2022-11-17
Remove undefined behavior bug from cpp example projects

florianGla:


.github/workflows/pipeline_e2e_installer.yml
```diff
@@ -109,8 +107,6 @@ jobs:
           cd examples/cmake
           cifuzz run my_fuzz_test 2>&1 | tee fuzzing.log
           cat fuzzing.log | grep "heap buffer overflow"
-          # we are not testing for the ubsan finding 
-          # as we currently do not support ubsan on windows
```
jochil, line 113: This was the reason we had separate blocks for windows and mac|linux. We can now run the cmake test for all OSs in one block

.github/workflows/pipeline_e2e_source.yml
```diff
@@ -128,17 +126,14 @@ jobs:
...
+          cat fuzzing.log | grep -E "Findings:\s+1"
         shell: bash
 
       - name: Run CMake Example (Windows)
         if: contains(matrix.os, 'windows')
         run: |
           cd examples/cmake
           cifuzz run -v my_fuzz_test 2>&1 | tee fuzzing.log
-          # we are not testing for the ubsan finding 
-          # as we currently do not support ubsan on windows
```
jochil, line 141: same as above... no need for the distinction anymore

jochil (commented):
The integration tests have their own projects in testdata, or? This is the reason for not having changes in them?

jochil (approved):
all the removed code :heart:

PR #446, opened 2022-11-17
add python and bazel dependencies to docs

MarkusZoppelt:
Python is a dependency of the `rules_fuzzing` package (bazel) and needs to be installed on the users system.
On a minimal Arch install, python is not necessarily installed, so we mention it as a dependency.

The following error message will appear otherwise:

    Error in fail: python interpreter `None` not found in PATH
    ERROR: /__w/cifuzz/cifuzz/examples/bazel/src/BUILD.bazel:10:13: //src:explore_me_fuzz_test_corpus depends on @rules_fuzzing//fuzzing/tools:make_corpus_dir in repository @rules_fuzzing which failed to fetch. no such package '@fuzzing_py_deps//': python interpreter `None` not found in PATH

Also add Windows section according to e2e pipeline.


README.md
```diff
@@ -137,18 +137,31 @@ choco install cmake llvm
 <!-- when changing this, please make sure it is in sync with the E2E pipeline -->
 ```bash
 sudo apt install clang llvm lcov
+curl -L https://github.com/bazelbuild/bazelisk/releases/download/v1.15.0/bazelisk-linux-amd64 -o /usr/bin/bazel
+chmod +x /usr/bin/bazel
```
adombeck, line 141: * This will fail when executed as non-root, because `/usr/bin` is only writable by root
* Only package managers should write to `/usr/bin`. `/usr/local/bin` would be more appropriate.
* I would prefer this URL to always install the latest bazelisk: https://github.com/bazelbuild/bazelisk/releases/latest/download/bazelisk-linux-amd64

README.md
```diff
@@ -137,18 +137,31 @@ choco install cmake llvm
...
 ```
 
 **Arch**
 <!-- when changing this, please make sure it is in sync with the E2E pipeline -->
 ```bash
-sudo pacman -S clang llvm lcov
+sudo pacman -S clang llvm lcov python
```
adombeck, line 147: Does this install Python 2or Python 3 on arch? Because we should install Python 3.

Lets mention in the commit message why Python is required:

Python is a dependency of the `rules_fuzzing` package, when not installed it fails with:

    Error in fail: python interpreter `None` not found in PATH
    ERROR: /__w/cifuzz/cifuzz/examples/bazel/src/BUILD.bazel:10:13: //src:explore_me_fuzz_test_corpus depends on @rules_fuzzing//fuzzing/tools:make_corpus_dir in repository @rules_fuzzing which failed to fetch. no such package '@fuzzing_py_deps//': python interpreter `None` not found in PATH

MarkusZoppelt: It installs python3: https://archlinux.org/packages/core/x86_64/python/

README.md
MarkusZoppelt: updated it.

README.md
MarkusZoppelt: added the explanation to the commit message

README.md
```diff
@@ -137,18 +137,31 @@ choco install cmake llvm
...
+chmod +x /usr/local/bin/bazel
 ```
 
 **Arch**
 <!-- when changing this, please make sure it is in sync with the E2E pipeline -->
 ```bash
-sudo pacman -S clang llvm lcov
+sudo pacman -S clang llvm lcov python
+curl -L https://github.com/bazelbuild/bazelisk/releases/latest/download/bazelisk-linux-amd64 -o /usr/local/bin/bazel
+chmod +x /usr/local/bin/bazel
```
adombeck, line 149: Same here regarding `sudo`

README.md
```diff
@@ -137,18 +137,31 @@ choco install cmake llvm
 <!-- when changing this, please make sure it is in sync with the E2E pipeline -->
 ```bash
 sudo apt install clang llvm lcov
+curl -L https://github.com/bazelbuild/bazelisk/releases/latest/download/bazelisk-linux-amd64 -o /usr/local/bin/bazel
+chmod +x /usr/local/bin/bazel
```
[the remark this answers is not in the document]
MarkusZoppelt: dang, of course! fixed.

adombeck (approved):

PR #449, opened 2022-11-17
docs: regression test guide

mlsandnerCI:
I added a gif of a screen recording to show the use in the IDE like we did in the cmake section.


docs/How-To-Write-A-Fuzz-Test.md
```diff
@@ -94,3 +94,27 @@ ctest --preset="cifuzz (Regression Test)"
...
+to be added to the `.bazelrc` file in the project workspace.
+
+```
+# --config=asan-replay
+build:asan-replay --@rules_fuzzing//fuzzing:cc_engine=@rules_fuzzing//fuzzing/engines:replay
+build:asan-replay --@rules_fuzzing//fuzzing:cc_engine_instrumentation=none
+build:asan-replay --@rules_fuzzing//fuzzing:cc_engine_sanitizer=asan
+```
+
+This allows the bazel flag `--config=asan-replay` to be added to a fuzz test run.
```
adombeck, line 110: Lets explain here what the flag does:

This allows to run fuzz tests via `bazel test --config=asan-replay`, which executes the fuzz test with all the input files in the corpus directory (`<fuzz test name>_inputs` by default) and then exits with exit code 0 if none of the inputs resulted in a finding. This can be used as a regression test, testing that the inputs which previously resulted in findings do not result in findings anymore.

docs/How-To-Write-A-Fuzz-Test.md
```diff
@@ -94,3 +94,27 @@ ctest --preset="cifuzz (Regression Test)"
...
+build:asan-replay --@rules_fuzzing//fuzzing:cc_engine_sanitizer=asan
+```
+
+This allows the bazel flag `--config=asan-replay` to be added to a fuzz test run. 
+
+To run a fuzz test with this configuration in your IDE, it has to be
+executed from the `BUILD.bazel` where it is defined as a `cc_fuzz_test`. 
+Choose "Modify Run Configuration" and add the bazel flag.
+
+Running the fuzz test now via "Run '...'" executes it in regression test mode.
```
adombeck, line 116: This part is specific to JetBrains IDEs

```suggestion
In JetBrains IDEs with the [bazel plugin](https://plugins.jetbrains.com/plugin/8609-bazel) installed, a fuzz test can be run with this configuration from the `BUILD.bazel` file which defines the `cc_fuzz_test` after adding the flag via "Modify Run Configuration".
```

docs/How-To-Write-A-Fuzz-Test.md
mlsandnerCI: This is also the general explanation of replaying/running it as a regression test, right? Maybe we should write that in the general section?

mlsandnerCI: We do have that explanation in the "Getting started" guide section "Regression testing" (except the exit code mentioning). Maybe we should expand that?

docs/Regression-Testing.md
```diff
@@ -0,0 +1,69 @@
...
+2. As a **regression test**, by invoking it through your IDE/editor or by directly
+   executing the replayer binary. This will run the fuzz test on all inputs stored 
+   in the`<fuzz-test-name>_inputs` next to the fuzz test. 
+
+The regression test mode can be used to test that inputs which previously resulted 
+in findings do not result in findings anymore. 
+
+In case a crash was found in this mode, a directory will be created and the crashing 
+input is added to it automatically. The fuzz test will stop immediately after 
+applying all inputs or earlier if a regression occurs.
```
adombeck, line 20: That's not correct, the replayer doesn't store any new inputs, it only reads inputs and runs the fuzz test on them.

mlsandnerCI: Then I'm confused what this part was referring to before.

adombeck: I think the "In case a crash was found" was meant as "In case a crash was found while fuzzing". So in the other mode, not the regression test mode. IMO we should rework this whole section at some point, but for now the changes are already an improvement - just drop the "in this mode" again.

mlsandnerCI: Ah okay, thanks 👍

jochil:
> I created a (draft) regression test guide and linked it in the getting started guide. @MarkusZoppelt @jochil @adombeck What do you think of this split up?

I think the split is OK that way... unfortunately the remaining guides, especially "How to write a fuzz test" are getting a little bit empty and I am sure there is more to tell than we are doing right now

adombeck (approved):

PR #450, opened 2022-11-17
cmd(run): expanding usage message for bazel

mlsandnerCI:


adombeck:
> I personally would have kept the <fuzz test> is the name of the cc_fuzz_test target **as defined in your BUILD.bazel**

The issue I see with that is that bazel supports both the name `BUILD` and `BUILD.bazel`. The bazel documentation refers to them as BUILD files:
https://bazel.build/concepts/build-files
https://bazel.build/concepts/build-ref

We could use that name.

adombeck:
> I can add this format to the bundle command where the usage message needs to be expanded for bazel too. It would probably make sense to keep outputs in a similar format.

Sounds good!

adombeck:
> The bazel documentation refers to them as BUILD files:
> https://bazel.build/concepts/build-files
> https://bazel.build/concepts/build-ref
> 
> We could use that name.

Done

adombeck (approved):

PR #452, opened 2022-11-18
docker: Add cifuzz-builder Dockerfile

adombeck:
The cifuzz-builder Dockerfile builds a cifuzz installer on Ubuntu 20.04
when built. This can be done via the `installer-via-docker` Makefile
target.


Makefile
```diff
@@ -172,3 +172,8 @@ site/update:
 	git -C site add -A
 	git -C site commit -m "update docs" || true
 	git -C site push
+
+.PHONY: installer-via-docker
+installer-via-docker:
+	mkdir -p build
+	DOCKER_BUILDKIT=1 docker build -f docker/cifuzz-builder/Dockerfile . --target bin --output build/bin
```
MarkusZoppelt, line 179:
```suggestion
	DOCKER_BUILDKIT=1 docker build --platform linux/amd64 -f docker/cifuzz-builder/Dockerfile . --target bin --output build/bin
```

This allows macOS M1 users to build the image on ARM.

adombeck: I assume that the resulting installer doesn't work on macOS though, right? We should probably make it clear that the `installer-via-docker` creates a linux installer.

MarkusZoppelt: Yes, this is the error:
```
> make installer-via-docker
mkdir -p build
DOCKER_BUILDKIT=1 docker build --platform linux/amd64 -f docker/cifuzz-builder/Dockerfile . --target bin --output build/bin
[+] Building 0.7s (12/12) FINISHED
 => [internal] load build definition from Dockerfile                                                                                                                                                 0.0s
 => => transferring dockerfile: 37B                                                                                                                                                                  0.0s
 => [internal] load .dockerignore                                                                                                                                                                    0.0s
 => => transferring context: 2B                                                                                                                                                                      0.0s
 => [internal] load metadata for docker.io/library/ubuntu:20.04                                                                                                                                      0.5s
 => [build 1/7] FROM docker.io/library/ubuntu:20.04@sha256:450e066588f42ebe1551f3b1a535034b6aa46cd936fe7f2c6b0d72997ec61dbd                                                                          0.0s
 => [internal] load build context                                                                                                                                                                    0.1s
 => => transferring context: 293.95kB                                                                                                                                                                0.1s
 => CACHED [build 2/7] RUN rm -f /etc/apt/apt.conf.d/docker-clean &&     echo 'Binary::apt::APT::Keep-Downloaded-Packages "true";' > /etc/apt/apt.conf.d/keep-cache                                  0.0s
 => CACHED [build 3/7] RUN --mount=type=cache,target=/var/cache/apt,sharing=locked   --mount=type=cache,target=/var/lib/apt,sharing=locked     export DEBIAN_FRONTEND=noninteractive &&     apt-get  0.0s
 => CACHED [build 4/7] RUN --mount=type=cache,target=/var/cache/go-src     SHA256=$(sha256sum /var/cache/go-src/go.tar.gz | cut -d ' ' -f 1);     if [ "${SHA256:-}" != "5e8c5a74fe6470dd7e055a461a  0.0s
 => CACHED [build 5/7] WORKDIR /src                                                                                                                                                                  0.0s
 => CACHED [build 6/7] COPY . .                                                                                                                                                                      0.0s
 => CACHED [build 7/7] RUN --mount=type=cache,target=/root/.cache/go-build     --mount=type=cache,target=/root/go/pkg/mod     make installer                                                         0.0s
 => ERROR [bin 1/1] COPY --from=build /src/build/bin/cifuzz_installer_linux /                                                                                                                        0.0s
------
 > [bin 1/1] COPY --from=build /src/build/bin/cifuzz_installer_linux /:
------
failed to compute cache key: failed to walk /var/lib/docker/tmp/buildkit-mount2173173800/src/build/bin: lstat /var/lib/docker/tmp/buildkit-mount2173173800/src/build/bin: no such file or directory
make: *** [installer-via-docker] Error 1

```

adombeck: Which docker version are you using?

MarkusZoppelt: > I assume that the resulting installer doesn't work on macOS though, right? We should probably make it clear that the installer-via-docker creates a linux installer.

Sorry, didn't read that right. Yes, the installer would be for an x86 Linux target.

docker/cifuzz-builder/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+ENV PATH="${PATH}:/usr/lib/llvm-12/bin"
+ENV CC=/usr/lib/llvm-12/bin/clang
+ENV CXX=/usr/lib/llvm-12/bin/clang++
+
+# Install Go
+# We cache the go source tarball in /var/cache/go-src
+RUN --mount=type=cache,target=/var/cache/go-src \
+    SHA256=$(sha256sum /var/cache/go-src/go.tar.gz | cut -d ' ' -f 1); \
+    if [ "${SHA256:-}" != "5e8c5a74fe6470dd7e055a461acda8bb4050ead8c2df70f227e3ff7d8eb7eeb6" ]; then \
+      curl --fail --silent --show-error --location -o /var/cache/go-src/go.tar.gz https://go.dev/dl/go1.19.2.linux-amd64.tar.gz; \
```
MarkusZoppelt, line 28: Maybe we can get the latest go release instead?

MarkusZoppelt: This would mean that we cannot easily check the SHA356 hash. Maybe there is some other clever way?

It would be nice if we do not have to update the Dockerfile every time there is a newer Go version available.

adombeck: I couldn't find a URL to download the latest Go. Also, I don't think we have to update the Go version every time there is a new Go release - only if we actually use new Go features. How about we revisit automating this when it happens? I could add comment for that.

adombeck: but we have to use an old libc, else we get libc incompatibility errors when the installer is used on a system which has an older libc

MarkusZoppelt: Gotcha. Then this is ok imho.

MarkusZoppelt (approved):

docker/cifuzz-builder/Dockerfile
adombeck: I'll try to use LLVM's own Debian package repo (https://apt.llvm.org/)

adombeck: ah I got confused, we're talking about installing Go, not LLVM

adombeck: I tried to install Go 1.19 from the Ubuntu kinetic sources but that has a dependency on the libc from Ubuntu kinetic so it's not an option. Will keep it as is and add a comment.

adombeck: > Will keep it as is and add a comment.

Done

jochil (approved):
nice caching

adombeck:
> nice caching

thanks :)

PR #453, opened 2022-11-18
Makefile: Fix make exiting with 0 when commands fail

adombeck:
Using the .ONESHELL target has the effect that all targets in the Makefile pass with exit code 0 as long as the last command of that target exits with 0. For example this target:

    fail:
        @false
        @echo success

succeeds when .ONESHELL appears anywhere in the Makefile:

    $ make fail
    success
    $ echo $?
    0

### DIFF
--- a/docker/cifuzz-builder/Dockerfile
+++ b/docker/cifuzz-builder/Dockerfile
@@ -21,7 +21,10 @@ ENV CC=/usr/lib/llvm-12/bin/clang
 ENV CXX=/usr/lib/llvm-12/bin/clang++
 
 # Install Go
-# We cache the go source tarball in /var/cache/go-src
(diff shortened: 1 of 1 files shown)


adombeck:
> Looks like this has already been committed / merged via https://github.com/CodeIntelligenceTesting/cifuzz/pull/452 (https://github.com/CodeIntelligenceTesting/cifuzz/commit/d33aebcf4d34f6e1aaaa8b982046bd686b6f2f11)

Oops, you're right.

> I believe this can be closed.

Did you or @jochil review it as part of #452?

MarkusZoppelt:
> Did you or @jochil review it as part of #452?

I read the commit message in the other PR and it made sense, so yes LGTM.

PR #455, opened 2022-11-21
docs(coverage): add bazel section to coverage guide

mlsandnerCI:


docs/Coverage-ide-integrations.md
```diff
@@ -11,9 +13,41 @@ See the [CLion documentation on presets
...
+This will run the fuzz test with the finding inputs from the `*_inputs` directory
+and produce a coverage report which can be viewed in your IDE **if the fuzz test
+passes**. If the fuzz test fails, the bugs must be fixed before a coverage report 
+can be produced.
+
+![Bazel fuzz test in IntelliJ](assets/coverage_intellij_bazel.gif)
 
 ## VS Code
+
+### CMake & Bazel
```
adombeck, line 50: This section does not apply to Bazel because we currently don't have a working VS Code task for Bazel and I'm not sure that we can create one which works with Bazel targets. We should document how to create an lcov report via the command-line instead (`cifuzz coverage --format lcov --output lcov.info <target>`) and that the Coverage Gutters extension automatically uses the `lcov.info` report.

mlsandnerCI: I rewrote that section. I couldn't test the process because generating a coverage report for the bazel example right now results in an empty lcov.info file for me so I wasn't sure if there have to be other steps after creating the report for Coverage Gutters to pick it up or if it is completely automatic.

adombeck (approved):

PR #456, opened 2022-11-21
build/bazel: Fix UBSan C++ runtime not linked

adombeck:


fmeum (approved):

adombeck:
I added a test

fmeum (approved):
@adombeck Looks good, but shouldn't the last two commits be switched? It seems as if the fix is coming after instead of before or with the test.

fmeum:
@adombeck This is problematic with `git bisect` though. I agree that it is useful to verify that the test actually would have caught the issue and always do that locally, but would prefer the other order or a single squashed commit to keep the pipeline passing at every commit.

adombeck:
> @adombeck This is problematic with git bisect though. I agree that it is useful to verify that the test actually would have caught the issue and always do that locally, but would prefer the other order or a single squashed commit to keep the pipeline passing at every commit.

Sorry, was about to rebase but apparently I accidentally merged the branch already :/

fmeum:
> Sorry, was about to rebase but apparently I accidentally merged the branch already :/

Don't worry, it does not matter that much in the end. :-) We do have a fixup commit on main from an earlier PR merge, but I that's of course not a big deal as well.

adombeck:
> We do have a fixup commit on main from an earlier PR merge, but I that's of course not a big deal as well.

Yeah, I also saw that and was annoyed about it, but I think force-pushing to main would cause more trouble than it would solve

adombeck:
I wish it was possible to have a check in GitHub which avoids "WIP", "XXX" and "fixup" commits. I tried to add a test for that in CLI-676-check-wip-commits but there is no way to ignore this test for certain commits (sometimes we might actually do want commit messages with these strings in them).

fmeum:
Could we restrict the check to commits that start with these terms? I don't see a situation in which we would want to merge such commits.

PR #459, opened 2022-11-21
Test and improve pure C fuzz tests with `other` build system

fmeum:


fmeum:
@jochil Friendly ping :-)

fmeum:
@adombeck Could you review in @jochil's stead?

adombeck:
> The flag value doesn't differ from `FUZZ_TEST_CXXFLAGS`, but is more in

I think you mean here that the value doesn't differ from `FUZZ_TEST_CFLAGS`

adombeck (approved):

PR #460, opened 2022-11-21
tests: include bazel example in e2e_installer pipeline

mlsandnerCI:
I added the bazel example to the e2e installer pipeline and comments where we have to change the bazel version as soon as the stable version 6 is released. I will also create a ticket for this to keep track of it.


.github/actions/vm_setup/action.yml
```diff
@@ -48,6 +48,7 @@ runs:
           # use below opts to fix linking issues on macOS, e.g. that "assert.h" is missing, etc.
           echo "common --repo_env=BAZEL_CXXOPTS=-isysroot/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk" >> ~/.bazelrc
           echo 'common --repo_env="BAZEL_LINKOPTS=--sysroot=/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk -fuse-ld=lld"' >> ~/.bazelrc
+          # change bazel version to stable release as soon as version 6 is released
           echo "USE_BAZEL_VERSION=6.0.0rc1" >> $GITHUB_ENV
```
fmeum, line 52: How about using 8349c95fc98beb4008085942a67a57f0c4da074b here as well for consistency?

fmeum (approved):

.github/workflows/pipeline_e2e_installer.yml
```diff
@@ -52,7 +52,11 @@ jobs:
           echo "CC=$LLVM_PATH/bin/clang" >> $GITHUB_ENV
           echo "CXX=$LLVM_PATH/bin/clang++" >> $GITHUB_ENV
           echo "PATH=$PATH:$LLVM_PATH/bin" >> $GITHUB_ENV
-
+          echo "common --repo_env=BAZEL_CXXOPTS=-isysroot/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk" >> ~/.bazelrc
```
fmeum, line 55: We should also add a comment to these lines saying that we would ideally want to get rid of them, but gave up investigating why they are needed.

.github/actions/vm_setup/action.yml
mlsandnerCI: Done

.github/workflows/pipeline_e2e_installer.yml
mlsandnerCI: I added comments about this 👍

fmeum (approved):

.github/workflows/pipeline_e2e_installer.yml
```diff
@@ -52,7 +52,13 @@ jobs:
           echo "CC=$LLVM_PATH/bin/clang" >> $GITHUB_ENV
           echo "CXX=$LLVM_PATH/bin/clang++" >> $GITHUB_ENV
           echo "PATH=$PATH:$LLVM_PATH/bin" >> $GITHUB_ENV
-
+          # use below opts to fix linking issues on macOS, e.g. that "assert.h" is missing, etc.
```
fmeum, line 55:
```suggestion
          # use below opts to fix compilation issues on macOS, e.g. that "assert.h" is missing, etc.
```

fmeum: I missed this at first, header files are only used during compilation, not linking.

mlsandnerCI: Good catch 👍

PR #461, opened 2022-11-21
Fix local run of `go test ./...`

fmeum:


MarkusZoppelt (commented):
Changes look good to me but what exactly is the error that this PR fixes?

What do you mean by "fix local run"? AFAIK you mostly changed files for the integration test.

I think a short comment in the PR message would be helpful. Thanks!

fmeum:
> Changes look good to me but what exactly is the error that this PR fixes?
> 
> What do you mean by "fix local run"? AFAIK you mostly changed files for the integration test.
> 
> I think a short comment in the PR message would be helpful. Thanks!

Good point, I added the error I was seeing after the first commit to commit and PR message.

fmeum:
On second thought, we should probably just add a version check, even though almost anyone will use a wrapper.

fmeum:
On third thought, I am not aware of any gradle projects not using the wrapper and the first version that should work is still very old (~2018), so I think we are fine without a version check for now.

MarkusZoppelt (approved):

PR #463, opened 2022-11-22
Create pull_request_template.md

mlsandnerCI:
I think that including a PR template could improve our repository documentation and would help all contributors (inside CI and outside).


.github/pull_request_template.md
```diff
@@ -0,0 +1,15 @@
...
+<!--- describe your changes in detail -->
+
+### Motivation & Context
+<!--- why is this change required? what problem does it solve? -->
+<!--- if it fixes an open issue, please link to the issue here -->
+ 
+<!--- Mental Checklist
+Go over these following points and check that your PR is complete.
+- My code follows the code style and contributing guidelines of this project.
+- The title and commits are clean, concise and in the style of Conventional Commits.
```
jochil, line 13: Maybe we can mention that the PR title is used in the change log and therefor it should should be as descriptive as possible and also prefixed with the type of change (feature, bugfix, ...)

I like conventional commits but we agreed on using them is optional

mlsandnerCI: Good points 👍

MarkusZoppelt:
LGTM 👍🏼

PR #465, opened 2022-11-22
Unify coverage output

florianGla:
Use genhtml for html coverage generation for cmake and other build systems.


adombeck:
>  fix(integration-tests): update html coverage report tests 

Do I understand correctly that this commit fixes the tests were failing because of the previous commit? In that case, please squash the commit into the previous one to ensure that our tests pass on each commit, which is important to make `git bisect` useful.

florianGla:
Yes, you are right. I will change that.

.github/actions/vm_setup/action.yml
```diff
@@ -51,6 +51,8 @@ runs:
           echo 'common --repo_env=BAZEL_LINKOPTS=--sysroot=/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk:-fuse-ld=lld' >> ~/.bazelrc
           # change bazel version to stable release as soon as version 6 is released
           echo "USE_BAZEL_VERSION=8349c95fc98beb4008085942a67a57f0c4da074b" >> $GITHUB_ENV
+          # install lcov which is need for coverage generation
```
adombeck, line 54:
```suggestion
          # install lcov which is needed for coverage report generation
```

adombeck (approved):

PR #467, opened 2022-11-22
all: Use American spelling of "behavior" throughout

fmeum:
UBSan itself uses the AE spelling, both in its name and in output it generates. For consistency, we should do the same.


fmeum:
@adombeck Does this require any changes to the backend?

adombeck:
> @adombeck Do you think this requires any changes to the backend?

It will require changes in the core repo when we update cifuzz there

adombeck (approved):

PR #468, opened 2022-11-22
Emit inputs on non-fatal UBSan findings

fmeum:
We hook `__sanitizer_report_error_summary` to run code whenever any
sanitizer produces a finding and call the callback provided to
`__sanitizer_set_death_callback` by libFuzzer to dump the current input
if the finding is a UBSan finding.

This currently applies only to the "cmake" and "other" build system.


fmeum:
https://github.com/CodeIntelligenceTesting/cifuzz/actions/runs/3526997726 passed

integration-tests/cmake/cmake_test.go
```diff
@@ -91,43 +91,46 @@ func TestIntegration_CMake_InitCreateRunCoverageBundle(t *testing.T) {
...
 	// (unless we're running on Windows, in which case UBSan is not
-	// supported)
-	if runtime.GOOS != "windows" {
-		expectedOutputs := []*regexp.Regexp{
-			regexp.MustCompile(`^SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior`),
-		}
-		cifuzzRunner.Run(t, &shared.RunOptions{ExpectedOutputs: expectedOutputs})
-	}
-
+	// supported) and the heap buffer overflow.
```
adombeck, line 94:
```suggestion
	// supported) and the use-after-free.
```

integration-tests/other/other_test.go
```diff
@@ -41,14 +43,109 @@ func TestIntegration_Other_RunCoverage(t *testing.T) {
...
+	// Check that Minijail is used (if running on Linux, because Minijail
+	// is only supported on Linux)
+	if runtime.GOOS == "linux" {
+		expectedOutputs = append(expectedOutputs, regexp.MustCompile(`bin/minijail0`))
+	}
 
-	// Run the fuzz test with --recover-ubsan and verify that it now
-	// also finds the heap buffer overflow
-	expectedFinding = regexp.MustCompile(`heap buffer overflow in exploreMe`)
-	runFuzzer(t, cifuzz, dir, "my_fuzz_test", expectedFinding, "--recover-ubsan")
```
adombeck, line 51: The `runFuzzer` function is now unused and can be deleted

tools/cmake/cifuzz/src/dumper.c
```diff
@@ -0,0 +1,112 @@
...
+  ((void (*)(void (*)()))(real_sanitizer_set_death_callback))(callback);
+}
+
+/*
+ * Ensure that ASan's verify_interceptors check passes: It checks that puts and
+ * __sanitizer_report_error_summary are defined in the same object. Since we
+ * hook the latter, we also have to (trivially) hook the former.
+ * https://github.com/llvm/llvm-project/blob/f8a469fc572778d05b72f34a772082cf3abd3cda/compiler-rt/lib/sanitizer_common/sanitizer_mac.cpp#L987-L993
+ * Older versions of LLVM check for pthread_create instead:
+ * https://github.com/llvm/llvm-project/blob/abc51fac09593ec048b3b298fa274af823e0a22d/compiler-rt/lib/sanitizer_common/sanitizer_mac.cpp#L1061-L1067
```
adombeck, line 48: Thorough work!

tools/cmake/cifuzz/src/dumper.c
```diff
@@ -0,0 +1,112 @@
...
+ * function, which consists of a single call to an internal implementation of
+ * Printf.
+ */
+void __sanitizer_report_error_summary(const char *error_summary) {
+  _ZN11__sanitizer6PrintfEPKcz(error_summary);
+  if (sanitizer_death_callback == NULL) {
+    return;
+  }
+  /*
+   * Do not emit the input twice for ASan, which is always fatal.
```
adombeck, line 99: because we don't support a `--recover-asan` flag yet, right?

tools/cmake/cifuzz/src/dumper.c
```diff
@@ -0,0 +1,112 @@
...
+  sanitizer_death_callback = callback;
+  __real___sanitizer_set_death_callback(callback);
+}
+
+/* clang mangling applied to __sanitizer::Printf(const char *format, ...) */
+void _ZN11__sanitizer6PrintfEPKcz(const char *format, ...);
+
+/*
+ * If wrapped with --wrap, the __wrap_ version of this function is never called.
+ * It is not clear why. Instead, we inline the real implementation of this
```
adombeck, line 89: Might this be the reason for the `__wrap_` version not being called?

https://linux.die.net/man/1/ld:
> You may wish to provide a "__real_malloc" function as well, so that links without the --wrap option will succeed. **If you do this, you should not put the definition of "__real_malloc" in the same file as "__wrap_malloc"; if you do, the assembler may resolve the call before the linker has a chance to wrap it to "malloc".**

adombeck (approved):

adombeck:
Great work!

internal/build/other/other.go
```diff
@@ -208,7 +231,15 @@ func (b *Builder) setLibFuzzerEnv() error {
 	// Users should pass the environment variable FUZZ_TEST_LDFLAGS to
 	// the linker command building the fuzz test. For libfuzzer, we set
 	// it to "-fsanitize=fuzzer" to build a libfuzzer binary.
```
adombeck, line 233: Lets move this comment back to the `envutil.Setenv(b.env, "FUZZ_TEST_LDFLAGS",` line

adombeck: Ah nevermind, it makes sense here as well

adombeck:
The bazel build command still needs to be adapted to also support non-fatal UBSan findings, right?

tools/cmake/cifuzz/src/dumper.c
fmeum: Thanks goes out to the CI for this one :-)

tools/cmake/cifuzz/src/dumper.c
fmeum: Good find, but I think it's not relevant here: We only provide declarations of `__real_*` functions, not definitions.

tools/cmake/cifuzz/src/dumper.c
fmeum: Yes, I added a second TODO mentioning that.

fmeum:
Yes, I will adjust the Bazel command as soon as non-fatal UBSan support has been landed. We can hide the `dumper` in our `@cifuzz` dependency.

PR #472, opened 2022-11-23
bazel: Dump input of non-fatal UBSan findings

adombeck:
Based on CLI-664-bazel-support-ubsan


fmeum (approved):

fmeum:
@adombeck I added a new test and some additional fixes. Could you take another look?

adombeck:
> @adombeck I added a new test and some additional fixes. Could you take another look?

LGTM. Not entirely sure about the `local_repository` in the example, because users might want to use the example projects as the basis for their own projects, but the commented out `git_repository` should help in those cases.

fmeum:
@adombeck I set up https://github.com/CodeIntelligenceTesting/cifuzz-bazel and moved rather than copied the Bazel-only bits to it. I had to copy some files and we should investigate how we could automate that in the future.

adombeck:
Out of curiosity: I see that you renamed the file made available via bazel from `dumper.c` to `dumper.cpp` - does that make any difference?

fmeum:
> Out of curiosity: I see that you renamed the file made available via bazel from `dumper.c` to `dumper.cpp` - does that make any difference?

I did that already in the version of this PR that still had the `BUILD.bazel` file in it explicitly. The reason is subtle, but explained here: https://github.com/CodeIntelligenceTesting/cifuzz-bazel/blob/1ccb0bb7f27864626f668cca6d6e87776e6f87bd/BUILD.bazel#L27-L32

adombeck:
> Out of curiosity: I see that you renamed the file made available via bazel from dumper.c to dumper.cpp - does that make any difference?

Ah, I see now that it's explained in a comment in the `BUILD.bazel`?

adombeck:
LGTM

PR #474, opened 2022-11-23
BREAKING: build: Remove `--recover-ubsan`

fmeum:
### Motivation & Context
Now that we were able to have non-fatal UBSan findings with inputs, we no longer have to offer this flag. If users really want this behavior, they can always toggle it via `UBSAN_OPTIONS`. We are planning to add a more general `--keep-going` or `--continue-on-finding` flag.


adombeck:
Maybe we want to remove the `--recover-ubsan` flag with the same PR which introduces a `--keep-going` flag, else we have to find the places again where we have to pass the value of this flag and eventually set the correct environment variable for UBSan.

adombeck:
Also, with this PR, we get the default behavior of UBSan again, which is to not halt on errors, right? Do we already know if we also want that as the default behavior once we have a `--keep-going` flag, or do we want UBSan findings to be fatal again by default then? If the latter, it would be breaking change again, which would be another to reason to do both in one PR instead.

fmeum:
> Maybe we want to remove the `--recover-ubsan` flag with the same PR which introduces a `--keep-going` flag, else we have to find the places again where we have to pass the value of this flag and eventually set the correct environment variable for UBSan.

Couldn't we just base such a future PR on a revert of this commit? We will probably spend a bit more time discussing the UX and name of `--keep-going`, but might want to get rif of `--recover-ubsan` as such a change only becomes more breaking over time.

> Also, with this PR, we get the default behavior of UBSan again, which is to not halt on errors, right?

We already changed the default in https://github.com/CodeIntelligenceTesting/cifuzz/pull/468, this PR really only removes the flag.

adombeck (approved):

fmeum:
@jochil Would like to hear your opinion on merging this now before I do so. :-)

jochil:
I would also vote for removing it now. From what I got from CS, there is usually no reason to stop on ubsan findings so IMO it makes sense to keep the default behavior (continue on ubsan) even when the `--keep-going` flag gets introduced.
Maybe we will decide in the future that `--keep-going` is default on.... who knows

jochil (approved):

PR #475, opened 2022-11-24
ci: Move test and linting jobs from PR pipeline to new pipeline

adombeck:
The PR pipeline now really only runs on pull requests, which allows us to use variables that are only available in workflows triggered by pull requests, for example $GITHUB_BASE_REF and $GITHUB_HEAD_REF.


jochil:
If you rename/move the existing `pipeline_pr.yml` you have to update the branch protection rules and the README (the status is linked there).

adombeck:
> If you rename/move the existing pipeline_pr.yml you have to update the branch protection rules 

It looks like the branch protection rules uses job names and not branch names - since I did not change the job names, no changes should be necessary.

> and the README (the status is linked there).

Good catch, done

jochil:
> It looks like the branch protection rules use job names and not pipeline names - since I did not change the job names, no changes should be necessary.

Ok, just wanted to make sure we are aware of it

jochil (approved):

README.md
```diff
@@ -5,7 +5,7 @@
   <a href="https://github.com/CodeIntelligenceTesting/cifuzz/releases">
     <img src="https://img.shields.io/github/v/release/CodeIntelligenceTesting/cifuzz">
   </a>
-  <a href="https://github.com/CodeIntelligenceTesting/cifuzz/actions/workflows/pipeline_pr.yml">
+  <a href="https://github.com/CodeIntelligenceTesting/cifuzz/actions/workflows/pipeline_tests.yml">
     <img src="https://img.shields.io/github/workflow/status/CodeIntelligenceTesting/cifuzz/PR%20Pipeline?logo=github" />
```
jochil, line 9: WIP

adombeck: another good catch, fixed

PR #476, opened 2022-11-24
don't check for llvm-symbolizer when using jazzer

MarkusZoppelt:
We no longer get the fuzzer environment from the libfuzzer runner because Jazzer sets various options itself.

This removes `llvm-symbolizer` as a strict dependency when using cifuzz with Java projects.


pkg/runner/libfuzzer/libfuzzer_runner.go
```diff
@@ -373,9 +373,15 @@ func (r *Runner) RunLibfuzzerAndReport(ctx context.Context, args []string, env [
...
 
 func (r *Runner) FuzzerEnvironment() ([]string, error) {
-	env, err := fuzzer_runner.FuzzerEnvironment()
-	if err != nil {
-		return nil, err
+	var env []string
+	var err error
+
+	if !r.SupportJazzer {
+		// only check for llvm-symbolizer if we are not using jazzer
```
fmeum, line 380: Isn't the real reason that `jazzer_runner` is meant to check its dependencies itself? Singling out `llvm-symbolizer` seems confusing.

MarkusZoppelt: Yeah, you're right. I have changed the commit to build the environment in jazzer_runner itself. Can you please review again? Thanks!

fmeum: Sorry, I didn't want you to change the implementation, only the comment. Or maybe extract the dependency check. I think it's fine and actually beneficial to share the env setup as far as possible.

fmeum: On second thought... Maybe we can just delete this method? We don't need any of these flags for Jazzer as long as native libraries aren't involved and we don't support them yet in cifuzz.

fmeum: As soon as we do need this setup, we also need llvm-symbolizer and can again share the libfuzzer setup.

fmeum (approved):

PR #477, opened 2022-11-24
fileutil: Remove CanonicalPath and its usages

adombeck:
Resolving the path of the project directory and all files and directories in is causing problems, for example LCOV reports contain unresolved source file paths, which we can't make relative to the resolved project directory.

We found that the correct solution is to not resolve symlinks in paths. The reason we did that in the first place was to make files and directories from the host available in the minijail sandbox, which should not be a problem anymore since we now make the root filesystem of the host available as a read-only mount in the sandbox.

Ref: CLI-674


adombeck:
@florianGla Could you check if this solves the issue for you?

florianGla:
Yes, it works. Thank you

fmeum (approved):

fmeum:
Missed that the integration tests fail (due to the missing bindings?), but the change looks good...

adombeck:
> Missed that the integration tests fail (due to the missing bindings?), but the change looks good...

Taking a look..

florianGla:
On MacOS examples/other fails locally, because the path of the fuzz test binary is not absolute anymore.

adombeck:
> The failing minijail invocation in CI contains "-b" "/tmp/cifuzz-other-testdata-1781564592,/tmp/cifuzz-other-testdata-1781564592,1" "-b" "/tmp/cifuzz-other-testdata-1781564592/my_fuzz_test". We may have to deduplicate bindings?

The issue was not duplicate bindings, but that paths which were previously made absolute were not absolute anymore. That's fixed by replacing all calls to `fileutil.CanonicalPath` with `filepath.Abs`. 

But even then the test on Windows fails, because the temporary directory we create via `os.MkdirTemp` contains a FAT 8.3 filename (see https://github.com/CodeIntelligenceTesting/cifuzz/actions/runs/3567695061/jobs/5995693415#step:5:134 which has `RUNNER~1` in the path) but the compiler uses paths with full filenames (see https://github.com/CodeIntelligenceTesting/cifuzz/actions/runs/3567695061/jobs/5995693415#step:5:142). Therefore the [stack trace](https://github.com/CodeIntelligenceTesting/cifuzz/actions/runs/3567695061/jobs/5995693415#step:5:334) has full filenames while the project directory we try to make it relative to has a FAT 8.3 filename, so it appears as if the source file is not below the project directory (see https://github.com/CodeIntelligenceTesting/cifuzz/actions/runs/3567695061/jobs/5995693415#step:5:442), so it's treated by the parser as a third-party library.

I did not find a way to get the full filename from a FAT 8.3 filename, so this is currently blocked until we find a solution.

adombeck:
> I did not find a way to get the full filename from a FAT 8.3 filename, so this is currently blocked until we find a solution.

I pushed a commit which should to the same as the `filepath.EvalSyminks` function on Windows, except for resolving symlinks. Lets see if that works.

fmeum:
@adombeck I went with a different approach based on the assumption that temp dirs are the only common source of 8.3 paths. Please take a look.

fmeum:
@adombeck This may resolve CLI-674, which has as its root cause that we resolve versioned shared library symlinks before adding them to the archive. I will add a test case to this PR to verify that's true.

fmeum:
@adombeck I added the tests for versioned shared libraries.

@pnoltof Could you check whether CLI-674 is fixed by this PR?

adombeck:
> @adombeck I added the tests for versioned shared libraries.

LGTM

PR #478, opened 2022-11-24
fix: use local deps from maven directly

jochil:
For maven projects we manually added some directories  to the classpath based on some static values. As these directories are configurable, the right way is to fetch them from maven, as done in this change.


internal/build/maven/xml_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestParsTestParseXML_BuildDirs(t *testing.T) {
```
fmeum, line 11: The name is garbled.

fmeum (approved):
Thanks a lot, this is much cleaner!

internal/build/maven/xml_test.go
jochil: :see_no_evil: fixed

fmeum (approved):

florianGla (approved):

PR #480, opened 2022-11-24
ci: Fix Bazel E2E test expected findings

fmeum:
We now also find the UBSan finding with Bazel.


fmeum:
@jochil The source test is passing now, the installer test will pass once we release. Or do we also have to fix it before we release?

jochil (approved):

jochil:
> @jochil The source test is passing now, the installer test will pass once we release. Or do we also have to fix it before we release?

Not necessary, the E2E Installer test is not blocking the release. It is just an additional check afterwards

PR #482, opened 2022-11-25
docs: Update bazel replay options

adombeck:
* Build with UBSan instrumentation to also find undefined behavior when replaying inputs
* Build with debug symbols to have stack traces symbolized
* Build with -U_FORTIFY_SOURCE to avoid fortification wrapper code to abort (without stack traces) before sanitizers can catch the bug


examples/bazel/.bazelrc
```diff
@@ -2,10 +2,15 @@
 build --repo_env=CC=clang
 build --repo_env=CXX=clang++
 
-# --config=asan-replay
-build:asan-replay --@rules_fuzzing//fuzzing:cc_engine=@rules_fuzzing//fuzzing/engines:replay
-build:asan-replay --@rules_fuzzing//fuzzing:cc_engine_instrumentation=none
-build:asan-replay --@rules_fuzzing//fuzzing:cc_engine_sanitizer=asan
+# Replay cifuzz findings (C/C++ only)
+build:cifuzz-replay --@rules_fuzzing//fuzzing:cc_engine=@rules_fuzzing//fuzzing/engines:replay
+build:cifuzz-replay --@rules_fuzzing//fuzzing:cc_engine_instrumentation=none
```
fmeum, line 7: This is the default, we should be able to leave it out.

fmeum (approved):

examples/bazel/.bazelrc
adombeck: done

PR #483, opened 2022-11-28
bazel: Include the generated corpus in coverage runs

fmeum:
### Description
Making it so that Bazel coverage runs also use the generated corpus.

### Motivation & Context
This behavior is already the default for all other build systems and required for accurate coverage runs outside cifuzz.


internal/cmd/integrate/integrate.go
```diff
@@ -178,6 +177,11 @@ func setupGitIgnore(projectDir string) error {
 			"/CMakeUserPresets.json",
 		)
 	}
+	if buildSystem == config.BuildSystemBazel {
+		filesToIgnore = append(filesToIgnore, ".*_corpus/")
```
fmeum, line 181: I could see a case for using a more distinctive name, e.g. `.foobar_fuzz_test_cifuzz_corpus`, since we are ignoring it everywhere, not just at the top level.

adombeck: IMO changing the suffix to `_cifuzz_corpus` is a good idea

fmeum: Done

integration-tests/bazel/bazel_test.go
```diff
@@ -266,3 +272,71 @@ func testNoCIFuzz(t *testing.T, cifuzzRunner *shared.CIFuzzRunner) {
...
+	cifuzz := cifuzzRunner.CIFuzzPath
+	testdata := cifuzzRunner.DefaultWorkDir
+
+	// Temporarily rename the crashing inputs to verify that the following
+	// coverage command operates on the generated corpus only.
+	fuzzTestInputs := filepath.Join(testdata, "src", "parser", "parser_fuzz_test_inputs")
+	backupFuzzTestInputs := filepath.Join(testdata, "src", "parser", "parser_fuzz_test_inputs.backup")
+	err := os.Rename(fuzzTestInputs, backupFuzzTestInputs)
+	require.NoError(t, err)
+	defer os.Rename(backupFuzzTestInputs, fuzzTestInputs) //nolint:errcheck
```
adombeck, line 290: I'm not sure if this deferred call is executed when any of the `require` statements below fails, which could lead to other tests failing with spurious errors when the inputs directory is not restored. Using [`t.Cleanup`](https://pkg.go.dev/testing#T.Cleanup) instead should avoid that.

adombeck (approved):

PR #484, opened 2022-11-28
Ensure that Bazel always picks up values of `CC/CXX`

fmeum:
### Description
Use `--repo_env=CC=<value>` to force invalidation whenever the value changes.

This also allows us to use an unmodified environment for Bazel - all variables that aren't explicitly mentioned on the command-line won't have been picked up by Bazel anyway.

### Motivation & Context
With `--repo_env=CC`, Bazel caches the value of the `CC` environment variable at least until a server shutdown.


adombeck:
What about these occurrences? https://github.com/CodeIntelligenceTesting/cifuzz/blob/7dd25d7d655cd5548e6e8e483c55a489db07ba21/internal/cmd/coverage/bazel/bazel.go#L56-L57

adombeck:
For single commit PRs, I would prefer it if the PR description matches the git commit message, else I have to read and comprehend the same thing twice, just phrased differently.

fmeum:
@adombeck Thanks for the finding these remaining usages, I fixed them and also cleaned them up a bit more.

adombeck (approved):

PR #489, opened 2022-11-28
ci(e2e): fix min. requirements pipeline

jochil:
... by checking for two findings inside the bazel test, as it failed for the v0.10.0 release


jochil:
verified with this run: https://github.com/CodeIntelligenceTesting/cifuzz/actions/runs/3566580245

PR #490, opened 2022-11-28
Actually fix integration tests sometimes not exiting on keyboard interrupt

adombeck:
Placing the child process group in the foreground breaks when multiple child processes are started in parallel (which will then override each other as the foreground process group).

Letting the parent process handle the signal and terminate the child process group is the better solution. What was missing there is that the signal must be re-raised to allow other go routines running in parallel to also terminate their child processes.


fmeum (approved):
Very clean solution (and much simpler)!

PR #491, opened 2022-11-29
feat(coverage): implement --format=jacocoxml

mlsandnerCI:
### Description
`cifuzz coverage --format=jacocoxml <fuzz test>` will now only create a `jacoco.xml` and no html report.

### Motivation & Context
Enable the user to only create the files actually needed.

Edit: I could not find a way to change the file name of the `jacoco.xml` for maven only for gradle and decided to not include it so we have the same behavior for both and only specify the output directory.


internal/coverage/coverage.go
```diff
@@ -2,10 +2,14 @@ package coverage
 
 import "code-intelligence.com/cifuzz/internal/config"
 
+const OutputHTML = "html"
+const OutputLCOV = "lcov"
+const OutputJacocoXML = "jacoco.xml"
```
jochil, line 7: I think `jacocoxml` instead of `.xml` will make it harder to confuse this with a filename

internal/cmd/coverage/coverage.go
```diff
@@ -218,7 +221,7 @@ func (c *coverageCmd) run() error {
 			StdOut: c.OutOrStdout(),
 			StdErr: c.OutOrStderr(),
 		}
-		reportPath, err = gen.Generate()
+		reportPath, err = gen.Generate(c.opts.OutputFormat)
```
jochil, line 224: why using a parameter here instead of adding it to the opts?

internal/coverage/coverage.go
```diff
@@ -2,10 +2,14 @@ package coverage
 
 import "code-intelligence.com/cifuzz/internal/config"
 
+const OutputHTML = "html"
```
jochil, line 5: wouldn't it be more like FormatHTML... like the `--format` flag

internal/coverage/coverage.go
mlsandnerCI: I thought it looked a bit weird but I see that this could be confusing 👍

internal/cmd/coverage/coverage.go
mlsandnerCI: Good catch, I just didn't notice that it already gives the output format to the maven and gradle coverage generators.

PR #493, opened 2022-11-29
showcase CLI with gifs

MarkusZoppelt:
### Description
Add some gifs that showcase the CLI via [VHS](https://github.com/charmbracelet/vhs).
Tape files can be found in docs/assets/tapes/

### Motivation & Context
We want to bring back some of the main focus points of the project and display them at first glance in the README.

This PR choses the cmake example as a demo example.

Closes [CLI-641](https://code-intelligence.atlassian.net/browse/CLI-641?atlOrigin=eyJpIjoiMDg5NWE0YjlmNWNlNGE2ZjhhMWNlNTIwNWJhZmFhMjIiLCJwIjoiaiJ9)


MarkusZoppelt:
@jochil @fmeum What else are good examples to showcase?

fmeum (commented):
WIP

fmeum (changes requested):
(block automerge)

MarkusZoppelt:
> checking them into our main git repo will negatively affect clone times - in fact, I just did a `git fetch` that picked up this branch and noticed it taking longer than usual. Assuming that we will want to update these files going forward, maybe we should rather move them into a separate Git repository, use LFS or host them somewhere where we can reference them from the README?

Good thinking! We could create, e.g., CodeIntelligenceTesting/demo-gifs and store/update them there. However, this would infringe our "keep documentation as close to the code as possible" mantra. For gifs and images, we could make an exception.

@adombeck @jochil What do you think?

jochil:
Basically we are using just one gif right now, IMO we should remove all the unused ones and then we are fine. For 1.8MB I would prefer to not start using a another repo or LFS. If it is getting more and more, of course we should think about it again

MarkusZoppelt:
Ok, I removed the unused gifs from the commit/repo.

jochil (approved):

fmeum (approved):

PR #495, opened 2022-11-30
add note to RELEASE.md to check for `RELEASE_BLOCKER` tickets

MarkusZoppelt:
### Description
Add a note at the top of the RELEASE.md to remind checking for open `RELEASE_BLOCKER` tickets in Jira before releasing.

### Motivation & Context
It may be a good idea to mention this in the docs so that we do not publish releases with unfixed bugs.
Developers should check before every new release that there are no open tickets with the `RELEASE_BLOCKER` tag in Jira.


docs/RELEASE.md
```diff
@@ -1,5 +1,10 @@
 ## How to create a new cifuzz release
 
+___
+> **IMPORTANT:** Before publishing a new release, please make sure that there are no
+open tickets with the `RELEASE_BLOCKER` tag in Jira.
```
adombeck, line 5: Lets add a link to a Jira search for open tickets with this tag, to make the life of the release manager a bit easier

MarkusZoppelt: Good point! Added a link.

adombeck (approved):

PR #497, opened 2022-11-30
bazel: Show a helpful error when cifuzz-bazel commit is incompatible

fmeum:


internal/build/bazel/build.go
```diff
@@ -477,3 +489,48 @@ func PathFromLabel(label string, flags []string) (string, error) {
...
 }
+
+// Parses formatted bazel query --output=build output such as:
+//
+// git_repository(
+//  name = "cifuzz",
+//  remote = "https://github.com/CodeIntelligenceTesting/cifuzz-bazel",
+//  commit = "ccb0bb7f27864626f668cca6d6e87776e6f87bd",
+//)
+var cifuzzCommitRegex = regexp.MustCompile(`(?m)^\s*commit\s*=\s*"([^"]*)",$`)
```
adombeck, line 500: before v0.10.0, we used `branch = "bazel-support"`, so the regex should match that too

adombeck: To clarify: With the current status of this branch, a workspace I set up with v0.9.0 fails with this error:
```
❌ Failed to parse the definition of the "cifuzz" repository in the
WORKSPACE file, run 'cifuzz init' to see setup instructions.
bazel query output:
# /home/adrian/projects/bazel-fuzzing-example/WORKSPACE:19:15
git_repository(
  name = "cifuzz",
  remote = "https://github.com/CodeIntelligenceTesting/cifuzz",
  branch = "bazel-support",
  strip_prefix = "tools/cmake/cifuzz/include/cifuzz",
)
# Rule cifuzz instantiated at (most recent call last):
#   /home/adrian/projects/bazel-fuzzing-example/WORKSPACE:19:15 in <toplevel>
# Rule git_repository defined at (most recent call last):
#   /home/adrian/.cache/bazel/_bazel_adrian/2acd487a0e392d6bc28933c100067c18/external/bazel_tools/tools/build_defs/repo/git.bzl:199:33 in <toplevel>
```

I would prefer it to fail this error instead:
```
Please update the reference specified for the "cifuzz" repository in the
WORKSPACE file.
Required: commit = "b013aa0f90fe8ac60adfc6d9640a9cfa451dda9e"
Current: branch = "bazel-support"
```

adombeck:
> @jochil Do you know whether there is a way to get `cifuzz init` to show the setup instructions with an existing `cifuzz.yml`? This PR would ideally just point to `cifuzz init` instead of repeating the instructions.

AFAIK there is no way currently. Maybe a `--ignore-existing` flag or similar would be useful.

fmeum:
How about we move the setup instruction print further up so that it is always printed, even if the command later fails with an error due to the config already existing? That would be a flagless solution, but I'm not sure whether it is a good one.

adombeck:
> How about we move the setup instruction print further up so that it is always printed, even if the command later fails with an error due to the config already existing? That would be a flagless solution, but I'm not sure whether it is a good one.

I tried it and it doesn't look too bad IMO:

```
$ cifuzz init
Add the following to your WORKSPACE file:

    load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
    load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")

    http_archive(
        name = "rules_fuzzing",
        sha256 = "4beab98d88e4bf2d04da0412d413a1364f95e5eb88963e15e603bee1410fcedf",
        strip_prefix = "rules_fuzzing-ca617e846d0f92e00a903903b0554ea9142e1132",
        urls = ["https://github.com/CodeIntelligenceTesting/rules_fuzzing/archive/ca617e846d0f92e00a903903b0554ea9142e1132.tar.gz"],
    )

    load("@rules_fuzzing//fuzzing:repositories.bzl", "rules_fuzzing_dependencies")

    rules_fuzzing_dependencies()

    load("@rules_fuzzing//fuzzing:init.bzl", "rules_fuzzing_init")

    rules_fuzzing_init()

    git_repository(
        name = "cifuzz",
        commit = "b013aa0f90fe8ac60adfc6d9640a9cfa451dda9e",
        remote = "https://github.com/CodeIntelligenceTesting/cifuzz-bazel",
    )

⚠️ Config already exists in /home/adrian/cifuzz/examples/bazel/cifuzz.yaml
❌ Failed to create config
```

@jochil WDYT?

internal/build/bazel/build.go
fmeum: Done, PTAL.

adombeck: Tested it and it works

adombeck (approved):

PR #502, opened 2022-12-02
cmd/remote-run: Support hostname as --server value

adombeck:
Users were having trouble using the --server flag correctly, trying to specify just the hostname or a URL without a scheme. To improve UX, the --server flag now supports URLs without scheme, in which case "https://" is assumed. 

This mimics the behavior of `docker login`.


internal/cmd/remote-run/remote_run.go
```diff
@@ -146,6 +147,17 @@ https://github.com/CodeIntelligenceTesting/cifuzz/issues`, system)
...
 			opts.Stderr = cmd.ErrOrStderr()
 
+			// Check if the server option is a valid URL
+			err = validateURL(opts.Server)
+			if err != nil {
+				// See if prefixing https:// makes it a valid URL
+				opts.Server = "https://" + opts.Server
+				err = validateURL(opts.Server)
+				if err != nil {
+					log.Error(err, fmt.Sprintf("server %q is not a valid URL", err))
```
fmeum, line 157: If `opts.Server` is `htp://foo.bar`, what would this print? If it's `https://htp://foo.bar`, then that would certainly not be ideal.

I'm also not sure whether `log.Error(err, fmt.Sprintf("...%s...", err))` doesn't end up duplicating parts of the error message.

adombeck: > fmt.Sprintf("server %q is not a valid URL", err)

oops, `err` is wrong here, I'll fix that

adombeck: > oops, err is wrong here, I'll fix that

done

> If opts.Server is htp://foo.bar, what would this print? If it's https://htp://foo.bar, then that would certainly not be ideal.

`--server htp://foo.bar` will now not fail and ask for an API token for `https://htp://foo.bar`. I agree that's not ideal but it's hard to do better with Go's URL parsing when we want to allow `foo.bar:443` but fail on `htp://foo.bar`. Note that Docker also doesn't handle this, it tries to log in at `https://htp/v2`.

fmeum: Ah, I was only worried about the error message: If the user types X, then the error should be `X is not valid`, not `https://X is not valid`. As long as that is ensured, I'm fine with this behavior.

fmeum (approved):

PR #503, opened 2022-12-02
implement `cifuzz login`

MarkusZoppelt:
### Motivation & Context
We want users to authenticate against a SaaS backend in order to sync local findings.
This allows for more advanced features on top of what the CLI provides.

The goal of this PR is to implement ~~an `auth`~~ a `login` command that enables
authentication with a backend, such as https://app.code-intelligence.com.

### Description
Implement `cifuzz login` to allow login.

### Tasks

- [x] add `login` command
- [x] move api specific code from remote_run to internal/api


internal/api/api.go
```diff
@@ -0,0 +1,274 @@
...
+func ValidateURL(s string) error {
+	u, err := url.Parse(s)
+	if err != nil {
+		return errors.WithStack(err)
+	}
+	if u.Scheme != "http" && u.Scheme != "https" {
+		return errors.Errorf("unsupported protocol scheme %q", u.Scheme)
+	}
+	return nil
+}
```
adombeck, line 274: This function is unused. It was copied from the `remote_run` package and the function defined there is still used.

internal/api/api.go
```diff
@@ -0,0 +1,274 @@
...
+	DisplayName           string `json:"display_name"`
+	OwnerOrganizationName string `json:"owner_organization_name"`
+}
+
+type Artifact struct {
+	DisplayName  string `json:"display-name"`
+	ResourceName string `json:"resource-name"`
+}
+
+func (api *Client) UploadBundle(path string, projectName string, token string) (*Artifact, error) {
```
adombeck, line 42: > func (api *Client)

This package uses both `api` and `client` as receiver names for the `Client` struct

internal/cmd/remote-run/remote_run.go
```diff
@@ -166,13 +164,10 @@ https://github.com/CodeIntelligenceTesting/cifuzz/issues`, system)
 		cmdutils.AddProjectDirFlag,
 		cmdutils.AddSeedCorpusFlag,
 		cmdutils.AddTimeoutFlag,
+		cmdutils.AddServerFlag,
+		cmdutils.AddProjectFlag,
```
adombeck, line 168: This list is sorted alphabetically, lets keep it that way

internal/cmdutils/pflag.go
```diff
@@ -187,3 +186,19 @@ func AddUseSandboxFlag(cmd *cobra.Command) func() {
...
+
+func AddProjectFlag(cmd *cobra.Command) func() {
+	// TODO: Make the project name more accessible in the web app (currently
+	//       it's only shown in the URL)
+	cmd.Flags().StringP("project", "p", "", `The name of the CI Fuzz project you want to start a fuzzing run for,
+e.g. "my-project-c170bc17".`)
+	return func() {
+		ViperMustBindPFlag("project", cmd.Flags().Lookup("project"))
+	}
+}
```
adombeck, line 204: These functions are also sorted alphabetically

internal/cmd/auth/auth.go
```diff
@@ -0,0 +1,33 @@
...
+
+	authLoginCmd "code-intelligence.com/cifuzz/internal/cmd/auth/login"
+	"code-intelligence.com/cifuzz/internal/cmdutils"
+)
+
+func New() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "auth <command>",
+		Short: "Authenticate with a remote backend to manage findings",
+		Long: `This command is used to authenticate with a remote backend to manage findings.
```
adombeck, line 14: We currently use the term `remote fuzzing server` in other user-facing strings in the CLI. Lets use that consistently unless we decide that we should use a different name (in which case we should change all occurrences).

internal/cmd/auth/auth.go
```diff
@@ -0,0 +1,33 @@
...
+After authenticating, users can sync local findings with a remote backend and
```
adombeck, line 15: IMO the term "syncing" implies that local and remote will have the same state afterwards, which is not really the case, because we only ever upload findings, never download findings, right? So "upload" would be more precise.

internal/cmd/auth/auth.go
```diff
@@ -0,0 +1,33 @@
...
+use advanced features on top of the features that the CLI provides.
+`,
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			// Bind viper keys to flags. We can't do this in the New
+			// function, because that would re-bind viper keys which
+			// were bound to the flags of other commands before.
+			return nil
```
adombeck, line 22: We can delete the `PreRunE` since we don't do anything here

internal/cmd/auth/auth.go
```diff
@@ -0,0 +1,33 @@
...
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return cmd.Help()
+		},
```
adombeck, line 26: This is the default behavior, so we can also delete the `RunE`.

internal/cmd/auth/auth.go
```diff
@@ -0,0 +1,33 @@
...
+	}
+
+	cmd.AddCommand(authLoginCmd.New())
```
adombeck, line 29: Is it the plan to have other subcommands beside `login`? Which ones? Unless we plan to add more subcommands in a follow-up task, I would prefer to get rid of the `auth` command and just have `cifuzz login`.

internal/cmd/auth/auth.go
```diff
@@ -0,0 +1,33 @@
...
+		Use:   "auth <command>",
+		Short: "Authenticate with a remote backend to manage findings",
+		Long: `This command is used to authenticate with a remote backend to manage findings.
+After authenticating, users can sync local findings with a remote backend and
+use advanced features on top of the features that the CLI provides.
+`,
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			// Bind viper keys to flags. We can't do this in the New
+			// function, because that would re-bind viper keys which
+			// were bound to the flags of other commands before.
```
adombeck, line 21: This comment belongs to https://github.com/CodeIntelligenceTesting/cifuzz/blob/63474e94daf3e8d9384e8fd684fcda8eb66875cd/internal/cmd/auth/login/login.go#L27-L27 where we actually bind flags

internal/cmd/auth/login/login.go
```diff
@@ -0,0 +1,95 @@
...
+			return nil
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			server := cmd.PersistentFlags().Lookup("server").Value.String()
+
+			// check if server URL is valid, if not try to add https://
+			_, err := url.ParseRequestURI(server)
+			if err != nil {
+				server = fmt.Sprintf("https://%s", server)
+			}
```
adombeck, line 37: The check in the `remote-run` command handles more cases: https://github.com/CodeIntelligenceTesting/cifuzz/blob/63474e94daf3e8d9384e8fd684fcda8eb66875cd/internal/cmd/remote-run/remote_run.go#L134-L143

internal/cmd/auth/login/login.go
```diff
@@ -0,0 +1,95 @@
...
+		msg := fmt.Sprintf(`Enter an API access token and press Enter. You can generate a token for
+your account at %s/dashboard/settings/account.`+"\n", server)
+		token, err := dialog.ReadSecret(msg, os.Stdin)
+		if err != nil {
+			return err
+		}
+
+		// Try to authenticate with the access token
+		if !isTokenValid(server, token) {
+			return errors.New("failed to authenticate with the provided API access token")
```
adombeck, line 72: This will cause hard to debug errors if we don't print the error which the `ListProjects` call failed with.

pkg/dialog/dialog.go
```diff
@@ -48,3 +52,13 @@ func MultiSelect(message string, items map[string]string) ([]string, error) {
 
 	return result, nil
 }
+
+func ReadSecret(message string, file *os.File) (string, error) {
+	log.Info(message)
+	log.Warnf("No characters will be displayed.")
```
adombeck, line 58: A warning seems a bit too much for this. Also, present tense seems more appropriate. How about this?

```suggestion
	log.Info("Note: Entered characters are not displayed")
```

adombeck: Also, I prefer the UX of `gh`, which prints `*` characters instead. It seems like none of the packages we currently import provide this functionality, so I suggest we add a `TODO` and keep the current behavior for now.

fmeum: Given that we are doing the right thing (not showing the characters), maybe we don't even have to print this message? The behavior is pretty common in CLI applications.

adombeck: Would work for me

internal/cmd/auth/login/login.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+func isTokenValid(server string, token string) bool {
+	client := api.Client{Server: server}
+	// TOOD: Change this to use another check without querying projects
+	_, err := client.ListProjects(token)
+	return err == nil
+}
+
+func Login(server string) error {
```
adombeck, line 55: We should be able to share the code from the `remote_run` package for the same purpose. There we also have some additional functionality which IMO would also be useful here:
* Reading the token from stdin when stdin is not a TTY
* Supporting `--interactive=false` (and `interactive: false` in the `cifuzz.yaml`)
* Not using interactive input when stdin or stdout is not a TTY (the `login` command currently fails with `inappropriate ioctl for device` when stdin is not a TTY, to reproduce: `true | cifuzz auth login`)

internal/api/api.go
MarkusZoppelt: Done. I renamed the struct to `APIClient` and all receiver names to `client`.

internal/cmd/remote-run/remote_run.go
MarkusZoppelt: Done

internal/cmdutils/pflag.go
MarkusZoppelt: Done

internal/cmd/auth/auth.go
MarkusZoppelt: Done

internal/cmd/auth/auth.go
MarkusZoppelt: Done. Renamed it to "upload".

internal/cmd/auth/auth.go
MarkusZoppelt: Done

internal/cmd/auth/auth.go
MarkusZoppelt: Done

internal/cmd/auth/login/login.go
MarkusZoppelt: I changed the `isTokenValid` to print the error when it is not nil. Is this ok?
We should update the function nonetheless once another API call for testing auth with the backend is available. Listing Projects works, but it's not ideal..

internal/cmd/auth/login/login.go
```diff
@@ -0,0 +1,95 @@
...
+		log.Successf("Successfully authenticated with %s", server)
+	} else {
+		// Token might be invalid, so try to authenticate with it
+		if !isTokenValid(server, token) {
+			log.Warnf(`cifuzz detected an API access token, but failed to authenticate with it.
+This might happen if the token has been revoked.
+Please remove the token from %s and try again.`,
+				access_tokens.GetTokenFilePath())
+			return errors.New("failed to authenticate with the provided API access token")
+		}
```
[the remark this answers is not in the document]
MarkusZoppelt: See other comment.

pkg/dialog/dialog.go
MarkusZoppelt: Removed the Warn/Info and added a Todo comment. I will also create a ticket.

internal/cmd/auth/login/login.go
adombeck: I now that `ListProjects` already handles request errors by printing the error and returning a silent error. That was fine when it was an internal function of the `remote_run` package, but now that it's an exported function of the `api` package, it should leave it to the caller to handle errors. 

Ideally, we return a custom error which allows the caller to inspect the status code, which allows us to handle actual authentication errors differently from other errors in the `isTokenValid` function.

internal/cmd/auth/login/login.go
MarkusZoppelt: Thx!

internal/api/api.go
MarkusZoppelt: Fixed

internal/cmd/auth/auth.go
MarkusZoppelt: Done. I removed `auth`. There is only `cifuzz login` now. If users are already logged in, we display the location of the access_token as an info:

```
> cifuzz login --server app.staging.code-intelligence.com
✅ You are already logged in.
Your API access token is stored in /Users/mz/.config/cifuzz/access_tokens.json
```

internal/cmd/auth/auth.go
```diff
@@ -0,0 +1,33 @@
...
+	"code-intelligence.com/cifuzz/internal/cmdutils"
+)
+
+func New() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "auth <command>",
+		Short: "Authenticate with a remote backend to manage findings",
+		Long: `This command is used to authenticate with a remote backend to manage findings.
+After authenticating, users can sync local findings with a remote backend and
+use advanced features on top of the features that the CLI provides.
```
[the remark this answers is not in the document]
jochil: IMO we can name and link our product directly. Something like

> This command is used to authenticate with a CI Fuzz Server instance. Learn more about it at: https://www.code-intelligence.com/

internal/cmd/auth/login/login.go
```diff
@@ -0,0 +1,95 @@
...
+		Use:     "login",
+		Short:   "Login to a remote backend",
+		Long:    `This command is used to authenticate with a remote backend to manage findings.`,
+		Example: "$ cifuzz auth login",
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			bindFlags()
+			return nil
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			server := cmd.PersistentFlags().Lookup("server").Value.String()
```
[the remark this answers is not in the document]
MarkusZoppelt: Done

internal/cmd/auth/login/login.go
MarkusZoppelt: Good point. Done.

internal/cmd/auth/login/login.go
MarkusZoppelt: Done

internal/cmd/auth/login/login.go
MarkusZoppelt: Done. There is now a file `cmdutils/login/login.go` with the shared code.
We needed to create a new package (login), otherwise there would have been cyclic imports.

jochil (dismissed):
lgtm

adombeck:
Taking over from @MarkusZoppelt (as discussed) because he is now on vacation.

internal/cmd/auth/auth.go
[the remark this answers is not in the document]
adombeck: I pushed a commit which uses the term "CI Fuzz Server instance" consistently in all user-visible strings.

adombeck:
I would prefer not opening the browser automatically, that feels a bit too invasive to me. The GitHub CLI for example also only prints the URL and leaves it to the user to open it in their browser (unless the "Login with a web browser" option is chosen, which is a different workflow than ours and still waits for the user to press Enter before opening the browser):

```
gh auth login 
? What account do you want to log into? GitHub.com
? What is your preferred protocol for Git operations? HTTPS
? Authenticate Git with your GitHub credentials? Yes
? How would you like to authenticate GitHub CLI? Paste an authentication token
Tip: you can generate a Personal Access Token here https://github.com/settings/tokens
The minimum required scopes are 'repo', 'read:org', 'workflow'.
? Paste your authentication token: 
```

@jochil WDYT?

jochil:
> I would prefer not opening the browser automatically, that feels a bit too invasive to me. The GitHub CLI for example also only prints the URL and leaves it to the user to open it in their browser (unless the "Login with a web browser" option is chosen, which is a different workflow than ours and still waits for the user to press Enter before opening the browser):
> 
> @jochil WDYT?

It surprised me too, for the first time I tested it... IMO we should at least wait for `Enter` too, so it should not surprise the user and is still convenient. 
And if I got it correctly it is bind to the `--interactive` flag. So user have a way to disable it

adombeck:
> It surprised me too, for the first time I tested it... IMO we should at least wait for `Enter` too, so it should not surprise the user and is still convenient. 

We are asking the user to enter the token, so we can't wait for Enter to open the browser.

> And if I got it correctly it is bind to the `--interactive` flag. So user have a way to disable it

Opening the browser is independent of the `--interactive` flag, both here and in the `coverage` command.

jochil (commented):
why got the `cmdutils/login` package removed? Now there is duplicated code again in the commands... re-implementing the same dialog in `remote-run` and `login`.

jochil:
> > It surprised me too, for the first time I tested it... IMO we should at least wait for `Enter` too, so it should not surprise the user and is still convenient.
> 
> We are asking the user to enter the token, so we can't wait for Enter to open the browser.

What do you think about a [yes|no] dialog

> 
> > And if I got it correctly it is bind to the `--interactive` flag. So user have a way to disable it
> 
> Opening the browser is independent of the `--interactive` flag, both here and in the `coverage` command.

Seems like the code changed... in the `login` package it was inside an if branch with a `opts.Interactive` check

adombeck:
> why got the cmdutils/login package removed? Now there is duplicated code again in the commands... re-implementing the same dialog in remote-run and login.

Which duplicated code do you mean? With b8590db599d2659f58ae21dc0ce2f38805c87d0f the remote-run command doesn't have a login dialog anymore. It tries to read the API token from the environment variable or the config file and if it doesn't find any it tells the user to set the environment variable or use `cifuzz login`.

adombeck:
> > > It surprised me too, for the first time I tested it... IMO we should at least wait for `Enter` too, so it should not surprise the user and is still convenient.
> > 
> > 
> > We are asking the user to enter the token, so we can't wait for Enter to open the browser.
> 
> What do you think about a [yes|no] dialog

I implemented something. Please take a look.

> > > And if I got it correctly it is bind to the `--interactive` flag. So user have a way to disable it
> > 
> > 
> > Opening the browser is independent of the `--interactive` flag, both here and in the `coverage` command.
> 
> Seems like the code changed... in the `login` package it was inside an if branch with a `opts.Interactive` check

It is, but then the whole dialog is skipped. There is no flag to only avoid that the browser is opened.

jochil:
> > why got the cmdutils/login package removed? Now there is duplicated code again in the commands... re-implementing the same dialog in remote-run and login.
> 
> Which duplicated code do you mean? With [b8590db](https://github.com/CodeIntelligenceTesting/cifuzz/commit/b8590db599d2659f58ae21dc0ce2f38805c87d0f) the remote-run command doesn't have a login dialog anymore. It tries to read the API token from the environment variable or the config file and if it doesn't find any it tells the user to set the environment variable or use `cifuzz login`.

Duplicated code was not quite right, more like duplicated logic. Now the two commands are both checking tokens and respond to the user, but in different ways. IIRC one outcome of talking about the login integration was, that all commands that needs (or want to use) an authentication should switch to the same login dialog when one is missing... this was the reason why Markus and I implemented it that way. So basically this (if not logged in already):

`cifuzz login` -> show login dialog
`cifuzz remote-run` -> show login dialog -> bundle -> upload
`cifuzz run` -> build -> run -> show login dialog -> sync findings
`cifuzz sync` -> show login dialog -> sync findings

jochil:
> > What do you think about a [yes|no] dialog
> 
> I implemented something. Please take a look.

This is nice... I think this can work well :+1: 

> It is, but then the whole dialog is skipped. There is no flag to only avoid that the browser is opened.

ok, I also think we don't need a specific one... was just thinking if this already covers our use case here

adombeck:
>  IIRC one outcome of talking about the login integration was, that all commands that needs (or want to use) an authentication should switch to the same login dialog when one is missing... this was the reason why @MarkusZoppelt and I implemented it that way. So basically this (if not logged in already):
> 
> `cifuzz login` -> show login dialog `cifuzz remote-run` -> show login dialog -> bundle -> upload `cifuzz run` -> build -> run -> show login dialog -> sync findings `cifuzz sync` -> show login dialog -> sync findings

I changed that because with this workflow I didn't find a way to handle the `$CIFUZZ_API_TOKEN` in a way that's not surprising. IMO `cifuzz login` clearly should have the outcome that a token is stored in the config file, so that subsequent cifuzz commands use that token. But if the token stored in `$CIFUZZ_API_TOKEN` is changed, that new token should be used, but it should not automatically update the token in the config file. The best solution I found was to handle `$CIFUZZ_API_TOKEN` as an alternative to the config file: When `$CIFUZZ_API_TOKEN` is set, `cifuzz remote-run` uses that, else it uses the config file. If neither contains a valid token, we tell the user to set `$CIFUZZ_API_TOKEN` or run `cifuzz login` to store a token in the config file.

adombeck:
@jochil: As discussed, I added a commit which allows logging in in the `remote-run` command when no token was found.

jochil:
> @jochil: As discussed, I added a commit which allows logging in in the `remote-run` command when no token was found.

I like it very much... feels fluent :+1:

jochil (approved):

PR #504, opened 2022-12-02
Fix dependency check failing when CC is set and clang is not in $PATH

adombeck:
Fixes the issue described in the title but also includes some refactoring, see commit messages for explanations.


adombeck:
@jochil: This PR touches a lot of files so I would be grateful if you could review this soon to avoid conflicts

internal/cmd/create/create.go
```diff
@@ -207,13 +202,19 @@ and $FUZZ_TEST_LDFLAGS to the linker.`)
...
 	case config.BuildSystemOther:
 		deps = []dependencies.Key{dependencies.CLANG}
 	}
-	return dependencies.Check(deps, dependencies.CMakeDeps, runfiles.Finder)
+	err := dependencies.Check(deps, dependencies.CMakeDeps, runfiles.Finder)
+	if err != nil {
+		// we ignore errors here because this command has no actual
+		// dependencies and we just want to give recommendations
+		// instead of letting the command fail
+		log.Debug(err)
```
jochil, line 217: This is actually a change of the current behavior of `create`, or? Was this on intent?

adombeck: It doesn't change the behavior of the command. The `Check` function used to only return an error if it was passed a dependency which doesn't have a definition. Now it doesn't return an error in that case but panics (because that's a programming error). Instead, it now returns an error if there are any unmet dependencies, which was and still is ignored by the `create` command.

pkg/dependencies/definitions.go
```diff
@@ -17,6 +22,26 @@ var deps = Dependencies{
...
+					clang = cxx
+				}
+			}
+
+			if clang != "" {
+				_, err := exec.LookPath(clang)
+				if err == nil {
+					return true
+				}
+			}
```
jochil, line 43: I am not sure, but this code might overlapping with: https://github.com/CodeIntelligenceTesting/cifuzz/blob/main/pkg/dependencies/version.go#L48
There we already have similar checks

adombeck: I saw that, but there we check the clang version and here we only check if clang is found in the path. I found no good way to reuse code.

PR #505, opened 2022-12-02
Forward all environment variables to the sandbox

fmeum:
Since we aren't using the sandbox for security and don't claim that we are using for hermeticity, we might as well pass all environment variables to the fuzz test. This can help improve compatibility with local dependencies or choices such as `JAVA_HOME`.


internal/cmd/run/run.go
```diff
@@ -479,10 +480,16 @@ func (c *runCmd) runFuzzTest(buildResult *build.Result) error {
...
+	env, err = envutil.Setenv(env, "NO_CIFUZZ", "1")
+	if err != nil {
+		return err
+	}
+
 	runnerOpts := &libfuzzer.RunnerOptions{
 		Dictionary:         c.opts.Dictionary,
 		EngineArgs:         c.opts.EngineArgs,
-		EnvVars:            []string{"NO_CIFUZZ=1"},
+		EnvVars:            env,
```
adombeck, line 492: Since we already made changes to the minijail package under the assumption that it's not used for security (e.g not preloading libminijailpreload.so and mounting the entire filesystem), we can also remove the functionality to only pass specified environment variables there (and in the process_wrapper), making the code simpler.

I'll add a commit for that.

The core repo still depends on minijail for security, so we'll have to deviate from our code there (but that's already the case).

adombeck: > Since we already made changes to the minijail package under the assumption that it's not used for security (e.g not preloading libminijailpreload.so and mounting the entire filesystem), we can also remove the functionality to only pass specified environment variables there (and in the process_wrapper), making the code simpler.
> 
> I'll add a commit for that.

On second thought, me might want to wait with that until we have a plan for how we want to continue with fuzzer-runner in the core repo.

adombeck (approved):

PR #512, opened 2022-12-05
Fix `make fmt` and `make fmt/check` exiting with 0 on failure

adombeck:


fmeum (approved):

fmeum:
These are the moments in which I seriously consider writing workflows in Go rather than Shell/Make...

PR #516, opened 2022-12-07
add Kotlin support for Gradle

MarkusZoppelt:
### Description
Add a Kotlin example to our code examples

### Tasks

- [X] add example src and test
- [x] update instructions in `internal/cmd/init/instructions/gradlekotlin`
- [x] add template file for kotlin in `pkg/stubs/fuzztest.ktl`
- [x] add tests


MarkusZoppelt:
> Merge branch 'main' into CLI-541-add-kotlin-support-for-gradle

Please try to avoid merge commits and instead rebase the branch on top of main.

internal/cmd/init/instructions/gradlekotlin
```diff
@@ -0,0 +1,15 @@
...
+
+
+We highly recommend using cifuzz with JUnit >=5 to ensure
+easy IDE integration. You can add it with the following
+dependency to your build.gradle.kts:
+
+testImplementation ("org.junit.jupiter", "junit-jupiter","5.9.0")
+
+For Jacoco coverage reports, please add the jacoco plugin to your build.gradle.kts.
+
```
adombeck, line 15: These instructions are currently not used. The `init` command should be updated to automatically detect if the project that's being initialized is a Kotlin project and print these instructions in that case.

adombeck (dismissed):
Thanks for working on this @wojtekCodeInt, this is very helpful!

You even wrote a test, that's great! It isn't passing yet though - the tests need to pass before we can merge this. You can run it locally via `go test code-intelligence.com/cifuzz/integration-tests/gradlekotlin`. 

A few notes:
* Please install our pre-push git hook which formats all files and runs a linter. See https://github.com/CodeIntelligenceTesting/cifuzz/tree/main/githooks for instructions on how to install it.
* I would prefer the directory name `kotlin-gradle` (or maybe `kotlin+gradle`) for the example project. We might also want to change the names of the other integration tests to `java-gradle` and `java-maven` (or `java+gradle` and `java+maven`).
* The commit messages should be a bit more specific and ideally be prefixed with the part of the code that's being changed, for example instead of "add instructions" use "cmd/init: Add instructions for Kotlin + Gradle".

wojtekCodeInt:
Can someone please help me with this?
```
Error: integration-tests/gradlekotlin/gradlekotlin_test.go:155:36: cannot use reportPath (variable of type string) as type io.Reader in argument to summary.ParseJacocoXML:
	string does not implement io.Reader (missing Read method)
```
	
This code was already in the Java gradle test.

florianGla:
> Can someone please help me with this?
> 
> ```
> Error: integration-tests/gradlekotlin/gradlekotlin_test.go:155:36: cannot use reportPath (variable of type string) as type io.Reader in argument to summary.ParseJacocoXML:
> 	string does not implement io.Reader (missing Read method)
> ```
> 
> This code was already in the Java gradle test.

I think you copied the gradle test, but I changed the ParseJacocoXML signature in the meantime. You now have to pass an object, which implements the io.Reader interface. That can be a file or a strings.Reader. Maybe have a look at the current version of the gradle integration test.

florianGla:
It would be nice if you could merge the changes from the last commit in the other commits and maybe use conventional commit messages (https://www.conventionalcommits.org/)

PR #517, opened 2022-12-07
Simplify install dir structure

florianGla:
This PR simplifies the directory structure of the cifuzz installation.


florianGla:
I am not sure if the second commit is necessary. It changes the folder permissions of the completions folder to 0755, which is consistent with all other folders in the installation directory.

florianGla:
Is it ok to use one general commit for all refactorings or do you prefer more specific commits and messages?

adombeck:
> tools/cmake/cifuzz/share/CMakePresets.json → examples/cmake/CMakeUserPresets.json

Was this on purpose? If we want to add a `CMakeUserPresets.json` to our CMake example. lets do so in a separate commit.

adombeck:
LGTM beside the one comment above.

We should create a follow-up ticket to also simplify the structure of paths like this: `~/.local/share/cifuzz/share/include/cifuzz/cifuzz/cifuzz.h`

florianGla:
> > tools/cmake/cifuzz/share/CMakePresets.json → examples/cmake/CMakeUserPresets.json
> 
> Was this on purpose? If we want to add a `CMakeUserPresets.json` to our CMake example, lets do so in a separate commit.

No, that was an accident. Good catch.

PR #518, opened 2022-12-07
Use reader interface for coverage parsers

florianGla:


adombeck (changes requested):
Please explain why you do this change in the commit message

adombeck:
Great explanation, thanks for adding it!

adombeck (approved):

PR #520, opened 2022-12-08
cmd/run: Support passing flags to cmake

florianGla:


internal/build/cmake/cmake.go
```diff
@@ -132,7 +133,11 @@ func (b *Builder) Configure() error {
 		cacheArgs = append(cacheArgs, "-DCMAKE_BUILD_RPATH_USE_ORIGIN:BOOL=ON")
 	}
 
-	cmd := exec.Command("cmake", append(cacheArgs, b.ProjectDir)...)
+	args := cacheArgs
+	args = append(args, b.Args...)
```
fmeum, line 137: CMake cache variables are cached in the build directory, so users may get unexpected results if they vary arguments. We can either:

1. hash the args and add the hash to the name of the build directory or
2. research how we can force an override of the cache variables.

fmeum (changes requested):

internal/build/cmake/cmake.go
florianGla: I implemented the first variant.

internal/build/cmake/cmake.go
```diff
@@ -90,16 +115,26 @@ func (b *Builder) BuildDir() string {
...
 		sanitizersSegment = "none"
 	}
+
+	var bytes []byte
+	for _, arg := range b.Args {
+		bytes = append(bytes, []byte(arg)...)
+	}
+	sum := sha1.Sum(bytes)
+	hash := fmt.Sprint(binary.BigEndian.Uint64(sum[:]))
+	buildDir := fmt.Sprintf("%s-%s", sanitizersSegment, hash)
```
fmeum, line 131: Let us append the hash only if `b.Args` isn't empty so that common use cases don't add it.

internal/build/cmake/cmake.go
```diff
@@ -90,16 +115,26 @@ func (b *Builder) BuildDir() string {
...
 	sanitizersSegment := strings.Join(b.Sanitizers, "+")
 	if sanitizersSegment == "" {
 		sanitizersSegment = "none"
 	}
+
+	var bytes []byte
+	for _, arg := range b.Args {
+		bytes = append(bytes, []byte(arg)...)
+	}
+	sum := sha1.Sum(bytes)
```
fmeum, line 129: Don't use SHA-1 (or MD5) for anything anymore. If security is a concern or performance is not, use SHA-256/512. If it's certain that security is no concern (such as in this case) and performance is important (it isn't in this case), use [xxHash](https://github.com/cespare/xxhash).

internal/build/cmake/cmake.go
```diff
@@ -90,16 +115,26 @@ func (b *Builder) BuildDir() string {
...
+	// Currently, this includes the fuzzing engine, the choice of sanitizers
+	// and additional user arguments
 	sanitizersSegment := strings.Join(b.Sanitizers, "+")
 	if sanitizersSegment == "" {
 		sanitizersSegment = "none"
 	}
+
+	var bytes []byte
+	for _, arg := range b.Args {
+		bytes = append(bytes, []byte(arg)...)
```
fmeum, line 127: This scheme has a subtle bug that is commonly made when hashing collections: `[]string{"foo", "bar"}` and `[]string{"foobar"}` will hash to the same value.

Instead, use `sha256.New()` to get an `io.Writer` and, for each arg, use `binary.Write` to first write the length of the argument followed by the bytes of the argument.

fmeum (changes requested):

fmeum (approved):

PR #521, opened 2022-12-12
Makefile: Fix `make fmt` not formatting

adombeck:


adombeck:
I had to push another commit because `make fmt/check` was also broken again. Doing more than just a simple command invocation in a Makefile target is really error prone..

PR #523, opened 2022-12-12
feat: source file resolver

mlsandnerCI:
### Motivation & Context
Last step to fix the `vscode` integration to run coverage commands without errors. 

### Description
With the flag `-r`/`--resolve` the user can now list a source file instead of a fuzz test and it will be automatically resolved into the fuzz test identifier. This process involved a bit of refactoring to avoid duplicating code.


internal/build/build.go
```diff
@@ -201,3 +185,28 @@ func ListJazzerFuzzTests(projectDir string) ([]string, error) {
 
 	return fuzzTests, nil
 }
+
+func ConstructJavaClassPathForFuzzTest(path, testDir string) (string, error) {
```
jochil, line 189: this file already contains a few jazzer specific functions. What do you think about moving them to a `jazzer.go` file in the same directory or even to a `jazzer_build` package a level deeper?

internal/build/build.go
```diff
@@ -201,3 +185,28 @@ func ListJazzerFuzzTests(projectDir string) ([]string, error) {
...
+	bytes, err := os.ReadFile(path)
+	if err != nil {
+		return "", err
```
jochil, line 192: We usually include the stack trace (`errors.WithStack(err)`) for errors that are not thrown by our own packages

internal/build/build.go
```diff
@@ -201,3 +185,28 @@ func ListJazzerFuzzTests(projectDir string) ([]string, error) {
...
+	}
+
+	var classPath string
+	match := jazzerFuzzTestRegex.MatchString(string(bytes))
+	if match {
+		classPath, err = filepath.Rel(testDir, path)
+		if err != nil {
```
jochil, line 199: same `errors.WithStack(err)` thing as above

internal/cmdutils/pflag.go
```diff
@@ -155,6 +155,14 @@ func AddProjectDirFlag(cmd *cobra.Command) func() {
 	}
 }
 
+func AddResolveSourceFileFlag(cmd *cobra.Command) func() {
+	cmd.Flags().BoolP("resolve", "r", false, "Resolve source file path to fuzz test name.\n"+
+		"The path has to be relative from the current working directory.")
```
jochil, line 160: why do we have to limit it to relative paths?
I would guess if it is an absolute path and we know the project dir it should be easy to convert it to a relative one. Depending on the tool/ide this flag is used it might be helpful to use it with absolute paths

jochil:
If I am not wrong the `bundle` and `remote-run` commands should also support the `--resolve` flag, or?

internal/cmdutils/pflag.go
```diff
@@ -155,6 +155,14 @@ func AddProjectDirFlag(cmd *cobra.Command) func() {
 	}
 }
 
+func AddResolveSourceFileFlag(cmd *cobra.Command) func() {
+	cmd.Flags().BoolP("resolve", "r", false, "Resolve source file path to fuzz test name.\n"+
```
jochil, line 159: Maybe it would be helpful for the user to point out that this affects the first argument. Something like:

> With this flag set, the argument of this command can be a path to a source file instead of the test identifier (eg. CMake Target or Java Class Name)

internal/cmdutils/resolve.go
```diff
@@ -0,0 +1,137 @@
...
+	log.Warn("No fuzz test found for the given source file.")
+	return "", err
+}
+
+func findAllCMakeLists(projectDir string) ([]string, error) {
+	var cmakeLists []string
+
+	err := filepath.WalkDir(projectDir, func(path string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
```
jochil, line 120: should be `errors.WithStack(err)`

jochil:
> > If I am not wrong the bundle and remote-run commands should also support the --resolve flag, or?
> 
> I will add it. Is it a problem that the user can only specify one fuzz test (the resolved source file) with this flag and not multiple?

this should not be a problem. For `run` and `coverage` we are not supporting multiple fuzz tests, as this would require parallel execution and coverage merging. For `bundle` and `remote-run` it should be just `one` or `all`
@adombeck Am I right?

adombeck:
> this should not be a problem. For run and coverage we are not supporting multiple fuzz tests, as this would require parallel execution and coverage merging. For bundle and remote-run it should be just one or all

`bundle` and `remote-run` can take multiple fuzz tests as arguments

> If I am not wrong the bundle and remote-run commands should also support the --resolve flag, or?

Is there a use case for that? If I understand correctly, we're only adding this flag to support running fuzz tests via a VS Code task. Is there a use case for using the `bundle` or `remote-run` command via VS Code tasks?

jochil:
> > this should not be a problem. For run and coverage we are not supporting multiple fuzz tests, as this would require parallel execution and coverage merging. For bundle and remote-run it should be just one or all
> 
> `bundle` and `remote-run` can take multiple fuzz tests as arguments

then just resolving all fuzz test arguments, or?

> > If I am not wrong the bundle and remote-run commands should also support the --resolve flag, or?
> 
> Is there a use case for that? If I understand correctly, we're only adding this flag to support running fuzz tests via a VS Code task. Is there a use case for using the `bundle` or `remote-run` command via VS Code tasks?

This is the primary use case, but IMO it would be nice if user can use it too... sometimes you know the file but not the build target you want to invoke.
Also it is handy when the auto completion for fuzz targets is not available

internal/cmdutils/pflag.go
mlsandnerCI: I checked and it works 👍  The path must either be relative to the current directory or an absolute path to the file.

internal/cmdutils/resolve.go
```diff
@@ -0,0 +1,148 @@
...
+			bs, err = os.ReadFile(file)
+			if err != nil {
+				return "", err
+			}
+
+			matches, found := regexutil.FindAllNamedGroupsMatches(bazelFuzzTestFileNamePattern, string(bs))
+			if found {
+				for _, match := range matches {
+					if match["file"] == filepath.Base(path) {
+						return ConstructBazelRelativeLabel(match["fuzzTest"], file), nil
```
adombeck, line 75: This might match on the wrong source file. For example:

```
$ tree
.
├── cifuzz.yaml
├── libbar
│   ├── BUILD.bazel
│   └── fuzz_test.cc
├── libfoo
│   ├── BUILD.bazel
│   └── fuzz_test.cc
└── WORKSPACE
$ cifuzz run --resolve libfoo/fuzz_test.cc
```
In this example project, the path `libfoo/fuzz_test.cc` resolves  to `libbar:fuzz_test`.

Instead of checking only if the basename matches, we should check if the full path matches.

adombeck: The tests should also cover this.

internal/cmd/bundle/bundle.go
```diff
@@ -134,7 +134,12 @@ https://github.com/CodeIntelligenceTesting/cifuzz/issues`, opts.BuildSystem, sys
 				return cmdutils.WrapSilentError(err)
 			}
 
-			opts.FuzzTests = args
+			fuzzTests, err := cmdutils.FuzzTestArgument(opts.ResolveSourceFilePath, args, opts.BuildSystem, opts.ProjectDir)
+			if err != nil {
+				return cmdutils.WrapSilentError(err)
```
adombeck, line 139: This silences the error without printing it. Print it first via `log.Error(err)`  to avoid that.

internal/cmdutils/resolve.go
mlsandnerCI: Good catch, thanks!

mlsandnerCI: I reworked the test to cover this.
I decided to create a `testdata` folder instead of using the examples so I could add fuzz tests and check that this case is covered. I also included testing a relative and an absolute label.

internal/cmdutils/resolve/resolve.go
```diff
@@ -0,0 +1,152 @@
...
+
+			matches, found := regexutil.FindAllNamedGroupsMatches(cmakeFuzzTestFileNamePattern, string(bs))
+			if found {
+				for _, match := range matches {
+					if (filepath.IsAbs(path) && filepath.Join(projectDir, filepath.Dir(list), match["file"]) == path) ||
+						filepath.Join(filepath.Dir(list), match["file"]) == path {
+						return match["fuzzTest"], nil
+					}
+				}
+			}
```
adombeck, line 51: We don't need to check `found` because `matches` is nil when no matches are found:
```suggestion
			matches, _ := regexutil.FindAllNamedGroupsMatches(cmakeFuzzTestFileNamePattern, string(bs))
			for _, match := range matches {
				if (filepath.IsAbs(path) && filepath.Join(projectDir, filepath.Dir(list), match["file"]) == path) ||
					filepath.Join(filepath.Dir(list), match["file"]) == path {
					return match["fuzzTest"], nil
				}
			}
```

internal/cmdutils/resolve/resolve.go
```diff
@@ -0,0 +1,152 @@
...
+		cmakeLists, err := findAllCMakeLists(projectDir)
+		if err != nil {
+			return "", err
+		}
+
+		for _, list := range cmakeLists {
+			var bs []byte
+			bs, err = os.ReadFile(filepath.Join(projectDir, list))
+			if err != nil {
+				return "", err
```
adombeck, line 36: We return all errors from the standard library or third-party packages with `errors.WithStack`:
```suggestion
				return "", errors.WithStack(err)
```

internal/cmdutils/resolve/resolve.go
```diff
@@ -0,0 +1,152 @@
...
+		buildFiles, err = cmdutils.FindBazelBuildFiles("", workspace)
+		if err != nil {
+			return "", err
+		}
+
+		for _, file := range buildFiles {
+			var bs []byte
+			bs, err = os.ReadFile(file)
+			if err != nil {
+				return "", err
```
adombeck, line 70:
```suggestion
				return "", errors.WithStack(err)
```

internal/cmdutils/resolve/resolve.go
```diff
@@ -0,0 +1,152 @@
...
+		}
+		var buildFiles []string
+		buildFiles, err = cmdutils.FindBazelBuildFiles("", workspace)
+		if err != nil {
+			return "", err
+		}
+
+		for _, file := range buildFiles {
+			var bs []byte
+			bs, err = os.ReadFile(file)
```
adombeck, line 68: `cmdutils.FindBazelBuildFiles("", workspace)` returns file paths relative to the directory containing the `WORKSPACE` file but in line 68 you try to open the file relative to the current working directory. This will fail when the current working directory is not the directory containing the `WORKSPACE` file.

adombeck: We could do this instead:
```
bs, err = os.ReadFile(filepath.Join(workspace, file))
```

internal/cmdutils/resolve/resolve.go
```diff
@@ -0,0 +1,152 @@
...
+
+	"github.com/pkg/errors"
+
+	"code-intelligence.com/cifuzz/internal/cmdutils"
+	"code-intelligence.com/cifuzz/internal/config"
+	"code-intelligence.com/cifuzz/util/regexutil"
+)
+
+var cmakeFuzzTestFileNamePattern = regexp.MustCompile(`add_fuzz_test\((?P<fuzzTest>[a-zA-Z0-9_.+=,@~-]+)\s(?P<file>[a-zA-Z0-9_.+=,@~-]+)\)`)
+var bazelFuzzTestFileNamePattern = regexp.MustCompile(`cc_fuzz_test *\([^)]* ?name *= *['"](?P<fuzzTest>[a-zA-Z0-9_.+=,@~-]*)['"]\,\s*srcs\s*=*\s\[\s*['"](?P<file>[a-zA-Z0-9_.+=,@~-]*)['"]`)
```
adombeck, line 18:
```suggestion
var bazelFuzzTestFileNamePattern = regexp.MustCompile(`cc_fuzz_test *\([^)]* ?name *= *['"](?P<fuzzTest>[a-zA-Z0-9_.+=,@~-]*)['"]\,\s*srcs *= *\[\s*['"](?P<file>[a-zA-Z0-9_.+=,@~-]*)['"]`)
```

adombeck: As far as I can tell, this regex only matches the first source file

adombeck: @fmeum: Is there a way to find find a `cc_fuzz_test` rule which uses a given source file using `bazel query` or `bazel cquery`?

internal/cmdutils/resolve/resolve.go
```diff
@@ -0,0 +1,152 @@
...
+	})
+
+	return cmakeLists, errors.WithStack(err)
+}
+
+func FuzzTestArgument(resolveSourceFile bool, args []string, buildSystem, projectDir string) ([]string, error) {
+	if resolveSourceFile {
+		fuzzTest, err := resolve(args[0], buildSystem, projectDir)
+		if err != nil {
+			return nil, errors.Errorf("Failed to resolve source file: %v", err.Error())
```
adombeck, line 144: This replaces `err` and therefore we lose the stacktrace. If you want to attach a message, wrap the message either with `errors.WithMessage` or `fmt.Errorf` and the `%w` format specifier.

internal/cmdutils/resolve/resolve.go
mlsandnerCI: I checked with regex101 that it would match multiple `cc_fuzz_test(...)` in one file. Or what do you mean?

adombeck: The `srcs` attribute of `cc_fuzz_test` is a list, it can contain multiple labels. This regex will only match the first label.

adombeck: The `srcs` attribute of `cc_fuzz_test` is a list, it can contain multiple labels. This regex will only match the first label.

mlsandnerCI: Oh okay, I didn't know. Then this will not work, no.

fmeum: Bazel query can return a list of all targets created by the `cc_fuzz_test` macro in the same package that have a given source file. This is the same technique used by the IntelliJ plugin:
```
$ cd examples/bazel
$ bazel query 'attr(generator_function, cc_fuzz_test, same_pkg_direct_rdeps("src/explore_me_fuzz_test.cpp"))'
//src:explore_me_fuzz_test_raw_
```
To get from the target name to the name given to the `cc_fuzz_test` macro, the fixed `_raw_` prefix has to be stripped off.

For CMake, this information could probably be collected similar to how we retrieve the paths of seed corpus and executable in https://github.com/CodeIntelligenceTesting/cifuzz/blob/00d760904d0acb4f555d2b2941c7053fd316a27e/tools/cmake/modules/cifuzz-functions.cmake#L278-L292. The source files are available in `_args_sources` in that function. This does require running the equivalent of `cifuzz reload` first to populate the info files.

adombeck: The bazel solution looks perfect, I think that's what we want to do.

mlsandnerCI: I implemented the bazel solution 👍

mlsandnerCI: @fmeum Could you review the bazel solution because Adrian is not available?

internal/bundler/jazzer_bundler.go
```diff
@@ -245,3 +246,29 @@ func (b *jazzerBundler) createManifestJar(targetClass string) (string, error) {
 
 	return jarPath, nil
 }
+
+func listFuzzTests(projectDir string) ([]string, error) {
+	testDir := filepath.Join(projectDir, "src", "test", "java")
+
+	var fuzzTests []string
+	err := filepath.WalkDir(testDir, func(path string, d fs.DirEntry, err error) error {
+		if !d.IsDir() {
```
fmeum, line 255: Flipping this if condition could reduce indentation.

internal/bundler/jazzer_bundler.go
```diff
@@ -245,3 +246,29 @@ func (b *jazzerBundler) createManifestJar(targetClass string) (string, error) {
...
+			if err != nil {
+				return errors.WithStack(err)
+			}
+			fuzzTests = append(fuzzTests, classPath)
+		}
+
+		return nil
+	})
+	if err != nil {
+		log.Error(err, err.Error())
```
fmeum, line 269: Could this be `log.Error(err)` or is there a particular reason we need the second argument?

internal/cmdutils/bazel.go
```diff
@@ -1,15 +1,53 @@
...
 	"code-intelligence.com/cifuzz/pkg/log"
+	"code-intelligence.com/cifuzz/util/fileutil"
+	"code-intelligence.com/cifuzz/util/regexutil"
 )
 
+// This regex is based on the bazel bash completion script, see:
+// https://github.com/bazelbuild/bazel/blob/021c2a053780d697899cbcbd76a032c72cd5cbbb/scripts/bazel-complete-template.bash#L173
+var bazelFuzzTestTargetPattern = regexp.MustCompile(`cc_fuzz_test *\([^)]* {0,1}name *= *['"](?P<name>[a-zA-Z0-9_.+=,@~-]*)['"][^)]*\)`)
+
+func ConstructBazelAbsoluteLabel(target, buildFile string) string {
```
fmeum, line 24: Are the additions to this file still needed now that we use `query`?

internal/cmdutils/jazzer.go
```diff
@@ -0,0 +1,49 @@
...
+	return filepath.Join(projectDir, ".cifuzz-corpus", targetClass)
+}
+
+func ConstructJavaClassPathForFuzzTest(path, testDir string) (string, error) {
+	bytes, err := os.ReadFile(path)
+	if err != nil {
+		return "", errors.WithStack(err)
+	}
+
+	var classPath string
```
fmeum, line 32: Could we call this `classFilePath` instead for clarity (it's not a classpath) and define a new variable for the class name instead of mutating this one?

internal/cmdutils/pflag.go
```diff
@@ -168,6 +168,16 @@ func AddProjectDirFlag(cmd *cobra.Command) func() {
 	}
 }
 
+func AddResolveSourceFileFlag(cmd *cobra.Command) func() {
+	cmd.Flags().BoolP("resolve", "r", false,
+		"Argument of the command is a path to a source file instead of a test identifier.\n"+
+			"The path can be either absolute or relative to the current working directory and \n"+
+			"will be resolved to the corresponding identifier.")
```
fmeum, line 175: How about "identifier of the corresponding fuzz test"?

internal/cmdutils/resolve/resolve.go
```diff
@@ -0,0 +1,157 @@
...
+	"strings"
+
+	"github.com/pkg/errors"
+
+	"code-intelligence.com/cifuzz/internal/cmdutils"
+	"code-intelligence.com/cifuzz/internal/config"
+	"code-intelligence.com/cifuzz/util/regexutil"
+)
+
+var cmakeFuzzTestFileNamePattern = regexp.MustCompile(`add_fuzz_test\((?P<fuzzTest>[a-zA-Z0-9_.+=,@~-]+)\s(?P<file>[a-zA-Z0-9_.+=,@~-]+)\)`)
```
fmeum, line 20: Does this handle multiple files? I think it's probably fine if it doesn't, but in that case we should add a comment about potentially using the info files instead in the future.

internal/cmdutils/resolve/resolve.go
```diff
@@ -0,0 +1,157 @@
...
+		out, err := cmd.Output()
+
+		if err != nil {
+			// if a bazel query fails it is because no target could be found but it would
+			// only return "exit status 7" as error which is no useful information for
+			// the user, so instead we return the custom error
+			return "", errNoFuzzTest
+		}
+
+		suffix1 := "_raw_\n"   // Unix
```
fmeum, line 82: Instead of stripping two suffixes, I would find it cleaner to first strip newlines and then strip `_raw_` just once.

fmeum (approved):
Thanks for sticking through the longish review process, looks pretty good now!

internal/cmdutils/jazzer.go
mlsandnerCI: Should the function name be changed too? As I understand your comment, we actually return the class name and not a path so would it be `ConstructJavaClassNameForFuzzTest()`?

fmeum: It does seem to construct and return a classpath, but I don't fully understand what it looks like. @MarkusZoppelt Can you help here and possibly suggest some additional comments and/or variable names?

internal/cmdutils/resolve/resolve.go
mlsandnerCI: This is a great tip, thanks! I was unhappy with the solution but couldn't think of a better way.

internal/cmdutils/resolve/resolve.go
mlsandnerCI: I don't think so. I will add a TODO and a ticket in Jira so that we can revisit this if the situation occurs and we need to fix it.

internal/cmdutils/bazel.go
mlsandnerCI: Nope, thanks for pointing this out 👍

internal/cmdutils/jazzer.go
mlsandnerCI: So for command line completion of the fuzz test in the maven example, it takes the `path` and the `testDir`
```
path: /Users/.../IdeaProjects/cifuzz/examples/maven/src/test/java/com/example/FuzzTestCase.java
testDir: /Users/.../IdeaProjects/cifuzz/examples/maven/src/test/java   
```
creates the file path
```
com/example/FuzzTestCase.java  
```
and returns this
```
com.example.FuzzTestCase  
```

mlsandnerCI: Reading the definition of `classpath` I would say that the result should be calles `classPath` and while creating it we should call it either `filePath` or `classFilePath `.

MarkusZoppelt: > and returns this
> 
> com.example.FuzzTestCase

This. The function returns a "reverse domain"-style identifier from the file location of the fuzztest. So maybe we call it simply `fuzzTestCaseName` or `fuzzTestCaseIdentifier`?

mlsandnerCI: I pushed a version that I think should be descriptive enough.

MarkusZoppelt: lgtm

PR #527, opened 2022-12-13
Fix cmake installation for root users

florianGla:


adombeck (approved):

adombeck:
@florianGla I pushed a commit which changes the "Creating symlink" message to print the same as the other "Creating symlink" messages printed by the installer (i.e. only the path of the new symlink). Please take a look.

PR #528, opened 2022-12-13
build: Emit inputs on non-fatal ASan findings

adombeck:


tools/cmake/cifuzz/src/dumper.c
```diff
@@ -99,7 +101,8 @@ void __sanitizer_report_error_summary(const char *error_summary) {
    * Do not emit the input twice for ASan, which is always fatal.
    * TODO: This will change if we introduce --recover-asan.
    */
-  if (strncmp(UBSAN_SUMMARY_PREFIX, error_summary, strlen(UBSAN_SUMMARY_PREFIX)) == 0) {
+  if (strncmp(UBSAN_SUMMARY_PREFIX, error_summary, strlen(UBSAN_SUMMARY_PREFIX)) == 0 ||
+      strncmp(ASAN_SUMMARY_PREFIX, error_summary, strlen(ASAN_SUMMARY_PREFIX)) == 0) {
```
fmeum, line 105: Given that we only support ASan and UBSan, we could just remove this check entirely. I

If we want to make sure we don't emit duplicate summaries, we would have to inspect the value of `ASAN_OPTIONS`/`UBSAN_OPTIONS` here (rudimentary `strstr` for `halt_on_error=1` or `0` should be good enough).

adombeck: I'm now checking `ASAN_OPTIONS`/`UBSAN_OPTIONS` to avoid duplicate summaries. This makes it necessary to still check if the finding comes from ASan or UBSan.

adombeck: We still get duplicate summaries for LSan findings reported by ASan. Those are still fatal with `ASAN_OPTIONS=halt_on_error=0` and I found no way to make them non-fatal. The error summary says that it comes from ASan, so it's not trivial to detect whether it's actually an LSan finding:
```
SUMMARY: AddressSanitizer: 1 byte(s) leaked in 1 allocation(s).MS: 0 ; base unit: 0000000000000000000000000000000000000000
```

fmeum: My feeling is that users running with `ASAN_OPTIONS=halt_on_error=0` would not be looking for leaks, so we could just disable LSan in the `--recover-asan` case.

Alternatively, `strstr` for ` leaked ` should be a decent signal.

fmeum (approved):

tools/cmake/cifuzz/src/dumper.c
```diff
@@ -11,20 +12,60 @@ extern "C" {
...
+    }
+  }
+
+  if (strncmp(UBSAN_SUMMARY_PREFIX, error_summary, strlen(ASAN_SUMMARY_PREFIX)) == 0) {
+    char *options = getenv("UBSAN_OPTIONS");
+    /*
+    * The default for UBSan is to *not* halt on error, so we check if
+    * it was configured to do halt on error
+    */
+    if (strstr(options, "halt_on_error=1") || \
```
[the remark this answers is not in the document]
adombeck: done

fmeum (approved):

tools/cmake/cifuzz/src/dumper.c
adombeck: > Alternatively, strstr for leaked should be a decent signal to prevent duplicated summary lines.

I'm now checking if the summary line comes from LSan and don't dump the input in that case.

PR #529, opened 2022-12-13
cmd/remote-run: Print warning when --bundle is used with incompatible flags

adombeck:
Flags which only effect the build of the bundle don't have any effect when an existing bundle is specified via the --bundle flag. Users were confused by that, so we now print a warning in that case.


internal/cmd/remote-run/remote_run.go
```diff
@@ -45,11 +45,15 @@ type artifact struct {
...
-	BundlePath   string `mapstructure:"artifacts-path"`
 	Interactive  bool   `mapstructure:"interactive"`
 	PrintJSON    bool   `mapstructure:"print-json"`
 	ProjectName  string `mapstructure:"project"`
 	Server       string `mapstructure:"server"`
+
+	// Fields which are not configurable via viper (i.e. via cifuzz.yaml
+	// and CIFUZZ_* environment variables), by setting
+	// mapstructure:"-"
+	BundlePath string `mapstructure:"-"`
```
mlsandnerCI, line 56: Could you explain this comment, please? Why is `mapstructure:"-"` needed?

adombeck: We use [`viper.Unmarshal`](https://pkg.go.dev/github.com/spf13/viper#Unmarshal) to populate the fields of our options structs with values from viper: https://github.com/CodeIntelligenceTesting/cifuzz/blob/19404671282386864bde6210d73f459646a85b51/internal/config/config.go#L118-L118

Internally, `viper.Unmarshal` uses [`mapstructure.Decode`](https://pkg.go.dev/github.com/mitchellh/mapstructure#Decode) which uses the `mapstructure` tags to map field names of the input and output structs. Using `-` in the tag tells `mapstructure.Decode` to ignore this field (I'm not sure where exactly that's documented, but it's referenced in the comment of `IgnoreUntaggedFields` in https://github.com/mitchellh/mapstructure/blob/v1.5.0/mapstructure.go#L201).

mlsandnerCI: Thanks, that cleared it up for me!

mlsandnerCI (approved):

adombeck:
The branch had the wrong prefix, this fixed issue CLI-759

PR #530, opened 2022-12-13
build/other: Set CIFUZZ_COVERAGE_BUILD=1 for coverage builds

adombeck:
Some users have an elaborate setup around our "other" build system support and want to run slightly different build commands when building for coverage (to place build artifacts into a different directory).

To support this use case, we now set CIFUZZ_COVERAGE_BUILD=1 in the environment of the build command for coverage builds.


adombeck:
@pnoltof You could try if this branch meets your needs

PR #532, opened 2022-12-13
Remove mentions of AFL + refactoring

adombeck:
We don't support AFL and it's not on the roadmap to support it, so this PR contains the following changes:
* Remove all mentions of AFL in user-facing docs.
* Make engine non-configurable and always use "libfuzzer" as the value

Also contains some minor refactoring of the bundler package.


jochil (approved):

adombeck:
> Maybe updating the cifuzz.yamls in the example projects, too?

Right, forgot that. Done.

florianGla (changes requested):
If think you can also remove this comment in pflag.go:
`// TODO(afl): Also link to https://aflplus.plus/docs/fuzzing_in_depth/#a-collecting-inputs`

adombeck:
> If think you can also remove this comment in pflag.go: `// TODO(afl): Also link to https://aflplus.plus/docs/fuzzing_in_depth/#a-collecting-inputs`

I left the `TODO(afl)` comments on purpose because those are still valid TODOs for when we actually work on AFL support (which is what the `(afl)` is supposed to mean).

florianGla (approved):

PR #537, opened 2022-12-14
refactor: Simplify cmake include dir structure

florianGla:


fmeum:
I just want to make sure it's an informed choice. I see the mild benefit of slightly simplifying the directory structure and the mild drawback that users get stuff on their include path that they shouldn't see. 

The decision can go either way and I'm fine with either. If this change allows for further simplifications down the line, that would be a clear argument in its favor.

adombeck:
> I just want to make sure it's an informed choice. I see the mild benefit of slightly simplifying the directory structure and the mild drawback that users get stuff on their include path that they shouldn't see.

IIUC, with this change the user sees the following in their include path:
```
cifuzz/cifuzz.h
fuzzer/FuzzedDataProvider.h
legacy/fuzz_macro.h
```

Because `fuzz_macro.h` is below the `legacy` directory, it should be clear to users that they should not use this header unless they have legacy code which needs to use it (in which case they could either include it manually or use the `CIFUZZ_USE_DEPRECATED_MACROS` CMake variable to have it automatically included via the `-include` compiler flag). I don't really see an issue with that.

The benefit is that the horrible path
```
~/.local/share/cifuzz/share/include/cifuzz/cifuzz/cifuzz.h
```
becomes a little bit less horrible:
```
~/.local/share/cifuzz/share/include/cifuzz/cifuzz.h
```

@florianGla I think we could improve things a bit more by moving `~/.local/share/cifuzz/share/include` to `~/.local/share/cifuzz/include`. The `include` directory is usually not below `share`, see for example `ls /usr`. The same is true for `~/.local/share/cifuzz/share/src`.

fmeum:
Convinved :⁠-⁠) Even more so with the suggested follow-up change.

fmeum (approved):

adombeck (approved):
@florianGla I will create a follow-up PR to move the `include` and `src` directories.

PR #538, opened 2022-12-14
docs: add zip as bazel dependency

mlsandnerCI:
### Motivation & Context
A user couldn't test the bazel example project because they were missing `zip` and it is not written in our ReadMe as a dependency. (It is however written in the Bazel/Bazelisk ReadMe as a dependency and we link to that so this is just extra information to prevent this situation)


adombeck (approved):

adombeck:
It could be helpful for future reference to note in the commit message why exactly `zip` is needed (i.e. because some rules provided by `rules_fuzzing` depend on it).

adombeck:
> I checked the bazel docs and zip is mentioned as a general dependency

Then maybe just add a link that doc

fmeum:
No need to change anything, just providing context: Bazel ships both a C and a Java implementation of zip/unzip and shouldn't rely on the host zip, but rules_fuzzing uses it for OSS-Fuzz bundling and removing that dependency would require non-trivial work.

PR #542, opened 2022-12-15
Further simplify directory structure

adombeck:


adombeck:
@fmeum: I would appreciate if you could take a look at a231a40ce3a51be67d49cb36cd3926e5ebe95e9d because there I'm removing the workaround for symlinks on Windows. The tests are still passing and at least since 0cc3831be0214cf2f970de809ebd96b84eaa3da7 I would expect them to fail if symlinks on Windows are not copied correctly, but I might have missed something.

fmeum:
> @fmeum: I would appreciate if you could take a look at a231a40ce3a51be67d49cb36cd3926e5ebe95e9d because there I'm removing the workaround for symlinks on Windows. The tests are still passing and at least since 0cc3831be0214cf2f970de809ebd96b84eaa3da7 I would expect them to fail if symlinks on Windows are not copied correctly, but I might have missed something.

If symlinks are enabled, they work well also on Windows (they are slow, but so is filesystem IO on Windows in general and it doesn't matter on our context). However, AFAIK symlink support requires Developer Mode to be enabled and thus admin rights, both of which I can see being potentially prohibitive in an enterprise scenario. Until that changes, I would suggest we do not rely on symlinks on Windows for core functionality.

That said, junctions are fine if we can use them to simplify the logic.

adombeck:
> However, AFAIK symlink support requires Developer Mode to be enabled and thus admin rights, both of which I can see being potentially prohibitive in an enterprise scenario. Until that changes, I would suggest we do not rely on symlinks on Windows for core functionality.

Yes, git's symlink support requires development mode and I think we should not require that for cifuzz users - but the builder is only used in tests and to build the installer and for contributors who want to do these things IMO it's fine to require development mode in exchange to make our code simpler. What do you think?

fmeum:
Okay, if that part of the code only runs during installer builds and not as part of the installer, that's completely fine as long as we document the symlink requirement in `CONTRIBUTING.md`. I somehow thought that change affected the installer's runtime behavior.

jochil (commented):
that are a lot of nice cleanups... thanks :+1: 

Only one thing: Adding `src` and `include` to the repo roots feels a little bit like polluting the projects first level... specially as we tried to follow https://github.com/golang-standards/project-layout as much as possible. Maybe we can think about a better directory to move both directories there. For me `tools` would still be ok.
And `tools/include/cifuzz/cifuzz.h` is already better than `tools/cmake/cifuzz/nclude/cifuzz/cifuzz.h`

adombeck:
We could get rid of the top-level `src` directory by moving the various source files to `tools/dumper`, `tools/launcher` and `tools/replayer` and letting the builder copy them to a top-level `src` directory in the build directory. But I see no good solution to get rid of the `include` directory.

florianGla:
> We could get rid of the top-level `src` directory by moving the various source files to `tools/dumper`, `tools/launcher` and `tools/replayer` and letting the builder copy them to a top-level `src` directory in the build directory. But I see no good solution to get rid of the `include` directory.

I think that this approach would be the best compromise in our situation.

jochil:
> We could get rid of the top-level `src` directory by moving the various source files to `tools/dumper`, `tools/launcher` and `tools/replayer` and letting the builder copy them to a top-level `src` directory in the build directory. But I see no good solution to get rid of the `include` directory.

would be ok for me... But even `share` or `etc` would be ok.  It just feels strange that the `src` folder in a project does not contain go source code. 

On the other hand we can just go with the current solution and see how it feels/works

jochil (approved):

adombeck:
> We could get rid of the top-level src directory by moving the various source files to tools/dumper, tools/launcher and tools/replayer and letting the builder copy them to a top-level src directory in the build directory

Done

PR #543, opened 2022-12-15
Fix help message for seed corpus flag

florianGla:
I replaced *_inputs with a general message in order to also cover maven and gradle. I am open for better wordings.


internal/cmdutils/pflag.go
```diff
@@ -172,7 +172,8 @@ func AddSeedCorpusFlag(cmd *cobra.Command) func() {
 	// TODO(afl): Also link to https://aflplus.plus/docs/fuzzing_in_depth/#a-collecting-inputs
 	cmd.Flags().StringArrayP("seed-corpus", "s", nil,
 		"A `directory` containing sample inputs for the code under test,\n"+
-			"which is used in addition to inputs found in <fuzz test>_inputs.\n"+
+			"which is used in addition to inputs found in the inputs\n"+
+			"directory of the fuzz test.\n"+
```
adombeck, line 176: I think with this change we don't document anywhere anymore where the inputs are stored. We should create a follow-up task to document that. We might want to document that in the build system specific parts of the [long usage message](https://github.com/CodeIntelligenceTesting/cifuzz/blob/3cb67abef194ea15d3062265a6db71fa7675f9e3/internal/cmd/run/run.go#L124-L124), but we should make sure that the `seed-corpus` flag can be understood in all commands which use them (currently not all of those also have a build-system specific long usage message, at least the coverage command does not).

adombeck: We might also come to the conclusion that it's not necessary to document this path because we print it in the output of the `run` command when a finding is found. IMO we should think through the user flow and potential uses of the inputs directory.

florianGla: I will create a follow up ticket for the documentation.

jochil (approved):

adombeck (approved):

PR #544, opened 2022-12-16
minijail: Pass all environment variables to the binary

adombeck:
Since we aren't using minijail for security and don't claim that we are using it for hermeticity, we might as well pass all environment variables to the fuzz test. This can help improve compatibility with local dependencies or choices such as `JAVA_HOME`.

PR #546, opened 2022-12-16
minijail: Remove unused minijail0 functionality

adombeck:
This is based on #544, please review and merge that first.


adombeck:
@jochil @mlsandnerCI The Ubuntu 20.04 integration test fails because building a recent minijail requires a recent `libc6-dev`. Do we really want to require that our tests run on old systems like Ubuntu 20.04? We have the end-to-end tests which test that cifuzz can be *used* on Ubuntu 20.04, but running our tests requires building cifuzz which IMO is not something we have to support on old systems.

jochil:
> @jochil @mlsandnerCI The Ubuntu 20.04 integration test fails because building a recent minijail requires a recent `libc6-dev`. Do we really want to require that our tests run on old systems like Ubuntu 20.04? We have the end-to-end tests which test that cifuzz can be _used_ on Ubuntu 20.04, but running our tests requires building cifuzz which IMO is not something we have to support on old systems.

They were running on ubuntu 20.04 all the time... just last week (15th) github switched the `ubuntu-latest` label from 20.04 to 22.04. So besides a few days last week, the integration tests were running on 20.04 all the time.

> running our tests requires building cifuzz which IMO is not something we have to support on old systems.

Don't we have to ensure that cifuzz is builded on an old system to support old systems due to g`libc`?

jochil (approved):

pkg/minijail/minijail.go
```diff
@@ -15,6 +15,12 @@ import (
 )
 
 const (
+	// BindingsEnvVarName is an environment variable which users can
+	// use to specify additional Minijail bindings. The bindings must
+	// be separated by colon and can be specified in the same format
+	// that is supported by minijail's --bind-mount flag:
+	// <src>[,[dest][,<writeable>]], where <src> must be an absolute
+	// path and <writeable> is either 0 or 1.
```
jochil, line 23: what do you think about making this information also available outside the source code... maybe in the "sandboxing" section of the docs?

florianGla (approved):

pkg/minijail/minijail.go
adombeck: done, please take a look

jochil:
@adombeck I removed the "Update minijail0" from the PR title to avoid confusion in the changelog

pkg/minijail/minijail.go
jochil: looks good. thanks!

PR #547, opened 2022-12-20
Fix URL for testing badge

kyakdan:
A minor fix of the URL for the testing badge. See https://github.com/badges/shields/issues/8671


jochil:
thanks for the fix, but https://github.com/CodeIntelligenceTesting/cifuzz/pull/548 covered it already

PR #549, opened 2022-12-20
cmake: Fix paths in CMake config when installed as root

adombeck:
When installed as root, the CMake package is found at `/usr/local/share/cifuzz/cmake`. We have to resolve that path to `<install dir>/share/cmake` in order to find the src and include directories relative to that path.


jochil (approved):

fmeum (changes requested):
This raises the minimum required CMake version to 3.19, see https://cmake.org/cmake/help/latest/command/file.html#path-conversion. We should clarify with CS whether that is acceptable and adjust the requirements in the README and the deps checker.

adombeck:
> This raises the minimum required CMake version to 3.19, see https://cmake.org/cmake/help/latest/command/file.html#path-conversion. We should clarify with CS whether that is acceptable and adjust the requirements in the README and the deps checker.

Good catch! I'll try to use [`get_filename_component`](https://cmake.org/cmake/help/latest/command/get_filename_component.html) instead, which should be supported since CMake 3.4.

adombeck:
> Good catch! I'll try to use [get_filename_component](https://cmake.org/cmake/help/latest/command/get_filename_component.html) instead, which should be supported since CMake 3.4.

done

fmeum (approved):

PR #553, opened 2022-12-21
Select llvm tools based on CC/CXX

florianGla:
If CC or CXX are set, we try to find the corresponding llvm tools in the same folder. Otherwise we look for those tools in the path.


util/envutil/envutil.go
```diff
@@ -126,3 +127,25 @@ func QuotedCommandWithEnv(args []string, env []string) string {
...
+func CXXFromEnv(env []string) string {
+	cxx, found := LookupEnv(env, "CXX")
+	// we only check if the name contains clang++
+	// in order to handle cases like clang++-14
+	if found && strings.Contains(path.Base(cxx), "clang++") {
+		return cxx
+	}
+
+	return ""
+}
```
jochil, line 151: what do you think about generalizing these functions into something like
```
func GetEnvWithPathSubstring(env []string, key string, substring string) string {
	value, found := LookupEnv(env, key)
	if found && strings.Contains(path.Base(value), substring) {
		return cxx
	}
	return ""
}
```
And maybe a simple unit test

pkg/dependencies/definitions.go
```diff
@@ -66,7 +66,7 @@ var deps = Dependencies{
 			if err != nil {
 				return nil, err
 			}
-			log.Debugf("Found llvm-cov version %s in PATH: %s", version, path)
+			log.Debugf("Found llvm-cov version %s", version)
```
jochil, line 69: IMO it could be helpful still having the actual path inside the debug message

jochil:
> fix(dependencies): Select llvm tools based on CC/CXX

I think some more context in this commit message can be helpful

pkg/runfiles/finder.go
```diff
@@ -151,3 +154,27 @@ func (f RunfilesFinderImpl) findFollowSymlinks(relativePath string) (string, err
 
 	return resolvedPath, nil
 }
+
+func (f RunfilesFinderImpl) llvmToolPath(name string) (string, error) {
```
jochil, line 158: what do you think about covering this function with some unit test?
It implements a very specific behavior and maybe we want this covered by a single test case

util/envutil/envutil.go
florianGla: Great idea!

pkg/dependencies/definitions.go
florianGla: I agree. I removed the path, because llvm-cov can be found in PATH, CC or CXX. Maybe I should change the wording. llvm-symbolizer and llvm-profdata should be handled in the same way.

internal/build/build.go
```diff
@@ -58,13 +58,13 @@ func CommonBuildEnv() ([]string, error) {
 		// Set the C/C++ compiler to clang/clang++ (if not already set),
 		// which is needed to build a  binary with fuzzing instrumentation
 		// gcc doesn't have -fsanitize=fuzzer.
-		if val, ok := envutil.LookupEnv(env, "CC"); !ok || !strings.HasPrefix(path.Base(val), "clang") {
+		if val, ok := envutil.LookupEnv(env, "CC"); !ok || !strings.Contains(path.Base(val), "clang") {
```
jochil, line 61: I guess we can replace these checks with the `envutil.GetEnvWithPathSubstring` function, or?

jochil (approved):

internal/build/build.go
florianGla: Yes, done.

PR #555, opened 2022-12-22
Feature: Added support for socks proxy via HTTP_PROXY env variable

pnoltof:
Added support for socks proxy via HTTP_PROXY env variable

### Motivation & Context
Using a SOCKS proxy for cifuzz remote-run was currently not possible.
These changes enable users to configure a SOCKS Proxy via the HTTP_PROXY environment variable.

### Description
The added dialer picks up any configured proxy settings. If no proxy settings are configured it will to a direct connection.


jochil:
Thanks for the contribution. It seems like the checks for code formatting are failing... can you use our git hooks (https://github.com/CodeIntelligenceTesting/cifuzz/tree/main/githooks) or run `make fmt`?

internal/api/api.go
```diff
@@ -301,3 +303,11 @@ func ValidateURL(s string) error {
 	}
 	return nil
 }
+
+func GetCustTransportLayer() *http.Transport {
+	dialer := proxy.FromEnvironment()
+	dialContext := func(ctx context.Context, network, address string) (net.Conn, error) {
+		return dialer.Dial(network, address)
+	}
+	return &http.Transport{DialContext: dialContext, DisableKeepAlives: true}
```
jochil, line 312: What's the reason for `DisableKeepAlives`?

internal/api/api.go
```diff
@@ -229,7 +231,7 @@ func (client *APIClient) sendRequest(method, endpoint, token string) (*http.Resp
 	}
 	req.Header.Add("Authorization", "Bearer "+token)
 
-	httpClient := &http.Client{}
+	httpClient := &http.Client{Transport: GetCustTransportLayer()}
```
jochil, line 234: I think you don't have to use a custom function for this... this should work too
```go
proxyTransport := &http.Transport{ Proxy: http.ProxyFromEnvironment }
httpClient : = &http.Client{ Transport: proxyTransport }
```

jochil:
> adjusted order of imports as requested by our code formatting checks
> changes done by go mod tidy

Some little addition... IMO you can squash these commits into the main one. It would help to keep the git log clean

internal/api/api.go
pnoltof: Sadly this doesn't work. The main reason for it being this: https://github.com/golang/go/issues/24135
It seams like the DNS lookups are not handled correctly with the default go http proxy.
This is why we needed the net.proxy and the Dialer.

internal/api/api.go
```diff
@@ -256,3 +258,11 @@ func ValidateURL(s string) error {
 	}
 	return nil
 }
+
+func GetCustTransportLayer() *http.Transport {
+	dialer := proxy.FromEnvironment()
+	dialContext := func(ctx context.Context, network, address string) (net.Conn, error) {
+		return dialer.Dial(network, address)
+	}
+	return &http.Transport{DialContext: dialContext, DisableKeepAlives: true}
```
MarkusZoppelt, line 267: Could you please add an explanatory comment to this function? I can't really tell what it does just from looking at it.

Also: Not sure if `Cust` is an acronym or just short for `Custom`.
If it is the latter, I vote for just calling it `GetCustomTransportLayer` or specifying what is "custom" in the function name if possible.

internal/api/api.go
jochil: removed

internal/api/api.go
jochil: I added a linked + explanation

MarkusZoppelt (approved):

jochil (approved):

PR #556, opened 2022-12-22
build/other: Fix fuzz test executable lookup

fmeum:
a628f3031c84c27337facff8de3f67f71b725b4d regressed the look up of the fuzz test executable after the build by always using the name of the fuzz test as a project dir relative path.

Ref: CLI-799


jochil (approved):

adombeck (commented):
Thanks for fixing this

PR #558, opened 2023-01-02
fix|refactor: manifest.jar creation

jochil:
adds a new package to build the `manifest.jar` for the jazzer artifact bundle and makes sure it meets the requirements of the [JAR File Specification](https://docs.oracle.com/javase/8/docs/technotes/guides/jar/jar.html#Notes_on_Manifest_and_Signature_Files)

### Motivation & Context
The current implementation is broken and creates an invalid `MANIFEST.MF` inside the `manifest.jar`, with no line break at the end of the file) and also is not handling line length limits at all


pkg/java/manifest.go
```diff
@@ -0,0 +1,97 @@
...
+func entriesToString(entries map[string]string) (string, error) {
+	content := ""
+	for k, v := range entries {
+		// headers are not allowed to be wrapped, so without the ": "
+		// separator a header is not allowed to be longer than 70 bytes
+		if len(k) > 70 {
+			return "", errors.Errorf("invalid header, size > 70: %s", k)
+		}
+
+		// max length for a row is 72 bytes
```
fmeum, line 81: Does the newline character count towards the limit? We should add the answer to that question to this comment.

pkg/java/manifest.go
```diff
@@ -0,0 +1,97 @@
...
+		if len(rawEntry) > 72 {
+			content += strings.Join(stringutil.SplitAfterNBytes(rawEntry, 71), "\n ")
+		} else {
+			content += rawEntry
+		}
+		content += "\n"
+	}
+
+	// make sure there is always an empty line at the end of the file
+	if content == "" || content[len(content)-1:] != "\n" {
```
fmeum, line 93: I would prefer `content[len(content)-1] != '\n'` if that works.

pkg/java/manifest.go
```diff
@@ -0,0 +1,97 @@
...
+		// separator a header is not allowed to be longer than 70 bytes
+		if len(k) > 70 {
+			return "", errors.Errorf("invalid header, size > 70: %s", k)
+		}
+
+		// max length for a row is 72 bytes
+		rawEntry := fmt.Sprintf("%s: %s", k, v)
+		// we split after 71 to avoid " " breaking the line limit
+		if len(rawEntry) > 72 {
+			content += strings.Join(stringutil.SplitAfterNBytes(rawEntry, 71), "\n ")
```
fmeum, line 85: `strings.Builder` should be preferred to repeated concatenations.

util/stringutil/stringutil.go
```diff
@@ -120,3 +121,28 @@ func MaxLen(elems []string) int {
...
+
+// Splits a given string after n bytes (not characters)
+func SplitAfterNBytes(input string, n int) []string {
+	if n <= 0 || len(input) == 0 {
+		return nil
+	}
+
+	// ceiling needed to catch splits that are
+	// smaller than n
+	subs := int(math.Ceil(float64(len(input)) / float64(n)))
```
fmeum, line 133: Why do we need a floating point division here? The integer division `input / n` combined with the remainder mod `n` should give you the same information without any potential for non-arbitrary precision floating point operations leading to incorrect results.

pkg/java/manifest.go
jochil: Yes, I updated the comment... the spec is not super clear about it so I also added this link: https://stackoverflow.com/a/33144934

pkg/java/manifest.go
jochil: of course

util/stringutil/stringutil.go
jochil: good point... I replaced it with `subs := (len(input) + n - 1) / n`

pkg/java/manifest.go
```diff
@@ -0,0 +1,97 @@
...
+		// separator a header is not allowed to be longer than 70 bytes
+		if len(k) > 70 {
+			return "", errors.Errorf("invalid header, size > 70: %s", k)
+		}
+
+		// max length for a row is 72 bytes, including the EOL bytes
+		// and the " " separator according to:
+		// https://stackoverflow.com/a/33144934
+		entry := fmt.Sprintf("%s: %s", k, v)
+		if len(entry) > 72 {
```
fmeum, line 85: Do we really need this `if`? Assuming `stringutil.SplitAfterNBytes` returns a singleton slice if the entry is short enough, the following line should be a noop in that case.

pkg/java/manifest.go
```diff
@@ -0,0 +1,97 @@
...
+			// we split after 71 to avoid " " or EOL breaking the line limit
+			entry = strings.Join(stringutil.SplitAfterNBytes(entry, 71), "\n ")
```
fmeum, line 87: 71 characters plus linebreak plus preceding continuation character would exceed the 72 char limit. Do we maybe have to split after 70 bytes?

util/stringutil/stringutil.go
```diff
@@ -120,3 +120,28 @@ func MaxLen(elems []string) int {
...
+// Splits a given string after n bytes (not characters)
+func SplitAfterNBytes(input string, n int) []string {
+	if n <= 0 || len(input) == 0 {
+		return nil
+	}
+
+	// this rounds up the integer division to catch splits
+	// that are smaller than n
+	subs := (len(input) + n - 1) / n
+	result := []string{}
```
fmeum, line 133:
```suggestion
	var result []string
```

util/stringutil/stringutil.go
```diff
@@ -120,3 +120,28 @@ func MaxLen(elems []string) int {
...
 }
+
+// Splits a given string after n bytes (not characters)
+func SplitAfterNBytes(input string, n int) []string {
+	if n <= 0 || len(input) == 0 {
+		return nil
+	}
+
+	// this rounds up the integer division to catch splits
+	// that are smaller than n
```
fmeum, line 131:
```suggestion
	// This rounds up the integer division result so that
	// the last chunk (< n chars) is counted.
```

util/stringutil/stringutil.go
```diff
@@ -120,3 +120,28 @@ func MaxLen(elems []string) int {
 	}
 	return res
 }
+
+// Splits a given string after n bytes (not characters)
+func SplitAfterNBytes(input string, n int) []string {
```
fmeum, line 125: I think it's possible to avoid divisions and multiplications entirely with an `n`-step for loop: https://go.dev/play/p/8JpogE7T9f0

util/stringutil/stringutil_test.go
```diff
@@ -0,0 +1,41 @@
+package stringutil
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestSplitAfterNBytes(t *testing.T) {
+	result := SplitAfterNBytes("ABCDEF", 2)
```
fmeum, line 10: This test contains many unrelated test cases. I would say that this is a situation in which a table-based test would work really well.

util/stringutil/stringutil_test.go
```diff
@@ -0,0 +1,41 @@
...
+	assert.Len(t, result, 3)
```
fmeum, line 11: I would prefer stronger assertions. The return value of `SplitAfterNBytes(s, n)` is correct if it satisfies these properties:
1. All but possibly the last chunk have length `n`.
2. The last chunk has length `[1, n-1]`.
3. `strings.Join(SplitAfterNBytes(s, n)) == s`

With a table-based test, you could easily check all these properties for a number of representative examples.

util/stringutil/stringutil.go
```diff
@@ -120,3 +120,28 @@ func MaxLen(elems []string) int {
...
+	if n <= 0 || len(input) == 0 {
```
fmeum, line 126: I think that `n <= 0` indicates a programming error and we should `panic` in that case. Returning `nil` can end up hiding bugs.

util/stringutil/stringutil.go
jochil: nice... I like this, more straightforward. I added some checks for invalid input

Out of curiosity I ran a few benchmarks... difference is not so big
```
name         time/op
SplitOld-16   472µs ± 8%
SplitNew-16   427µs ±10%

name         alloc/op
SplitOld-16  1.07MB ± 0%
SplitNew-16  1.07MB ± 0%

name         allocs/op
SplitOld-16     224 ± 0%
SplitNew-16     224 ± 0%
```

util/stringutil/stringutil.go
jochil: removed the code

util/stringutil/stringutil.go
jochil: ok, I agree for `n <= 0`. 
What do you think about `len(input) == 0`? `nil` or `[]string{""}`?

fmeum: I think that either produces `""` when fed into `strings.Join`, so I don't think there is a strong argument in favor of one or the other. Technically `[]string{}` is another reasonable, slightly different option. 

All in all, I would probably do whatever `strings.Split` does for the sake of consistency.

pkg/java/manifest.go
jochil: You are right... not necessary anymore

pkg/java/manifest.go
jochil: yes, it is correct... there was also an offset in the tests, so they didn't fail. I fixed it.
But I gave it another thought and actually the limit is 71 for the first (and maybe the only) chunk as it can not have the " " separator... and 70 for the rest. I would prefer not to handle that case, as it should not have any implications and would make the code harder to understand

util/stringutil/stringutil_test.go
jochil: I cleaned it up a bit with table-based tests and added these assertions

util/stringutil/stringutil.go
jochil: It seems like the return value of `strings.Split` with an empty string depends on the given `sep`... it alters between `[]string{}` and `[]string{""}` (see https://go.dev/play/p/u3LRttzx_nL)

Without any `if` check our current implementation returns `[]string{}`... so I went with this

pkg/java/manifest.go
```diff
@@ -0,0 +1,97 @@
...
+		if len(k) > 70 {
+			return "", errors.Errorf("invalid header, size > 70: %s", k)
+		}
+
+		rawEntry := fmt.Sprintf("%s: %s", k, v)
+		// The max length for a row is 72 bytes, including the EOL bytes
+		// and the " " separator, according to: https://stackoverflow.com/a/33144934
+		// so we split after 70 to avoid " " and EOL breaking the line limit.
+		// Theoretical the limit for the first (and maybe the only) chunk
+		// is 71 as there can not be a " " separator. Go keep it simple we
```
fmeum, line 86:
```suggestion
		// is 71 as there can not be a " " separator. To keep it simple we
```

pkg/java/manifest_test.go
```diff
@@ -0,0 +1,130 @@
...
+		"Test4": "ABC",
+	}
+	result, err := entriesToString(entries)
+	require.NoError(t, err)
+
+	assert.Equal(t, 8, strings.Count(result, "\n"))
+
+	lines := strings.Split(result, "\n")
+	for _, line := range lines {
+		// as we removed the line break for splitting into []string
```
fmeum, line 87: We should not refer to what a part of the code *used to* do.

util/stringutil/stringutil.go
```diff
@@ -120,3 +120,22 @@ func MaxLen(elems []string) int {
 	}
 	return res
 }
+
+// Splits a given string after n bytes (not characters)
+func SplitAfterNBytes(s string, n int) []string {
+	if n <= 0 {
+		panic(fmt.Sprintf("invalid check size %d for string splitting", n))
```
fmeum, line 127:
```suggestion
		panic(fmt.Sprintf("invalid chunk size %d for string splitting", n))
```

fmeum (approved):

pkg/java/manifest_test.go
jochil: I was referring to the split two lines up in the test itself... because I thought the 71 might be confusing. But I replaced it with `SplitAfter`, so now we can check for 72

Issue #550, opened 2022-12-20
Gradle no method 'register()' for printClasspath


lukaswoellhaf:
I'm using cifuzz inside a gradle project. When trying to run the fuzzing tests I'm getting following error:

```
FAILURE: Build failed with an exception.

* Where:
Initialization script '/Users/<username>/.local/share/cifuzz/share/cifuzz/share/init.gradle' line: 9

* What went wrong:
Could not find method register() for arguments [printClasspath, init_3nwb7i25vko1ydtq2x5p6v36f$_run_closure1$_closure3@34739756] on task set of type org.gradle.api.internal.tasks.DefaultTaskContainer.

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 3s
```
gradlew: exit status 1

Which leads to following:

``` Gradle
tasks.register("printClasspath") {
    println "cifuzz.test.classpath=${sourceSets.test.runtimeClasspath.asPath}";
}

tasks.register("printBuildDir") {
    println "cifuzz.test.buildDir=${layout.buildDirectory.get()}";
}
```

Is there a possibility that my build.gradle file interferes the init.gradle file for cifuzz, and how could I solve this?

jochil:
Thanks for the issue

> Is there a possibility that my build.gradle file interferes the init.gradle file for cifuzz, and how could I solve this?

I think this can be, but I also need to dive a little bit deeper. Can you provide some insight in your `build.gradle`? Maybe a stripped down version or some parts of it?
And which gradle version are you using?

lukaswoellhaf:
I used Gradle version `4.8.1` which is as I notice now not compatible wit cifuzz. I upgraded to Gradle version `4.9` and the issue mentioned above resolved.
But unfortunately a new issue with the `init.gradle` file  occurs:

``` bash
❌ 
FAILURE: Build failed with an exception.

* Where:
Initialization script '/Users/lwoellhaf/.local/share/cifuzz/share/cifuzz/share/init.gradle' line: 35

* What went wrong:
Could not create task 'cifuzzTest' (Test)
> Selected test name cannot be null or empty.

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 0s
gradlew: exit status 1
```

[...]

PR #560, opened 2023-01-03
fix: get init script working for gradle >= v5.0

jochil:
### Motivation & Context
Fixes: https://github.com/CodeIntelligenceTesting/cifuzz/issues/550

### Description
Older gradle versions handled default values (for tests/reports) differently. In order to get it working this change makes sure that these values are set explicitly. 

Unfortunately some of the report properties we need (`outputLocation`, `required`, ...) are not supported by gradle 4.9... so we have to increase the minimum requirements to gradle 5.0 (which is just 4 month newer than 4.9)


lukaswoellhaf:
Alsothere is a typo in the `init.gradle `file building a build directory named `jacaco`:

`def execFile = layout.buildDirectory.file("jacaco/cifuzz.exec").get().asFile` must be changed to:
`def execFile = layout.buildDirectory.file("jacoco/cifuzz.exec").get().asFile`

jochil:
@lukaswoellhaf good catches, thanks a lot. The `.destination` properties are also working with recent gradle versions... that's nice. I updated the PR

You can find new installers here: https://github.com/CodeIntelligenceTesting/cifuzz/actions/runs/3854773627

docs/Installation-Guide.md
```diff
@@ -70,3 +70,19 @@ rd /s %APPDATA%/cifuzz
...
+change on the `main` branch and can also be used on other branches too.
+
+To use one of these installers the following steps have to be executed:
+1. Download the `cifuzz_installer` artifact from the successful pipeline run 
+2. Unpack the artifact, for example with `unzip cifuzz_installer.zip` 
+3. Chose the installer for your operating system and run it
+```
+chmod u+x cifuzz_installer_linux
+./cifuzz_installer_linux
+```
```
MarkusZoppelt, line 88: Why is this section part of this PR?

It's nice, it just might be better placed in another PR if it references another (internal?) issue.

jochil: Just because I noticed this was missing when working on the task and preparing a pre-release for a user.
I can move it to another PR if you prefer

MarkusZoppelt: I think it is fine to keep it here

MarkusZoppelt (approved):

PR #562, opened 2023-01-04
fix(lsan): parse fatal errors of leak sanitizer correctly

jochil:
A fatal error from leak sanitizer with this output `==12345==LeakSanitizer has encountered a fatal error.` was not detected by the parser, even as it produced a crashing input. With this change this should result in a proper finding now.


jochil:
@wojtekCodeInt  does this fix your problem? 
You can find the installers here: https://github.com/CodeIntelligenceTesting/cifuzz/actions/runs/3837573180

wojtekCodeInt:
Probably yes.  We had a workshop with a customer this week and I saw that some of the attendees had leak sanitizer errors. I think they were showing as proper findings. The attendees understood that it was a finding and they proceeded to debug it.

pkg/parser/sanitizer/sanitizer_parser.go
```diff
@@ -14,19 +14,22 @@ var (
...
+		parseAsFatalErrorReport,
 	}
-
-	finding = parseAsErrorReport(line)
-	if finding != nil {
-		return finding
+	for _, parser := range parsers {
+		if f := parser(line); f != nil {
+			return f
+		}
```
MarkusZoppelt, line 31: iirc, we want to avoid short if statements according to our best practices guidelines. Maybe consider reformatting to increase readability.

MarkusZoppelt (approved):

PR #564, opened 2023-01-09
Remove usage / note of bazel pre-release version because bazel 6.0.0 has a stable release

MarkusZoppelt:
### Motivation & Context
Since bazel has a [stable release >= 6](https://github.com/bazelbuild/bazel/releases/tag/6.0.0), we do no longer need to set `USE_BAZEL_VERSION` explicitly on macOS.

### Description
* removed explicit bazel version in e2e pipeline and vm setup
* removed note in `CONTRIBUTING.md`


jochil:
> * [Bazel >= 5.3.2](https://bazel.build/install)

Is this from the readme also relevant or is v6 only necessary for contributing?

fmeum:
> > * [Bazel >= 5.3.2](https://bazel.build/install)
> 
> Is this from the readme also relevant or is v6 only necessary for contributing?

6.0.0 is needed for contributing as we can assume that users already have a working Bazel build on macOS (which may require workarounds with Bazel <6 and new macOS versions).

fmeum (approved):

jochil (approved):

PR #568, opened 2023-01-10
`cifuzz run`: ask users to authenticate with a remote fuzzing server during fuzzing runs

MarkusZoppelt:
### Motivation & Context
We want to encourage users to authenticate with a remote fuzzing server backend to allow them to sync findings and get more advanced features.

### Description
`cifuzz run` opens a TUI dialogue that allows users to authenticate.

_This feature is currently put behind a feature flag. Set `CIFUZZ_PRERELEASE=1 cifuzz run ..`. to test this feature._


internal/cmd/run/run.go
```diff
@@ -655,3 +669,85 @@ func countCorpusEntries(seedCorpusDirs []string) (uint, error) {
 	}
 	return numSeeds, nil
 }
+
+// check if user is authenticated
```
mlsandnerCI, line 673: If you want to document the function I would recommend using the common pattern 
`<function name> <description>.`, e.g. `getAuthSTatus checks if user is authenticated.`  
However I think the function name is enough here, especially with the comments in the function it is pretty clear what it does.

internal/cmd/run/run.go
```diff
@@ -655,3 +669,85 @@ func countCorpusEntries(seedCorpusDirs []string) (uint, error) {
...
+
+	// assume that the first server is the one we want to check
+	server := servers[0]
+
+	// Obtain the API access token
+	token := login.GetToken(server)
+
+	if token == "" {
+		return false, nil
+	} else {
```
mlsandnerCI, line 690: You don't need an `else` here, you can just follow along with the function since you exit early if there is an error.

internal/cmd/run/run.go
```diff
@@ -655,3 +669,85 @@ func countCorpusEntries(seedCorpusDirs []string) (uint, error) {
...
+removing the token from %s.`, access_tokens.GetTokenFilePath())
+			log.Warn(err.Error())
+
+			return false, err
+		}
+	}
+	return true, nil
+}
+
+// ask user if they want to use SaaS if they are not authenticated
```
mlsandnerCI, line 706: Same as above but I would keep the description, I think it makes sense for this function.

internal/cmd/run/run.go
```diff
@@ -655,3 +667,85 @@ func countCorpusEntries(seedCorpusDirs []string) (uint, error) {
...
+func getAuthStatus() (bool, error) {
+	// check if user is authenticated
+
+	servers := access_tokens.GetServerURLs()
+	if len(servers) == 0 {
+		return false, nil
+	}
+
+	// assume that the first server is the one we want to check
+	server := servers[0]
```
jochil, line 683: why checking the first one? Shouldn't it go with `--server` flag or the default one ("app.code-intelligence.com")?

MarkusZoppelt: We are checking the contents of `access_tokens.json` here.

So either they are authenticated _(in which case the json file exists and we are using the first entry found in the file)_ or they are not. For the latter, we try to authenticate against the default one.

If users want to pass `--server`, they can/should use cifuzz login imho.

internal/cmd/run/run.go
MarkusZoppelt: Thanks! Done

internal/cmd/run/run.go
```diff
@@ -655,3 +667,86 @@ func countCorpusEntries(seedCorpusDirs []string) (uint, error) {
...
+
+		return false, err
+	}
+
+	return true, nil
+}
+
+// askForSaaSDialog ask users if they want to use a SaaS backend
+// if they are not authenticated
+func askForSaaSDialog(interactive bool) error {
```
jochil, line 708: IMO `SaaS` is misleading here... it can run in/against a on-prem environment as well

jochil, line 708: Why using `interactive` as a parameter here? The function can be part of the `cmd` struct or (what I would prefer) skip it completely and not call it when the `interactive` flag is set

internal/cmd/run/run.go
jochil: But couldn't the first one a server which the current project is not even linked against? A user can have multiple cifuzz projects with different servers on one machine... so just picking the first one seems wrong to me.

> If users want to pass --server, they can/should use cifuzz login imho.

I am not sure about that. The best way for a user (with a non-default server) would be entering the server in the `cifuzz.yaml` but also every command that interacts with a server should allow two specify the actual URL via flag. If not set we can fallback to the configuration (or at the end to the default one)

internal/cmd/run/run.go
MarkusZoppelt: Agree.

What would be a better name?

Some ideas:
* `askForServerBackendDialog()`
* `askForAuthenticationDialog()`

internal/cmd/run/run.go
MarkusZoppelt: In that case, should we let users pick a server (each run) which server to use with `dialog.Select(servers)`?
What would be the default when `--interactive=false` is passed?

internal/cmd/run/run.go
jochil: In this case we should move the check for a revoked token outside of the dialog... I mean if the user sets the `--interactive=false` flag we still want the findings to be uploaded, just without a dialog. Therefore the token has to be checked either way

jochil: `askForServerConnection()`
to me personally the `ask` seems a little bit odd... maybe
`showServerConnectionDialog()`

MarkusZoppelt: Done

MarkusZoppelt: Should be updated now and working as intended. Thx!

internal/cmd/run/run.go
MarkusZoppelt: This should be resolved after the latest push. The server value is now parsed as usual from the `--server` flag, the `cifuzz.yml` or uses the default value.

mlsandnerCI (approved):

PR #573, opened 2023-01-13
feat: add bazel version as dependency

mlsandnerCI:
### Motivation & Context
Customer success noticed that while it is written in the installation guide, it would be easier to debug problems if cifuzz enforced that the correct bazel version is installed to rule this out as a problem.

### Description
Bazel was added as a dependency to the commands plus some refactoring.


MarkusZoppelt (changes requested):
While we are at it anyway, I would suggest adding an additional check when running on `darwin` because there the dependency is actually bazel >= 6.0.0

MarkusZoppelt:
> Will do, thanks for the info!

Nice, thank you!

I just noticed this error when running:
```
> USE_BAZEL_VERSION=5.3.2 cifuzz run -C examples/bazel src:explore_me_fuzz_test
⚠️ cifuzz requires bazel 6.0.0 or higher, have 5.3.2
❌ Unable to run command due to missing/invalid dependencies.
For installation instruction see:

	https://github.com/CodeIntelligenceTesting/cifuzz#installation
```

The [installation section](https://github.com/CodeIntelligenceTesting/cifuzz#installation) in our README says 5.3.2 which is confusing given the error message above.

I would suggest changing it to:


[Bazel >= 5.3.2 (>= 6.0.0 on macOS)](https://bazel.build/install)

or something similar.

florianGla:
Does this check fail, if I have only bazelisk installed? If I understand correctly bazelisk detects the required bazel version and installs it somewhere.

MarkusZoppelt:
> Does this check fail, if I have only bazelisk installed? If I understand correctly bazelisk detects the required bazel version and installs it somewhere.

No, it should not. (fwiw, I only have `bazelisk` installed)

Actually, installing bazelisk provides a `bazel` executable, see https://github.com/bazelbuild/bazelisk#installation

MarkusZoppelt (approved):

PR #575, opened 2023-01-16
fix(jazzer): exclude maven and gradle classes from being instrumented

florianGla:
Exclude maven and gradle classes from being instrumented for fuzzing.

### Motivation & Context
A user reported failing fuzz tests for which jazzer crashed in internal gradle classes.


fmeum (changes requested):
We should maintain these excludes in Jazzer rather than cifuzz as it applies equally to fuzz tests started from IDEs. Can you instead submit a PR to update the fuzzing and regression test cases of https://github.com/CodeIntelligenceTesting/jazzer/blob/main/src/main/java/com/code_intelligence/jazzer/junit/AgentConfigurator.java?

PR #576, opened 2023-01-17
feat(api): provide finding info for upload

MarkusZoppelt:
### Description
- fix(api): encoding issue in campaign run
- feat(api): fill finding fields with correct info


internal/api/campaign_run.go
```diff
@@ -3,6 +3,7 @@ package api
 import (
 	"bytes"
 	"crypto/rand"
+	b64 "encoding/base64"
```
fmeum, line 6: Is there a reason why we can't use the default import name?

internal/api/campaign_run.go
```diff
@@ -35,6 +36,9 @@ type Campaign struct {
 // run name is used to identify the campaign run in the API for consecutive
 // calls.
 func (client *APIClient) CreateCampaignRun(project string, token string, fuzzTarget string, numBuildJobs uint) (string, string, error) {
+	// encode with base64 to handle "." in fuzz target names
```
fmeum, line 39: Why can't we handle this without base64? What part of the stack can't deal with `.`?

internal/api/finding.go
```diff
@@ -24,23 +25,143 @@ type Finding struct {
...
+	CurlCommand *string `json:"curl_command,omitempty"`
+	TagHeader   *string `json:"tag_header,omitempty"`
 }
 
 func (client *APIClient) UploadFinding(project string, fuzzTarget string, campaignRunName string, fuzzingRunName string, finding *finding.Finding, token string) error {
+	fType := string(finding.Type)
+	fShortDesc := finding.ShortDescription()
+	fTag := fmt.Sprint(finding.Tag)
+
+	// encode with base64 to handle special cases
```
fmeum, line 119: Please make this comment more specific. What breaks if we don't use base64?

internal/api/campaign_run.go
MarkusZoppelt: The backend. It causes weird encoding issues in their UI. cc @simonresch

MarkusZoppelt: Not just for `.` but for names in general.

internal/api/campaign_run.go
MarkusZoppelt: Fixed

internal/api/finding.go
MarkusZoppelt: Changed to 
```
// encode with base64 because the CI Fuzz Server backend
// expects a base64 encoded string
```

internal/api/campaign_run.go
MarkusZoppelt: **Without base64:**
Screenshot 2023-01-17 at 14 23 41

**With base64:**
Screenshot 2023-01-17 at 14 22 25

internal/api/finding.go
```diff
@@ -24,23 +25,144 @@ type Finding struct {
...
+	CurlCommand *string `json:"curl_command,omitempty"`
+	TagHeader   *string `json:"tag_header,omitempty"`
 }
 
 func (client *APIClient) UploadFinding(project string, fuzzTarget string, campaignRunName string, fuzzingRunName string, finding *finding.Finding, token string) error {
+	fType := string(finding.Type)
+	fShortDesc := finding.ShortDescription()
+	fTag := fmt.Sprint(finding.Tag)
+
+	// encode with base64 because the CI Fuzz Server backend
```
fmeum, line 119: If this is part of the API specification then I don't think we actually need either of the two comment here. Having them suggests that this is something specific to this client, but that doesn't seem to be the case.

fmeum:
Looks good to me, but @simonresch should approve the API interactions.

internal/api/finding.go
```diff
@@ -24,23 +25,142 @@ type Finding struct {
...
+	FirstSeenFinding *string `json:"first_seen_finding,omitempty"`
+	IssueTrackerLink *string `json:"issue_tracker_link,omitempty"`
+	State            *string `json:"state,omitempty"`
 }
 
 type ErrorReport struct {
 	Logs    []string `json:"logs"`
 	Details string   `json:"details"`
+
+	Type                 *string               `json:"type,omitempty"`
```
simonresch, line 38: What is the reason for having string pointers in this struct?

internal/api/finding.go
```diff
@@ -24,23 +25,142 @@ type Finding struct {
...
+	Description *string `json:"description,omitempty"`
+	URL         *string `json:"url,omitempty"`
+}
+
+type Severity struct {
+	Description *string  `json:"description,omitempty"`
+	Score       *float32 `json:"score,omitempty"`
+}
+
+type WebAppFindingDetails struct {
```
simonresch, line 109: I would vote for removing all structs and maybe also fields that are not used currently. Especially the `WebAppFindingDetails` will very likely never be set by findings uploaded from `cifuzz`.

internal/api/campaign_run.go
```diff
@@ -35,6 +36,8 @@ type Campaign struct {
 // run name is used to identify the campaign run in the API for consecutive
 // calls.
 func (client *APIClient) CreateCampaignRun(project string, token string, fuzzTarget string, numBuildJobs uint) (string, string, error) {
+	fuzzTarget = base64.StdEncoding.EncodeToString([]byte(fuzzTarget))
```
simonresch, line 39:
```suggestion
	fuzzTarget = base64.URLEncoding.EncodeToString([]byte(fuzzTarget))
```

We use the base 64 URL Encoding for fuzz target names.

internal/api/finding.go
```diff
@@ -24,23 +25,142 @@ type Finding struct {
...
+	IssueTrackerLink *string `json:"issue_tracker_link,omitempty"`
+	State            *string `json:"state,omitempty"`
 }
 
 type ErrorReport struct {
 	Logs    []string `json:"logs"`
 	Details string   `json:"details"`
+
+	Type                 *string               `json:"type,omitempty"`
+	InputData            *string               `json:"input_data,omitempty"`
```
simonresch, line 39: We could leave the type for this field as `[]byte` and rely on `json.Marshal()` to do the base64 encoding (see https://pkg.go.dev/encoding/json#Marshal).

simonresch (commented):
The API call looks good but I think we could do with exposing a bit less about our V1 API in the serialization structs. WDYT?

internal/api/finding.go
MarkusZoppelt: Good point. Will do!

internal/api/finding.go
MarkusZoppelt: Removed

internal/api/finding.go
MarkusZoppelt: Thx!

internal/api/finding.go
MarkusZoppelt: Done

PR #580, opened 2023-01-20
Add tests for `cifuzz run` with finding upload

MarkusZoppelt:


internal/cmd/run/run.go
```diff
@@ -252,6 +252,7 @@ depends on the build system configured for the project.
 		cmdutils.AddBuildOnlyFlag,
 		cmdutils.AddDictFlag,
 		cmdutils.AddEngineArgFlag,
+		cmdutils.AddInteractiveFlag,
```
jochil, line 255: This looks like a workaround to make the code testable which should be actually hidden behind the PRERELEASE flag, or?

MarkusZoppelt: Yes

internal/cmd/run/run.go
```diff
@@ -260,7 +260,7 @@ depends on the build system configured for the project.
 		cmdutils.AddUseSandboxFlag,
 		cmdutils.AddResolveSourceFileFlag,
 	}
-	if os.Getenv("CIFUZZ_PRERELEASE") != "" {
+	if os.Getenv("CIFUZZ_PRERELEASE") != "" || os.Getenv("CI") != "" {
```
jochil, line 263: As mentioned before... I am not a fan of this approach. IMO the CI should set the `CIFUZZ_PRERELEASE` to test the features explicitly instead of doing it for every CI run (including users)

integration-tests/cmake/cmake_test.go
```diff
@@ -411,3 +414,11 @@ func testRemoteRun(t *testing.T, cifuzzRunner *shared.CIFuzzRunner) {
 	testdata := cifuzzRunner.DefaultWorkDir
 	shared.TestRemoteRun(t, testdata, cifuzz)
 }
+
+func testRunWithUpload(t *testing.T, cifuzzRunner *shared.CIFuzzRunner) {
+	t.Parallel()
+
+	cifuzz := cifuzzRunner.CIFuzzPath
+	testdata := cifuzzRunner.DefaultWorkDir
+	shared.TestRunWithUpload(t, testdata, cifuzz)
```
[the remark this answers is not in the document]
MarkusZoppelt: I only test running with upload for cmake. Should we also test it for, e.g. one/all of the Java examples? Then it makes more sense to keep the function...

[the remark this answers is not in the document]
jochil: The upload should not be different between the the build systems, or? So I think we don't need to test it for multiple ones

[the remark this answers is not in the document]
MarkusZoppelt: No, it should behave the same. Then I guess I will inline the test.

internal/cmd/run/run.go
MarkusZoppelt: fixed

internal/cmd/run/run.go
MarkusZoppelt: Should be fixed now with a dedicated prerelease pipeline job in `.github/workflows/pipeline_tests.yml`

jochil (approved):

PR #582, opened 2023-01-23
feat(run): log build output in file

mlsandnerCI:
**Description**

Build output from commands are now logged into a file (unless the command is run with `--verbose`) and replaced with a `Build in progress...` message.

https://user-images.githubusercontent.com/112617007/215549767-56845b4f-eef0-4b45-9ee5-5ed46c193b89.mov

https://user-images.githubusercontent.com/112617007/215549756-e70fc468-1b0a-41e2-8d40-7ab6d2f76a3d.mov


internal/cmdutils/buildlog.go
```diff
@@ -0,0 +1,99 @@
...
+
+	"code-intelligence.com/cifuzz/pkg/log"
+)
+
+// region Log
+
+var buildLogPath string
+
+func BuildOutputToFile(projectDir string) (io.Writer, error) {
+	buildLogPath = filepath.Join(projectDir, "cifuzz-build.log")
```
fmeum, line 21: Maybe place this in `.cifuzz-build`? That way it won't pollute the project directory.

internal/cmdutils/buildlog.go
```diff
@@ -0,0 +1,99 @@
...
+	return os.OpenFile(buildLogPath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0755)
```
fmeum, line 22: Concurrent runs of cifuzz could mess up the log. I don't know what the best way to fix that is though, time to think.

mlsandnerCI: True. We could have different logs for different runs with an ID, e.g. "cifuzz-build-my_fuzz_test-48521.log". Or we have a lock around the log and append every time instead of truncating. But I think that could cause confusion in the logs because it would be difficult to know what belongs to which.

jochil: When we thinking about this we might also consider rotating the logs, to make sure they are not pilling up

internal/build/bazel/build.go
```diff
@@ -114,6 +114,10 @@ func (b *Builder) BuildForRun(fuzzTests []string) ([]*build.Result, error) {
 	cmd := exec.Command("bazel", "info", "output_base")
 	out, err := cmd.Output()
 	if err != nil {
+		// Stop spinner here explicitly to display it correctly
+		// before logging the error
+		cmdutils.StopBuildInProgressSpinner(err)
```
jochil, line 119: Having these `StopBuildInProgressSpinner` in all the build system specific code seems a little bit odd to me... these packages should not care about any spinner created by the command itself. IMO this is prone to be forgotten or overlooked in future changes

internal/cmdutils/buildlog.go
```diff
@@ -0,0 +1,99 @@
...
+// region ProgressSpinner
+
+var buildProgressSpinner *pterm.SpinnerPrinter
+
+func DisplayBuildInProgressSpinner() {
+	if ShouldLogBuildToFile() {
+		// Ignore error here because pterm won't return an error anyway
+		buildProgressSpinner, _ = pterm.DefaultSpinner.Start("Build in progress...")
+	}
+}
```
jochil, line 75: IMO it would be more straightforward having a generic spinner method `DisplaySpinner(msg string)` and doing the check in front of calling this function

internal/cmdutils/buildlog.go
```diff
@@ -0,0 +1,99 @@
...
+var buildLogPath string
+
+func BuildOutputToFile(projectDir string) (io.Writer, error) {
+	buildLogPath = filepath.Join(projectDir, "cifuzz-build.log")
+	return os.OpenFile(buildLogPath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0755)
+}
+
+// LogBuildMessage will either print to stdout or to the log file
+// depending on command mode.
+func LogBuildMessage(msg string) error {
```
jochil, line 27: I am not sure if we should have this function at all... what we wanted to hide is the specific output of the build system... if we are printing something during the build it might be still relevant for the user (or is hidden behind  `--verbose`)

internal/bundler/jazzer_bundler.go
```diff
@@ -83,7 +90,11 @@ func (b *jazzerBundler) assembleArtifacts(buildResults []*build.Result) ([]*arch
 		}
 
 		for _, runtimeDep := range buildResult.RuntimeDeps {
-			log.Printf("runtime dept: %s\n", runtimeDep)
```
jochil, line 86: We should delete this or using `log.Debug`. It is already to verbose

internal/build/bazel/build.go
mlsandnerCI: I totally agree that it shouldn't be there but I have no idea where to set it otherwise except in the log.Error() function and that lead to cyclic imports when I tried it.

internal/cmd/bundle/bundle.go
```diff
@@ -152,16 +152,26 @@ https://github.com/CodeIntelligenceTesting/cifuzz/issues`, opts.BuildSystem, sys
...
 					opts.BuildSystem)
 			}
+
 			return opts.Validate()
 		},
 		RunE: func(c *cobra.Command, args []string) error {
-			opts.Stdout = c.OutOrStdout()
-			opts.Stderr = c.OutOrStderr()
-
+			// TODO: add progress spinner
```
mlsandnerCI, line 174: @jochil This is the only part where I am unsure now. We don't have a separate run() function for bundle and call the bundler directly. To keep it consistent I would like to add the progress spinner in here but then it would not be correct, because it would surround the whole bundle process and not just the buil. But I am hesitant to add in the Bundle() function of the bundler.

jochil (approved):

PR #585, opened 2023-01-24
fix: bazel coverage error handling

mlsandnerCI:
### Description
When bazel coverage ran into an error, it didn't return the error and showed a success message linking to a non-existent URI.


internal/cmd/coverage/coverage.go
```diff
@@ -207,7 +207,8 @@ func (c *coverageCmd) run() error {
 
 	switch c.opts.BuildSystem {
 	case config.BuildSystemBazel:
-		tmpDir, err := os.MkdirTemp("", "bazel-coverage-")
+		var tmpDir string
+		tmpDir, err = os.MkdirTemp("", "bazel-coverage-")
```
jochil, line 211: I don't understand how the actual fix relates to the description of the PR?
This basically changes the scoping of the `err` variable and the only possible effect should be that the `if err` could be triggered with an unhandled error from above (what hopefully is not the case). The error handling of the actual coverage build is happening after the switch block

mlsandnerCI: Right now the error is overwritten in the switch/case which means it gets logged but not returned, causing the 
error check in line 269 to not fail on error which means a link to a non-existent coverage report gets created and printed to the user. I added a screenshot to the ticket in Jira to show the behavior. 

Maybe I misunderstood how this error is supposed to be handled but I don't believe that this is the desired outcome. By not overwriting the error anymore, it gets handled properly in the rest of the function.

jochil: Ok, I took another look and it make sense... but the error is not "overwritten". In the old code the `err` variable is scoped to the switch/case block so the error would never bubble out. This is the problem with having the actual error handling outside the block which is executing the function...  the fix will work, but honestly I would prefer thinking about restructuring that code so we do not run in this kind of problems (which are easy to overlook).
For example I am not sure if it is necessary/senseful to expose the "bazel-coverage-" temp dir to the actual coverage command, maybe this could be handled transparently in the report generation itself.

mlsandnerCI: Yup, I removed it an everything still works plus the bug is fixed. Thanks 👍

jochil: great

PR #588, opened 2023-01-25
Kotlin fuzz test completion

florianGla:
Provide fuzz test completion for kotlin projects


MarkusZoppelt (approved):
lgtm

internal/bundler/jazzer_bundler.go
```diff
@@ -174,10 +175,17 @@ func (b *jazzerBundler) checkDependencies() error {
 
 func (b *jazzerBundler) runBuild() ([]*build.Result, error) {
 	var fuzzTests []string
-	var err error
 
 	if len(b.opts.FuzzTests) == 0 {
-		fuzzTests, err = listFuzzTests(b.opts.ProjectDir)
+		kts, err := fileutil.Exists(filepath.Join(b.opts.ProjectDir, "build.gradle.kts"))
```
jochil, line 180: I think you can also use a kotlin based gradle setup in a Java project without any kotlin. So the assumption that a project with a `build.gradle.kts` is always a kotlin project might be wrong

internal/bundler/jazzer_bundler.go
```diff
@@ -174,10 +175,17 @@ func (b *jazzerBundler) checkDependencies() error {
...
+		if err != nil {
+			return nil, err
+		}
+		if kts {
+			fuzzTests, err = listFuzzTests(config.KOTLIN, b.opts.ProjectDir)
+		} else {
+			fuzzTests, err = listFuzzTests(config.JAVA, b.opts.ProjectDir)
+		}
```
jochil, line 188: I am also not sure if this is a correct assumption...IIRC you can mix them, having Kotlin classes next to java classes

jochil (changes requested):
I might be wrong but I think the whole assumption that a project with a kotllin based gradle setup has to be a kotlin only project is wrong. 

You can have it all ways, like a java project with a kotlin based gradle or a kotlin project with a grovy based gradle... and so on

internal/bundler/jazzer_bundler.go
MarkusZoppelt: I removed this check and now add/collect both (Java + Kotlin) fuzztests when using Gradle.

jochil: why gradle? This should be true for maven as well

MarkusZoppelt: Changed. We will now collect all fuzztests

internal/bundler/jazzer_bundler.go
```diff
@@ -174,13 +174,19 @@ func (b *jazzerBundler) checkDependencies() error {
...
 
 	if len(b.opts.FuzzTests) == 0 {
-		fuzzTests, err = listFuzzTests(b.opts.ProjectDir)
+		tests, err := listFuzzTests(config.JAVA, b.opts.ProjectDir)
 		if err != nil {
 			return nil, err
 		}
+		fuzzTests = append(fuzzTests, tests...)
+
+		tests, err = listFuzzTests(config.KOTLIN, b.opts.ProjectDir)
```
jochil, line 185: what do you think about just having one `listFuzzTests(config.JVM...` instead of two code blocks

internal/config/config.go
```diff
@@ -171,6 +172,37 @@ func DetermineBuildSystem(projectDir string) (string, error) {
 	return BuildSystemOther, nil
 }
 
+func DetermineGradleTestType(projectDir string) (FuzzTestType, error) {
```
jochil, line 175: I think `TestType` is somehow misleading... IMO it has nothing to do with the tests.
So gradle itself calls it Groovy DSL/Kotlin DSL or "Gradle Build Language"... the function name should reflect this

internal/config/config.go
```diff
@@ -171,6 +172,37 @@ func DetermineBuildSystem(projectDir string) (string, error) {
...
+	kts, err := fileutil.Exists(filepath.Join(projectDir, "build.gradle.kts"))
+	if err != nil {
+		return "", err
+	}
+	if kts {
+		return KOTLIN, nil
+	}
+
+	return JAVA, nil
```
jochil, line 184: This should be `GROOVY`

internal/config/config.go
```diff
@@ -171,6 +172,37 @@ func DetermineBuildSystem(projectDir string) (string, error) {
...
+}
+
+func IsValidJVMFuzzTestType(testType FuzzTestType) bool {
+	validJVMFuzzTestTypes := []FuzzTestType{JAVA, KOTLIN}
+
+	return sliceutil.Contains(validJVMFuzzTestTypes, testType)
+}
```
jochil, line 191: If we check for `JVM` as a test type we can get rid of this function

internal/completion/fuzz_test_completion.go
```diff
@@ -49,7 +49,20 @@ func ValidFuzzTests(cmd *cobra.Command, args []string, toComplete string) ([]str
...
+		if directive == cobra.ShellCompDirectiveError {
+			return nil, cobra.ShellCompDirectiveError
+		}
+		validFuzzTests = append(validFuzzTests, tests...)
+
+		tests, directive = validJVMFuzzTests(config.JAVA, toComplete, conf.ProjectDir)
+		if directive == cobra.ShellCompDirectiveError {
+			return nil, cobra.ShellCompDirectiveError
+		}
+		validFuzzTests = append(validFuzzTests, tests...)
```
jochil, line 63: same here... This could be one block if we just check for `JVM` instead of Java and kotlin

internal/config/config.go
MarkusZoppelt: Renamed

internal/bundler/jazzer_bundler.go
MarkusZoppelt: Should be simpler now after my latest refactor commit.

internal/cmdutils/jazzer.go
```diff
@@ -55,3 +50,42 @@ func ConstructJVMFuzzTestIdentifier(testType config.FuzzTestType, path, testDir
...
 }
+
+// ListJVMFuzzTests returns a list of all fuzz tests in the project.
+// The returned list contains the fully qualified class name of the fuzz test.
+func ListJVMFuzzTests(projectDir string) ([]string, error) {
+	var fuzzTests []string
+
+	languages := []string{"java", "kotlin"}
+	fileNameExtensions := []string{".java", ".kt"}
+
```
jochil, line 61: I would use a `map[string]string` for this, then you can just iterate over the map with `k, v := range` and don't have to access a second slice by `i`

internal/cmdutils/jazzer.go
```diff
@@ -48,3 +50,42 @@ func ConstructJavaFuzzTestIdentifier(path, testDir string) (string, error) {
...
+	}
+
+	return fuzzTests, nil
+}
+
+// ListFilesWithExtension returns a list of all files in the directory and its
+// subdirectories with the given extension.
+func ListFilesWithExtension(dir string, extension string) []string {
+	var files []string
+	err := filepath.Walk(dir, func(path string, info os.FileInfo, err error) error {
```
jochil, line 81: I think this can be a glob search too... might be more straightforward than the `filepath.Walk`

internal/config/config.go
```diff
@@ -171,6 +170,31 @@ func DetermineBuildSystem(projectDir string) (string, error) {
 	return BuildSystemOther, nil
 }
 
+func DetermineGradleBuildLanguage(projectDir string) (GradleBuildLanguage, error) {
+	kts, err := fileutil.Exists(filepath.Join(projectDir, "build.gradle.kts"))
```
jochil, line 174: There can be multi-project setups in grade where the project root does not contain a `build.gradle` but a `settings.gradle(.kts)`

jochil: I will take care of this in another fix/pr

jochil (approved):

PR #589, opened 2023-01-25
Rename macOS release binaries from `darwin`->`mac`

MarkusZoppelt:
Closes [CLI-836](https://code-intelligence.atlassian.net/browse/CLI-836)

This PR changes only the public, user-facing binary names for macOS builds / installers.
Internally, we still use `darwin` to identify `GOOS`.

[CLI-836]: https://code-intelligence.atlassian.net/browse/CLI-836?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

... and add _amd64 suffix to existing installers


jochil:
How do you want to handle the the situation that the current installers are named "_darwin" and the pipeline/install.sh are already looking for the new suffix? For example in the compatibility check during the release pipeline?

jochil (changes requested):
How do you want to handle the the situation that the current installers are named "_darwin" and the pipeline/install.sh are already looking for the new suffix? For example in the compatibility check during the release pipeline? If I am not wrong this should brake  our ability to release (without further fixes)
Also the timeframe between merging this and creating a new release

MarkusZoppelt:
> For example in the compatibility check during the release pipeline?

This is indeed an issue. imho, we should discuss internally if the renaming is worth the effort.

jochil (approved):
This is our solution to the potential breaking of the release pipeline/install.sh:

We split this PR.. this one contains the actual renaming which does not affect the `install.sh` and should have no breaking effect as long no new release was created. To address this we have a second PR (https://github.com/CodeIntelligenceTesting/cifuzz/pull/594) to merge immediately after the release containing this PR

PR #590, opened 2023-01-25
Activate `unused` linter preset

MarkusZoppelt:
### Motivation & Context
We want to catch more warnings with the `unused` linter rule.

### Description
Closes [CLI-839](https://code-intelligence.atlassian.net/browse/CLI-839)

[CLI-839]: https://code-intelligence.atlassian.net/browse/CLI-839?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.golangci.yaml
```diff
@@ -6,6 +6,7 @@ linters:
     - gocritic
     - errcheck
     - nonamedreturns
+    - unused
```
jochil, line 9: I guess it would be better to use the "unused" profile instead of the "unused" linter... It includes more linter that also detect the error that triggered that change

.golangci.yaml
```diff
@@ -23,3 +24,5 @@ linters-settings:
       - commentFormatting
       - ifElseChain
       - singleCaseSwitch
+
+args: --timeout=1m
```
jochil, line 28: Why the timeout?

MarkusZoppelt: Not adding a timeout lead to a timeout error in our CI and the error message suggested adding this.

jochil: because the default timeout is 1m ... setting it to 1m should not  fix the problem permanently ;)
it seems like this particular run was slower than usual or at least the following ones were fast enough

MarkusZoppelt: I see. What do you suggest as a good increased value? `--timeout=3m`?

jochil: sounds good... if it takes longer than that we should rethink our linter selection

integration-tests/cmake/cmake_test.go
```diff
@@ -100,7 +101,7 @@ func TestIntegration_CMake(t *testing.T) {
 		regexp.MustCompile(`^==\d*==ERROR: AddressSanitizer: heap-use-after-free`),
 	}
 	if runtime.GOOS != "windows" {
-		expectedOutputs = append(expectedOutputs, regexp.MustCompile(`^SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior`))
+		_ = append(expectedOutputs, regexp.MustCompile(`^SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior`))
```
fmeum, line 104: This looks like it silences a bug. We should check why `expectedOutputs` isn't used.

internal/bundler/libfuzzer_bundler.go
```diff
@@ -84,7 +84,7 @@ func (b *libfuzzerBundler) bundle() ([]*archive.Fuzzer, error) {
 	//       then we do want to fail if the user explicitly asked for
 	//       UBSan.
 	if runtime.GOOS != "windows" {
-		sanitizers = append(sanitizers, "undefined")
+		_ = append(sanitizers, "undefined")
```
fmeum, line 87: Same as above, if `sanitizers` isn't used, we should investigate why rather than silence the warning.

pkg/runner/jazzer/jazzer_runner.go
```diff
@@ -19,7 +19,8 @@ import (
 )
 
 // List of non-open-sourced hooks that should be used for Java API fuzzing.
-var customHooks = []string{
+// currently, these are unused
+var _ = []string{
```
fmeum, line 23: These are only relevant for core, where we already maintain our own patch on top of this list. You can just delete it.

integration-tests/cmake/cmake_test.go
MarkusZoppelt: Should be fixed. I moved the `expectedOutputs` check to testRun

internal/bundler/libfuzzer_bundler.go
MarkusZoppelt: Thanks! I noticed we set the same sanitizers in `buildAllVariants` later and do not need them here. I deleted the duplicate code.

internal/config/config_test.go
```diff
@@ -149,6 +149,9 @@ func TestDetermineBuildSystem_CMake(t *testing.T) {
 	defer fileutil.Cleanup(projectDir)
 
 	err = os.WriteFile(filepath.Join(projectDir, "CMakeLists.txt"), []byte{}, 0644)
+	if err != nil {
+		t.Fatalf("Failed to create CMakeLists.txt: %v", err)
+	}
```
jochil, line 154: why not `require.NoError(t, err, msg)`?

MarkusZoppelt: Good point! Changed.

integration-tests/cmake/cmake_test.go
```diff
@@ -60,7 +60,8 @@ func TestIntegration_CMake(t *testing.T) {
 	// Execute the create command
 	outputPath := filepath.Join("src", "parser", "parser_fuzz_test.cpp")
 	linesToAdd = cifuzzRunner.Command(t, "create", &shared.CommandOptions{
-		Args: []string{"cpp", "--output", outputPath}},
+		Args: []string{"cpp", "--output", outputPath},
+	},
```
jochil, line 64: still seems a little bit off the formatting... I think the common/consistent one would be
```
linesToAdd = cifuzzRunner.Command(t, "create", &shared.CommandOptions{
    Args: []string{"cpp", "--output", outputPath},
})
```

jochil (approved):

integration-tests/cmake/cmake_test.go
MarkusZoppelt: Done

PR #591, opened 2023-01-26
refactor: move logging of ExecErrors into commands 

mlsandnerCI:
This is a preliminary PR to #582.

### Motivation & Context
This PR removes error logging from the build package and coverage generators to reduce their functionality to only returning the appropriate errors and letting the commands handle them themselves.

### Description
ExecErrors are returned instead of logged and silenced and are now handled by the command that receives these errors.


MarkusZoppelt (approved):
Nice cleanup!

jochil (approved):
I agree... very nice

PR #597, opened 2023-01-31
fix(login): show correct url for token creation

MarkusZoppelt:
Since the switch to `url.JoinPath()`, the URLs for token creation habe been wrongly encoded:

**Wrong:**
`%s/dashboard/settings/account/tokens%3Fcreate&origin=cli`

**Correct:**
`%s/dashboard/settings/account/tokens?create&origin=cli`


internal/cmdutils/login/login.go
```diff
@@ -27,10 +27,11 @@ func GetToken(server string) string {
 }
 
 func ReadTokenInteractively(server string) (string, error) {
-	url, err := url.JoinPath(server, "/dashboard/settings/account/tokens?create&origin=cli")
+	url, err := url.JoinPath(server, "/dashboard/settings/account/tokens")
```
mlsandnerCI, line 30:
```suggestion
	url, err := url.JoinPath(server, "dashboard", "settings", "account", "tokens")
```
I'm not sure but wouldn't this make more sense than to write the slashes manually?

mlsandnerCI (approved):

PR #599, opened 2023-01-31
feat: support setting CMAKE_BUILD_TYPE flag

mlsandnerCI:
... since this is not the recommended build type we print a warning when used

### Motivation & Context
Multiple customers have noticed that setting the CMake flag `CMAKE_BUILD_TYPE` resulted in an unsuccessful fuzz test build because this flag changes the build tree and necessary files could not be found in our build process.

### Description
`cifuzz run my_fuzz_test -- -D CMAKE_BUILD_TYPE=<type>` is now supported but will print a warning that we do not recommend using this flag.


internal/build/cmake/cmake.go
```diff
@@ -390,15 +395,32 @@ func (b *Builder) fuzzTestsInfoDir() (string, error) {
...
 }
+
+// determineCmakeBuildConfiguration checks if the user added the flag
+// "-D CMAKE_BUILD_TYPE=..." because it will change the build tree.
+func (b *Builder) determineCmakeBuildConfiguration() string {
+	for _, arg := range b.Args {
+		if strings.Contains(arg, "CMAKE_BUILD_TYPE") {
+			_, dir, found := strings.Cut(arg, "=")
+			if found {
+				log.Warnf("We recommend to not set the CMAKE_BUILD_TYPE flag because we cannot guarantee a successful fuzz test build.")
```
MarkusZoppelt, line 419:
```suggestion
				log.Warnf("We recommend not setting the CMAKE_BUILD_TYPE flag, as we cannot guarantee a successful fuzz test build.")
```

internal/build/cmake/cmake.go
```diff
@@ -390,15 +395,32 @@ func (b *Builder) fuzzTestsInfoDir() (string, error) {
...
+	fuzzTestsDir = filepath.Join(buildDir, b.cmakeBuildConfiguration, ".cifuzz", "fuzz_tests")
 	if fileutil.IsDir(fuzzTestsDir) {
 		return fuzzTestsDir, nil
 	}
 	return "", os.ErrNotExist
 }
+
+// determineCmakeBuildConfiguration checks if the user added the flag
+// "-D CMAKE_BUILD_TYPE=..." because it will change the build tree.
+func (b *Builder) determineCmakeBuildConfiguration() string {
```
MarkusZoppelt, line 414: Maybe just a nitpick but I would rename this function, e.g. to `checkForUnsupportedCmakeBuildConfiguration()` (or something shorter) since it is more in line whith what it does.

internal/build/cmake/cmake.go
```diff
@@ -23,10 +23,10 @@ import (
 	"code-intelligence.com/cifuzz/util/sliceutil"
 )
 
-// The CMake configuration (also called "build type") to use for fuzzing runs.
+// The default and recommended CMake configuration (also called "build type") to use for fuzzing runs.
 // See enable_fuzz_testing in tools/cmake/modules/cifuzz-functions.cmake for the rationale for using this
 // build type.
-const cmakeBuildConfiguration = "RelWithDebInfo"
+const defaultCmakeBuildConfiguration = "RelWithDebInfo"
```
MarkusZoppelt, line 29: since this is a const, I don't feel a need to rename this imho

internal/build/cmake/cmake.go
jochil: Maybe we can use the CMake naming... so `BuildType` instead of `BuildConfiguration`?

jochil (approved):

internal/build/cmake/cmake.go
mlsandnerCI: Done, I called it `checkForUserCmakeBuildType()`

PR #600, opened 2023-02-01
feat: tell user how to persist project conf after run

MarkusZoppelt:
### Motivation & Context
When running with `CIFUZZ_PRERELEASE=true`, users can connect and upload findings to a remote CI Fuzz server backend.
Previously, after selecting a server project in the pterm dialog, the users' choice was not remembered/persisted in `cifuzz.yaml`.
We should persist the choice in the config as a conveniance feature.

*Note:* This will write all set config options during run to the config file with `viper.WriteConfig()`.

### Description
Writes the current config options (incl. server and project) to the `cifuzz.yaml` config file.

Closes [CLI-859](https://code-intelligence.atlassian.net/browse/CLI-859).

[CLI-859]: https://code-intelligence.atlassian.net/browse/CLI-859?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jochil (changes requested):
We agreed on not writing to the configuration (as it is owned by the user) after initially creating it. Usually we just show a message on how to persist or what to change in the config

MarkusZoppelt:
> We agreed on not writing to the configuration (as it is owned by the user)

I agree that the config is owned by the user, but that doesn't mean that we can't help them updating it automatically.

E.g., we can show something like this:

```
[... user picks a server project... ]
Do you want to remember this server project?
cifuzz will sync findings to $PROJECT in the future.
This will change the contents of $pwd/cifuzz.yaml
Update cifuzz.yaml? (Y/n)
```

jochil:
> > We agreed on not writing to the configuration (as it is owned by the user)
> 
> I agree that the config is owned by the user, but that doesn't mean that we can't help them updating it automatically.
> 
> E.g., we can show something like this:
> 
> ```
> [... user picks a server project... ]
> Do you want to remember this server project?
> cifuzz will sync findings to $PROJECT in the future.
> This will change the contents of $pwd/cifuzz.yaml
> Update cifuzz.yaml? (Y/n)
> ```

In general I like the idea of the dialog, but we should make sure that the following things are ensured:
* not removing anything from the file, like user comments
* that we have a valid yaml file at the end (breaking structures/levels)

IMO for now it would be also ok to keep it consistent with existing stuff and just show a command to append, see: https://github.com/CodeIntelligenceTesting/cifuzz/blob/main/internal/cmd/run/run.go#L199

MarkusZoppelt:
> IMO for now it would be also ok to keep it consistent with existing stuff and just show a command to append

I've done that now. We now only show the info with an `echo` command.

For the record, I like the idea of changing the config automatically more... if we can come up with a way that ensures

> - not removing anything from the file, like user comments
> - that we have a valid yaml file at the end (breaking structures/levels)

jochil (approved):

PR #602, opened 2023-02-02
fix: fuzz tests in subdirectories failed due to build log path

mlsandnerCI:
### Motivation & Context
Running fuzz tests in sub directories failed because they unintentionally changed the path of the build log. 

### Description
I replaced `/` in the fuzz test names with `_` so build logs for the fuzz test `test/my_fuzz_test` are now `build-test_my_fuzz_test.log`. 

I decided against moving them into their own directory in the logs directory (as in the log would be `.cifuzz-build/logs/test/build-my_fuzz_test.log`) because I think it would overcomplicate the matter and clutter up the logs folder. 
I also didn't want to remove the folder from the fuzz test name, in case two fuzz tests are called the same and the user needs the directory as distinction.
(CLI-866)


internal/cmdutils/buildlog.go
```diff
@@ -23,6 +24,9 @@ func BuildOutputToFile(projectDir string, fuzzTestNames []string) (io.Writer, er
 		}
 	}
 
+	// Make sure that calling fuzz tests in subdirs don't mess up the build log path
+	fuzzTestName = strings.ReplaceAll(fuzzTestName, "/", "_")
```
jochil, line 28: we should use `os.PathSeparator` instead of "/" to make sure it is working with windows, too

jochil (approved):
worked for me, thanks!

internal/cmdutils/buildlog.go
mlsandnerCI: I really need to get that tattooed somewhere so I don't always forget it 😄

PR #603, opened 2023-02-02
Add golangci-lint caching to prevent linter timeouts

JackuB:
PR to resolve occasional linter timeouts in GitHub Actions.

We run `make lint` in two workflows - `pipeline_push` and `pipeline_test`, both were timing out.

Turns out the timeout we had in config was ignored and default `1m` was used: https://golangci-lint.run/usage/configuration/#run-configuration

When running the `golangci-lint` in loop, only the first (uncached) command runs around 1m, that's why it was tripping the timeout only sometimes. (Staggering 900MB difference of memory usage between the runs!)
Screenshot 2023-02-02 at 18 42 16

golangci-lint recommends using their GitHub Action https://github.com/golangci/golangci-lint-action that already comes with a cache, but I don't think it's worth having two different entrypoints for a linter and we should stick with makefile. Used GH Actions' cache, with a static key.

Uncached run takes ~60seconds, with cache ~15.


MarkusZoppelt (approved):
Nice!

mlsandnerCI (approved):

.github/workflows/pipeline_tests.yml
```diff
@@ -26,9 +26,21 @@ jobs:
       - name: Check go mod tidy
         run: make tidy/check
 
+      - name: Restore golangci-lint files from cache
+        uses: actions/cache/restore@v3
+        with:
+          key: ${{ runner.os }}-golangci-lint
+          path: /home/runner/.cache/golangci-lint
```
jochil, line 33: is it necessary to have an absolute path here? IIRC it should also work with something like `golangci-lint`

jochil:
Maybe this is a good opportunity to switch to the official github action of golangci-lint: https://github.com/golangci/golangci-lint-action
It already includes caching.... what do you think?

.github/workflows/pipeline_tests.yml
JackuB: The path is an output from `$ golangci-lint cache status` 

Absolute path should be more explicit, but I have no strong feelings either way. WDYT?

JackuB:
@jochil I've looked in the GH Action, and their cache handling is better. But we should stick with the `make lint` usage for consistency. And for example, if we decide to add a new linter like commitlint or something.

.github/workflows/pipeline_tests.yml
jochil: I have a slight tendency towards a relative path because it seems less depended on the host directory structure (having `/home/runner for example`) but most probably github will never change that... so I am fine with both

jochil:
> @jochil I've looked in the GH Action, and their cache handling is better. But we should stick with the `make lint` usage for consistency. And for example, if we decide to add a new linter like commitlint or something.

If our caching works well enough (what is the case IMO)  I am fine with using `make lint`

PR #604, opened 2023-02-04
fix: print artifact location after successful bundle run

MarkusZoppelt:
### Motivation & Context
Previously, the output path/filename of the created bundle artifact was not shown.

### Description
This PR sets the contents of `opts.OutputPath` correctly.

Closes [CLI-858](https://code-intelligence.atlassian.net/browse/CLI-858)

[CLI-858]: https://code-intelligence.atlassian.net/browse/CLI-858?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jochil (changes requested):
As it broke unnoticed in the past, we should cover this with a test to avoid regression

jochil (approved):

PR #605, opened 2023-02-06
Provide a dedicated clean command other build systems

kyakdan:
### Motivation & Context
In most cases, we need to build the project only once to get the dependencies we need for our fuzz tests. After that, fuzz tests can be built with single compiler invocations. This PR reduces the build time when the project needs a long time to build. Avoiding cleaning the build before each fuzz test is also not a viable option since we need to do this before the coverage build so that we don't end up using artifacts from the fuzzing build.

### Description
This PR adjusts the current behavior for other build systems by executing the build command once for all fuzz tests instead of once for each fuzz test.


jochil (changes requested):
I can see the need for `cifuzz bundle` to build all fuzz tests at once, but `cifuzz run` only supports executing one fuzz test. So not mentioning/supporting the `FUZZ_TEST` env for building single fuzz tests in favor of building all at once might be counterproductive. And if I am not wrong this is also a breaking change

What do you think about having a `bundle_build_command` property or a flag indicating that the `build-command` can be used for building all tests?

0roman:
Please consider remote-run also as a use-case here.

kyakdan:
> I can see the need for `cifuzz bundle` to build all fuzz tests at once, but `cifuzz run` only supports executing one fuzz test. So not mentioning/supporting the `FUZZ_TEST` env for building single fuzz tests in favor of building all at once might be counterproductive. And if I am not wrong this is also a breaking change
> 
> What do you think about having a `bundle_build_command` property or a flag indicating that the `build-command` can be used for building all tests?

@jochil I've implemented the clean command option, as we discussed yesterday. Could you have a look?

jochil (approved):
great, thanks

PR #608, opened 2023-02-06
Use ldd to resolve dynamic dependencies

kyakdan:
### Motivation & Context
We now use `ldd` to resolve the exact dependencies of fuzz tests and only copy those to the artifact.

### Description
We currently copy all .so files in the build directory to the artifact. This has two drawbacks:
1. The artifact gets unnecessarily large if we copy unneeded .so files
2. We cannot handle cases where the fuzz tests depend on non-system libraries outside the build directory


internal/build/other/other.go
```diff
@@ -356,10 +356,24 @@ func (b *Builder) findSharedLibraries(executable string) ([]string, error) {
...
+		return false
+	}
+
+	for _, stdLibTopDir := range []string{"/usr", "/lib", "/lib64", "/lib32", "libx32"} {
+		if strings.HasPrefix(library, stdLibTopDir) {
+			return false
+		}
+	}
+
+	return true
```
jochil, line 378: It seems like this is partly duplicated this information: https://github.com/CodeIntelligenceTesting/cifuzz/blob/main/internal/bundler/libfuzzer_bundler.go#L38

Maybe we can think about a way to avoid this duplication

internal/bundler/libfuzzer_bundler.go
```diff
@@ -323,13 +323,14 @@ func (b *libfuzzerBundler) copyArtifactsToTempdir(buildResult *build.Result, tem
...
-		if !isBelow {
-			// If the file is not below the build dir, we assume
-			// that it was not created during the build, so we
-			// don't need to copy it
-			continue
+		var topDir string
+		if isBelow {
+			topDir = buildResult.BuildDir
+		} else {
+			topDir = "/"
```
jochil, line 331: I am not sure if we should do that and scan the the whole file system... beside performance concerns I am also afraid that we run into access right issues a lot.

IMO it is reasonable for this use case to expect the user creating symlinks for the needed libraries

internal/build/other/other.go
```diff
@@ -11,6 +11,7 @@ import (
 	"strings"
 
 	"github.com/pkg/errors"
+	"github.com/u-root/u-root/pkg/ldd"
```
jochil, line 14: does this package use the actual `ldd` command? If so we either have to add it to our documentation/dependencies or try to find a solution using only the `debug/elf` package from go itself

jochil (commented):
I would appreciate a few tests to verify the introduced functionality

internal/bundler/libfuzzer_bundler.go
kyakdan: Could you clarify your performance concerns? We are not doing any scanning here. This only ensures that dependencies outside the build directory are saved into unique locations inside the artifact. Here, we basically only remove `/` from the absolute path and use this as the relative path to into the temp dir. This means that `/path/to/my/shiny/libfoo.so` will be stored in `<tempdir>/path/to/my/shiny/libfoo.so`

internal/build/other/other.go
kyakdan: The package uses `debug/elf` to get the loader in the `.interp` header and then execute it with the `--list` flag

internal/build/other/other.go
kyakdan: I've overseen this. How about adding a public function to `fileutil` to check whether a given path a standard library or not?

internal/bundler/libfuzzer_bundler.go
jochil: Ah ok... it is based on the result of `ldd`. I thought we are using it for collecting all the shared libraries. Ok, is fine for me then

internal/build/other/other.go
jochil: sounds good to me

internal/build/other/other.go
jochil: Ok, perfect. works for me

internal/build/other/ldd.go
```diff
@@ -0,0 +1,45 @@
...
+		return nil, err
+	}
+
+	for _, fileInfo := range filesInfo {
+		if !isStandardSystemLibrary(fileInfo.FullName) {
+			sharedObjects = append(sharedObjects, fileInfo.FullName)
+		}
+	}
+
+	return sharedObjects, err
```
jochil: In this case it should have no effect, but usually `return sharedObjects, nil` would be more explicit as this is the "success" path

internal/ldd/ldd_windows.go
```diff
@@ -0,0 +1,5 @@
+package ldd
+
+func NonSystemSharedLibraries(executable string) ([]string, error) {
```
jochil, line 3: WDYT about a warning or even a panic when calling this functions, to make it clear that unimplemented functionality was triggered?

internal/cmd/run/run.go
```diff
@@ -636,6 +637,12 @@ func (c *runCmd) runFuzzTest(buildResult *build.Result) error {
 
 	switch c.opts.BuildSystem {
 	case config.BuildSystemCMake, config.BuildSystemBazel, config.BuildSystemOther:
+		libraryPaths, err := ldd.LibraryPaths(buildResult.Executable)
+		if err != nil {
+			return err
+		}
+		runnerOpts.LibraryDirs = libraryPaths
+
```
jochil, line 645: If we need to set this runner opt for all cases we use `libfuzzer.Runner` we should consider moving this inside the runner package, to make sure that it is not forgotten when using the package somewhere else

integration-tests/shared/bundle.go
```diff
@@ -369,11 +369,11 @@ func TestBundleGradle(t *testing.T, lang string, dir string, cifuzz string, args
 	assert.Equal(t, "Jazzer-Fuzz-Target-Class: com.example.FuzzTestCase\n", string(content))
 }
 
-func TestRunBundle(t *testing.T, dir string, cifuzz string, bundlePath string, args ...string) (*archive.Metadata, string) {
+func TestRunBundle(t *testing.T, dir string, cifuzz string, bundlePath string, cifuzzEnv []string, args ...string) (*archive.Metadata, string) {
 	// Bundle all fuzz tests into an archive.
 	cmd := executil.Command(cifuzz, args...)
 	cmd.Dir = dir
-	env, err := envutil.Setenv(os.Environ(), "BAR", "bar")
+	env, err := envutil.Setenv(cifuzzEnv, "BAR", "bar")
```
jochil, line 376: as you touched the environment handling of the integration tests... do you think it makes sense using `t.Setenv` instead?
I can imagine it reducing possible side effects

integration-tests/other/other_test.go
```diff
@@ -308,3 +306,12 @@ func createAndVerifyLcovCoverageReport(t *testing.T, cifuzz string, dir string,
 		21, 31, 41},
 		uncoveredLines)
 }
+
+func cifuzzEnv(workDir string) []string {
```
jochil, line 310: I would like a function name describing the actual action... like `setupCifuzzEnv` or more specific `addLibraryPathToEnv`

integration-tests/other/other_test.go
```diff
@@ -315,3 +315,10 @@ func cifuzzEnv(workDir string) []string {
 	}
 	return nil
 }
+
+func buildCommand() string {
```
jochil, line 319: sorry, it's picky... but what's about `getBuildCommand`?

integration-tests/cmake/cmake_test.go
```diff
@@ -129,6 +129,9 @@ func TestIntegration_CMake(t *testing.T) {
 	})
 
 	t.Run("bundle", func(t *testing.T) {
+		if runtime.GOOS != "Linux" {
```
jochil, line 132: I think this should be lowercase -> `linux`

integration-tests/other/other_test.go
```diff
@@ -201,8 +201,8 @@ func TestIntegration_Other_Bundle(t *testing.T) {
 	if testing.Short() {
 		t.Skip()
 	}
-	if runtime.GOOS == "windows" {
-		t.Skip("Other build systems are currently only supported on Unix")
+	if runtime.GOOS != "Linux" {
```
jochil, line 204: lowercase here, too

jochil (approved):
I like these changes, looking good... I have only minor things to comment about

integration-tests/shared/bundle.go
kyakdan: Could you clarify what side effects you are concerned about? `envutil.Setenv` does not change the host environment but returns a new list we use in the executed command.

internal/cmd/run/run.go
kyakdan: We only need to do this when dealing with native fuzzer executables. For Jazzer the "executable" is not set IIRC, right?

jochil: Yes and AFAIK `libfuzzer.Runner` works only with executables... jazzer should be handled in `jazzer.Runner`

internal/ldd/ldd_windows.go
kyakdan: This would mean we cannot execute `cifuzz run` on Windows for the `other` build systems. Would this be a regression?

internal/cmd/run/run.go
kyakdan: Done

internal/ldd/ldd_windows.go
jochil: ah... then we should go with the warning. I am not sure if anyone is using `other` with windows... but I guess it should be possible. @fmeum Am I wrong?

fmeum: We shouldn't be supporting other on Windows (yet).

kyakdan: Agreed. We skip the tests for other on Windows and keep the `panic` when we try to resolve the dependencies in this case. We can also `panic` earlier for other on Windows as well, but this can be done in another PR.

jochil: @fmeum what is the reason we are not supporting `other` on windows? When you are using `clang` and providing the right `build-command` shouldn't it work. In a quick check I found no exit conditions for windows

fmeum: We may be with `clang-cl` or `gcc`, but MSVC-based custom build systems on Windows will not work with the flags we pass in.

jochil: ok, but afaik we do not have code in the `other` case that enforces MSVC, or? So technically it should have worked with `clang(-cl)` before this change, or?

fmeum: We never tested it, bit that may very well have been the case. I don't think anybody used that though.

PR #609, opened 2023-02-06
chore: add editor config

JackuB:
Adds editorconfig, hopefully to prevent/limit occasional formatting issues

Screenshot 2023-02-07 at 16 06 42


.editorconfig
```diff
@@ -0,0 +1,14 @@
...
+root = true
+
+[*]
+indent_style = tab
+indent_size = 2
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[*.md]
```
jochil, line 13: I guess we can add `max_line_length=72`  for markdown... it's something we agreed on a while ago

JackuB: I'm not sure it's implemented yet https://github.com/editorconfig/editorconfig/issues/89

jochil: ok...  I found it here: https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties#max_line_length

jochil (approved):

.editorconfig
JackuB: Aw, it's not in the spec https://spec.editorconfig.org/#supported-pairs
We could add it, if some editor picks up, it's not wrong 🤷

PR #610, opened 2023-02-07
fix(buildlog): add "all" identifier if no fuzz test is specified

mlsandnerCI:
### Motivation & Context
`cifuzz bundle` failed when called without any fuzz tests because a name was expected when creating the build log file.

### Description
If no fuzz tests is given as an argument, we create a `build-all.log` file. Plus some refactoring.


internal/cmdutils/buildlog.go
```diff
@@ -16,16 +16,22 @@ import (
...
+	var fuzzTestName string
+	switch {
+	case len(fuzzTestNames) == 0:
+		fuzzTestName = "all"
+	case len(fuzzTestNames) > 1:
+		fuzzTestName = fuzzTestNames[0]
 		for i := 1; i < len(fuzzTestNames); i++ {
 			fuzzTestName += "_" + fuzzTestNames[i]
 		}
+	default:
```
jochil, line 29: I think you can just use `strings.Join(fuzzTestNames, "_")` instead of the loop

internal/cmdutils/buildlog.go
```diff
@@ -16,16 +16,22 @@ import (
 var buildLogPath string
 
 func BuildOutputToFile(projectDir string, fuzzTestNames []string) (io.Writer, error) {
-	// Handle multiple fuzz tests
-	fuzzTestName := fuzzTestNames[0]
-	if len(fuzzTestNames) > 1 {
+	// Determine identifier for file
+	var fuzzTestName string
```
jochil, line 20: IMO now that we are using `all` another name would necessary... maybe `fileSuffix` or `logSuffix`... WDYT?

jochil (changes requested):
I guess we should add a regression test here, too... a basic function of the `bundle` command broke and the tests did not catch it. We should cover this

internal/cmdutils/buildlog.go
mlsandnerCI: `logSuffix` makes sense, thanks 👍

internal/cmdutils/buildlog.go
mlsandnerCI: That makes it way cleaner, thanks!

jochil (approved):
perfect, thanks

PR #611, opened 2023-02-07
fix: improve multi-project setups for gradle

jochil:
### Motivation & Context
Our current gradle integration does not handle multi-project setups very well. This PR should improve this

### Description

- showing a warning when cifuzz is initialized in the root of a multi-project setup
- support java v1.8, too
- search for gradle (wrapper) in parent directories. in multi-project setups the sub projects usually do not contain `gradlew`


internal/cmd/init/init.go
```diff
@@ -30,6 +30,10 @@ var gradleGroovySetup string
 //go:embed instructions/gradlekotlin
 var gradleKotlinSetup string
 
+const (
+	GradleMultiProjectWarningMsg = "For multi-project builds, you should setup cifuzz in the subprojects containing the fuzz tests."
+)
```
MarkusZoppelt, line 35: I think this message can be longer to be more helpful.

E.g.

```suggestion
	GradleMultiProjectWarningMsg = `For multi-project builds, you should run cifuzz init in each of the sub-projects
that contain fuzz tests and configure their build configuration individually.`
)
```

```

File: integration-tests/shared/cifuzz_runner.go [hunk 2]
```diff
@@ -51,24 +49,43 @@ func (r *CIFuzzRunner) Command(t *testing.T, command string, opts *CommandOption
...
 	return linesToAdd
 }
 
+// Command runs "cifuzz <command> <args>" and returns any indented lines
+// which the command prints to stderr (which we expect to be lines which
+// should be added to some source or config file).
+func (r *CIFuzzRunner) CommandAndGetLinesToAdd(t *testing.T, command string, opts *CommandOptions) []string {
+	allLines := r.Command(t, command, opts)
+	return FilterLinesToAdd(allLines)
+}
```
MarkusZoppelt, line 87: I was thrown off by the `And` in the function name, because it violates the single-responsibility principle...

imho, we should either
1. split this function up into two separate functions or 
2. rename the function that makes clear that this is a defined, cohesive task. 
Some ideas: `SetupCommand`, `InstructionCommand`, `CommandWithFilteredOutputs`.

pkg/dependencies/version.go
```diff
@@ -24,7 +24,7 @@ var (
 	clangRegex = regexp.MustCompile(`(?m)clang version (?P<version>\d+\.\d+(\.\d+)?)`)
 	cmakeRegex = regexp.MustCompile(`(?m)cmake version (?P<version>\d+\.\d+(\.\d+)?)`)
 	llvmRegex  = regexp.MustCompile(`(?m)LLVM version (?P<version>\d+\.\d+(\.\d+)?)`)
-	javaRegex  = regexp.MustCompile(`(?m)version \"(?P<version>\d+(\.\d+\.\d+)?)\"`)
+	javaRegex  = regexp.MustCompile(`(?m)version \"(?P<version>\d+(\.\d+\.\d+)?)(_\d+)?\"`)
```
MarkusZoppelt, line 27: do we also need a `kotlinRegex`?

jochil: This is checking the versions strings produced by `java --version`... there is no equivalent in kotlin as it runs in the same JVM

internal/cmd/init/init.go
jochil: I am not sure what this should clarify? In Gradle it is clear that you define the needed dependencies in the subproject,,, what else do you mean with "build configuration"?

MarkusZoppelt: the specific build configuration files (e.g. `build.gradle`, etc.)
Just a suggestion

pkg/dependencies/version.go
MarkusZoppelt: Both of these work, yet they use the same JRE as java, as you said.

```
> kotlin -version
Kotlin version 1.8.10-release-430 (JRE 19.0.2)
```

```
> kotlinc -version
info: kotlinc-jvm 1.8.10 (JRE 19.0.2)
```

internal/cmd/init/init.go
jochil: this should all be there already and it is not part of what cifuzz deal with... so if it is ok for you, I just would keep it like this

pkg/dependencies/version.go
jochil: ah, good to know... but as we only use `java` internally (eg for running the fuzz test) I think we don't need to check for `kotlin`

integration-tests/shared/cifuzz_runner.go
jochil: I had to go with `CommandWithFilterForInstructions()` as we already have a `Run` function on this struct

MarkusZoppelt (approved):

PR #613, opened 2023-02-07
feat: add error details to findings

MarkusZoppelt:
This is a WIP since the endpoint providing the `error-details.json` is not available yet.
Until then, the details are taken from a local file stored in `~/.local/share`.
Getting the infos from the file should fail gracefully, therefore this can be merged imho.

**Preview of the resulting backend upload:**
Screenshot 2023-02-07 at 16 14 45

Blocked by [APP-1249](https://code-intelligence.atlassian.net/browse/APP-1249)

Closes [CLI-785](https://code-intelligence.atlassian.net/browse/CLI-785)

[APP-1249]: https://code-intelligence.atlassian.net/browse/APP-1249?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CLI-785]: https://code-intelligence.atlassian.net/browse/CLI-785?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/api/error_details.go
```diff
@@ -0,0 +1,95 @@
...
+)
+
+// GetErrorDetails gets the error details from the API
+func (client *APIClient) GetErrorDetails(token string) (ErrorDetails, error) {
+	// FIXME: Until the endpoint is implemented, we parse the error details from the
+	// local file system at ~/.local/share/error-details.json
+	errorFile := filepath.Join(os.Getenv("HOME"), ".local", "share", "error-details.json")
+	if _, err := os.Stat(errorFile); os.IsNotExist(err) {
+		return nil, errors.Wrap(err, "error details file not found")
+	}
```
jochil, line 58: It's temporary but you can also use `fileutil.Exists`

internal/api/error_details.go
```diff
@@ -0,0 +1,95 @@
...
+	Description  string         `json:"description,omitempty"`
+	Severity     *ErrorSeverity `json:"severity,omitempty"`
+	Mitigation   string         `json:"mitigation,omitempty"`
+	Links        []Link         `json:"links,omitempty"`
+	CweDetails   *EDetails      `json:"cwe_details,omitempty"`
+	OwaspDetails *EDetails      `json:"owasp_details,omitempty"`
+	Language     int64          `json:"language,omitempty"`
+}
+
+type EDetails struct {
```
jochil, line 26: The name is confusing, first because of the existence of `ErrorDetail` and `ErrorDetails` and Second because it is plural but it is basically a smaller version of `ErrorDetail` which is singular... whats about `ExternalDetail`?

internal/api/error_details.go
```diff
@@ -0,0 +1,95 @@
...
+	err = json.NewDecoder(file).Decode(&errorDetails)
+	if err != nil {
+		return nil, errors.Wrap(err, "error decoding error details")
+	}
+
+	return errorDetails, nil
+}
+
+// GetErrorDetailFromID gets a specific error detail specified by the ID
+func (errorDetails ErrorDetails) GetErrorDetailFromID(id string) *ErrorDetail {
```
jochil, line 76: What do you think about `FindByName` and `FindByID`?

internal/api/finding.go
```diff
@@ -64,11 +64,21 @@ type Environment struct {
...
+	Links        []Link    `json:"links,omitempty"`
+	OwaspDetails *FDetails `json:"owasp_details,omitempty"`
+	CweDetails   *FDetails `json:"cwe_details,omitempty"`
+	Language     string    `json:"language,omitempty"`
+}
+
+type FDetails struct {
+	ID          int64  `json:"id,omitempty"`
+	Name        string `json:"name,omitempty"`
+	Description string `json:"description,omitempty"`
```
jochil, line 81: This are basically the same structs as `ErrorDetail` and `EDetails`... why are we not reusing them? This seems to be prone to forgot updating both structs when new fields where added

internal/api/finding.go
```diff
@@ -150,3 +153,66 @@ func (client *APIClient) UploadFinding(project string, fuzzTarget string, campai
 
 	return nil
 }
+
+func constructDetails(errorDetails ErrorDetails, finding *finding.Finding) MoreDetails {
```
jochil, line 157: 1. As mentioned before, I am not sure if we should have or need both structs
2. the `MoreDetails` return should be a pointer
3. can be a `New` function

jochil (changes requested):
IMO the whole adding of the information should happen in another place. When the application detects a finding we should immediately use the information from `error.json`  and not just before we upload it

internal/api/error_details.go
MarkusZoppelt: I like `ExternalDetail` - thank you for the suggestion!

internal/api/finding.go
MarkusZoppelt: Didn't catch that. Thx!

internal/api/finding.go
MarkusZoppelt: I refactored the code, so that finding details are now directly added before saved to disk. This happens in [report_handler.go](https://github.com/CodeIntelligenceTesting/cifuzz/pull/613/files#diff-89d321fb00b8c9f72672304282cd096ef5d0f3f186767e6a867e980244a62fbe)

MarkusZoppelt:
@jochil I added the changes you requested. Would you please review again? Thank you!

pkg/finding/finding.go
```diff
@@ -306,3 +324,86 @@ func LoadFinding(projectDir, findingName string) (*Finding, error) {
...
+		Mitigation:   "",
+		Links:        []Link{},
+		OwaspDetails: &ExternalDetail{},
+		CweDetails:   &ExternalDetail{},
+	}
+
+	// find error details for specific finding
+	var details *ErrorDetails
+	for _, d := range errorDetails {
+		if strings.EqualFold(d.Name, f.ShortDescriptionColumns()[0]) {
```
jochil, line 357: can you add a TODO here to make it clear that we have to optimize the matching of the errors?

pkg/finding/finding.go
```diff
@@ -306,3 +324,86 @@ func LoadFinding(projectDir, findingName string) (*Finding, error) {
...
+			}
+		}
+		// if the finding is an OWASP, add more details
+		if details.OwaspDetails != nil {
+			moreDetails.OwaspDetails = &ExternalDetail{
+				ID:          details.OwaspDetails.ID,
+				Name:        details.OwaspDetails.Name,
+				Description: details.OwaspDetails.Description,
+			}
+		}
```
jochil, line 393: Maybe I am missing something, but this whole block seems unnecessary and prone to making mistakes when changing it. Why not just `moreDetails = details`? It is basically the same struct, why duplicating it by hand?

pkg/finding/finding.go
```diff
@@ -306,3 +324,86 @@ func LoadFinding(projectDir, findingName string) (*Finding, error) {
...
+	// find error details for specific finding
+	var details *ErrorDetails
+	for _, d := range errorDetails {
+		if strings.EqualFold(d.Name, f.ShortDescriptionColumns()[0]) {
+			details = &d
+			break
+		}
+	}
+
+	if details != nil {
```
jochil, line 363: WDYT about an else case, informing about a finding that is not part of our `error.json`... might be interesting for the user and us as well

pkg/finding/finding.go
```diff
@@ -306,3 +324,86 @@ func LoadFinding(projectDir, findingName string) (*Finding, error) {
...
+		// if the finding is an OWASP, add more details
+		if details.OwaspDetails != nil {
+			moreDetails.OwaspDetails = &ExternalDetail{
+				ID:          details.OwaspDetails.ID,
+				Name:        details.OwaspDetails.Name,
+				Description: details.OwaspDetails.Description,
+			}
+		}
+
+		// if thhe score is still 0, then the finding is not in the
```
jochil, line 395:
```suggestion
		// if the score is still 0, then the finding is not in the
```

pkg/finding/finding.go
```diff
@@ -306,3 +324,86 @@ func LoadFinding(projectDir, findingName string) (*Finding, error) {
...
+		// error-details file, so use the finding's name and score
+		if moreDetails.Severity.Score == 0 {
```
jochil, line 397: couldn't this be handle in an else branch of `details != nil`?

jochil (commented):
thanks for the adjustments... IMO it would help for the reviewing to squash/revise these adjustments into the original commits

pkg/finding/finding.go
MarkusZoppelt: I added a `log.Info` statement

pkg/finding/finding.go
MarkusZoppelt: Oh wow, you're right. I changed the type of the details struct earlier so they weren't directly compatible.

Now they are:
`moreDetails = *details`

pkg/finding/finding.go
MarkusZoppelt: Done

jochil (approved):

PR #615, opened 2023-02-08
fix API authentication bug

MarkusZoppelt:
### Motivation & Context
There was a bug when trying to authenticate against a server without any projects for the user.
This was related to a specific quirk in the `ListProjects` function.

### Description
This PR checks for that specific behavior.
In the (near) future, we should replace the call to `ListProjects` entirely with calls to the `/v2/is-authenticated` endpoint.

Closes [CLI-885](https://code-intelligence.atlassian.net/browse/CLI-885)

[CLI-885]: https://code-intelligence.atlassian.net/browse/CLI-885?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/api/project.go
```diff
@@ -67,7 +67,9 @@ func (client *APIClient) ListProjects(token string) ([]*Project, error) {
 	var projects []*Project
 	err = json.Unmarshal(objmap["projects"], &projects)
 	if err != nil {
-		return nil, errors.WithStack(err)
+		// If the projects field is not present, it means there are no projects
+		// so we return an empty list of projects and no error.
+		return []*Project{}, nil
```
jochil, line 72: There can be other errors when unmarshaling we will not see anymore... I suggest doing something like this
```
if _ , ok := objmap["projects"]; !ok {
  return []*Project{}, nil
} 
err = json.Unmarshal(objmap["projects"], &projects)
if err != nil {
  return errors.WithStack(err)
}
```

MarkusZoppelt: Done, thanks!

jochil (approved):

PR #620, opened 2023-02-10
feat(bundle): use openjdk docker image for maven/gradle

JackuB:
Under `cifuzz bundle` command the `openjdk` Docker tag is now used for Maven and Gradle build systems.

This brings with a change to the `docker-image` option under `cifuzz bundle` command.
- Default is set to empty string, because the "default" (automatic) value depends on the build system being targetted
- Help text was updated to reflect that. Hopefully in a clear way, and no one will start using the `cifuzz bundle --docker-image` without any value 😶


jochil (approved):
nice, thanks

PR #621, opened 2023-02-10
fix(cmake): improve debug information to avoid setting `CMAKE_BUILD_TYPE`

jochil:
### Motivation & Context
Instead of setting a different build type (do get better debugging support), `cifuzz` should build the fuzz targets directly with the needed flags.

### Description
* Reverting a previous commit, handling `CMAKE_BUILD_TYPE` when passing it to cmake via command line
* Using `-O0` when building in CMake, too


tools/cmake/modules/cifuzz-functions.cmake
```diff
@@ -49,6 +49,8 @@ function(enable_fuzz_testing)
           # Undefine NDEBUG, which is explicitly defined by the RelWithDebInfo CMake configuration, so that asserts are
           # kept.
           -UNDEBUG
+          # Do optimizations which don't harm debugging
```
fmeum, line 52:
```suggestion
          # Limit to optimizations which don't harm debugging
```
At least that's the intention since `-O2` might still appear on the command line before this flag. We have to confirm that it actually overrides that flag in the expected way.

fmeum (approved):

PR #622, opened 2023-02-13
feat: support multiple `@FuzzTest` in single JVM class

MarkusZoppelt:
### Motivation & Context
cifuzz needs to handle Java JUnit fuzz tests with multiple [valid @FuzzTest annotations](https://github.com/CodeIntelligenceTesting/jazzer/blob/main/examples/junit/src/test/java/com/example/ValidFuzzTests.java).

### Description
Enables support for multiple target methods in single JVM classes.
E.g. `cifuzz run com.example.FuzzTestCase::myFuzzTest` runs the fuzz test for the specified target method `myFuzzTest`

Closes [CLI-825](https://code-intelligence.atlassian.net/browse/CLI-825)

[CLI-825]: https://code-intelligence.atlassian.net/browse/CLI-825?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/runner/libfuzzer/libfuzzer_runner.go
```diff
@@ -41,6 +41,7 @@ type RunnerOptions struct {
 	EngineArgs         []string
 	EnvVars            []string
 	FuzzTarget         string
+	TargetMethod       string
```
jochil, line 44: do we actually need this for libfuzzer? I am not aware of us supporting this for c/cpp based tests... or at least it is out of scope for this task

internal/cmd/run/run.go
```diff
@@ -239,6 +240,13 @@ depends on the build system configured for the project.
 				return cmdutils.WrapSilentError(err)
 			}
 
+			// check if the fuzz test is a method of a class
+			if strings.Contains(args[0], "::") {
+				fuzzTest := strings.Split(args[0], "::")[0]
+				opts.targetMethod = strings.Split(args[0], "::")[1]
+				args[0] = fuzzTest
```
jochil, line 247: I am not sure about this... fiddling around with the `args` variable seems a little bit transparent for the following code. Why do we need this? I guess the code after could just use `fuzzTest`

internal/cmd/run/run.go
```diff
@@ -239,6 +240,13 @@ depends on the build system configured for the project.
 				return cmdutils.WrapSilentError(err)
 			}
 
+			// check if the fuzz test is a method of a class
+			if strings.Contains(args[0], "::") {
+				fuzzTest := strings.Split(args[0], "::")[0]
+				opts.targetMethod = strings.Split(args[0], "::")[1]
```
jochil, line 246:
```suggestion
				split := strings.Split(args[0], "::")
				fuzzTest, testMethod  := split[0], split[1]
```
I think this would be more readable and avoids the second time splitting the string

internal/cmdutils/jazzer.go
```diff
@@ -26,6 +26,29 @@ func JazzerGeneratedCorpus(targetClass string, projectDir string) string {
...
+func GetTargetMethodsFromJVMFuzzTestFile(path string) ([]string, error) {
+	bytes, err := os.ReadFile(path)
+	if err != nil {
+		return nil, errors.WithStack(err)
+	}
+
+	var targetMethods []string
+
+	// Define a regular expression pattern to match @FuzzTest annotations
+	fuzzTestRegex := regexp.MustCompile(`@FuzzTest\s+(\w+)\s+(\w+)\s*\(`)
```
jochil, line 38: we already have a regex for this... same file, `jazzerFuzzTestRegex`

jochil (commented):
just some early feedback :slightly_smiling_face: 
Are you aware of supporting this for the autocompletion as well?

internal/cmdutils/jazzer.go
MarkusZoppelt: True, but the global var is only useful for determining _**if**_ the source file contains the expression. The new expression can be used to extract the method name from the next line.

internal/cmd/run/run.go
```diff
@@ -240,6 +241,13 @@ depends on the build system configured for the project.
 				return cmdutils.WrapSilentError(err)
 			}
 
+			// Check if the fuzz test is a method of a class
+			// And remove method from fuzz test argument
+			if strings.Contains(args[0], "::") {
```
bertschneider, line 246: Extract `::` into a constant, as it's used many times? But that's not really go idiomatic, isn't it?

internal/cmdutils/jazzer.go
```diff
@@ -26,6 +27,41 @@ func JazzerGeneratedCorpus(targetClass string, projectDir string) string {
...
+	// the @FuzzTest annotation
+	fuzzTestRegex := regexp.MustCompile(`@FuzzTest\s+(?P<prefix>\w*\s)*(?P<functionName>\w+)\s*\(`)
+
+	// Find all matches of the regular expression in the input string
+	// matches := fuzzTestRegex.FindAllStringSubmatch(string(bytes), -1)
+	matches, _ := regexutil.FindAllNamedGroupsMatches(fuzzTestRegex, string(bytes))
+
+	// Check if the file contains a fuzzerTestOneInput method
+	// and append it to the targetMethods slice if it does
+	fuzzerTestOneInputRegex := regexp.MustCompile(`\sfuzzerTestOneInput\(`)
```
bertschneider, line 51: There could be whitespace between the function name and `(`.

internal/cmdutils/jazzer.go
```diff
@@ -26,6 +27,41 @@ func JazzerGeneratedCorpus(targetClass string, projectDir string) string {
...
+	if err != nil {
+		return nil, errors.WithStack(err)
+	}
+
+	var targetMethods []string
+
+	// Define a regular expression pattern to match @FuzzTest annotations
+	// it will return the name of the fuzz test method in the line following
+	// the @FuzzTest annotation
+	fuzzTestRegex := regexp.MustCompile(`@FuzzTest\s+(?P<prefix>\w*\s)*(?P<functionName>\w+)\s*\(`)
```
bertschneider, line 43: - Great use of capturing groups confined by the annotation and parameter list, really like this solution :+1: 
- Is the `prefix` capture group name required? I couldn't find any usage and that confused me.
- There are situations in with `prefix` could include characters not matched by `\w`, e.g. `@FuzzTest @SomeOtherAnnotation public static void fuzz(byte[] data) {}`

bertschneider (commented):
Thx for picking up this suggestion! As I'm not familiar with the codebase, just a few general remarks regarding the regex.

internal/cmd/run/run.go
MarkusZoppelt: Personally, I would slightly prefer it inline, since it's already pretty short and helps readability.

internal/cmdutils/jazzer.go
MarkusZoppelt: Thanks! Will change to `\sfuzzerTestOneInput\s*\(`

internal/cmd/run/run.go
MarkusZoppelt: Unfortunately, I think we really need to change the `args[0]` element here. Otherwise we would need to check the element format in each of the functions we invoke with `args`.

internal/cmdutils/jazzer.go
MarkusZoppelt: In the future, we'll definitely need a more robust solution than a regex to extract AST elements, meta-annotations, etc. 

We should take a look at treesitter libraries/wrappers for Go or Java tools that can do this for us - as proposed by @fmeum 

I created a follow-up ticket for this.

jochil:
@MarkusZoppelt shouldn't we mentioned it somewhere in the docs (markdown or command line help)?

MarkusZoppelt:
> shouldn't we mentioned it somewhere in the docs (markdown or command line help)?

@jochil I added some info to `docs/Getting-Started.md`.
wdyt?

jochil (approved):
perfect, thanks

PR #624, opened 2023-02-14
Make install: handle control characters in zsh

JackuB:
Figured out why the make install was creating cifuzz completion script in a strange place - turns out my .zshrc emits a clear when it gets sourced. 

And our i staller script is parsing its output https://github.com/CodeIntelligenceTesting/cifuzz/blob/8869decf7aa3d3616ba846a5ef650225c58da124/cmd/installer/installer.go#L360 but is not stripping control characters, so the path for autocomplete also contained an invisible control character. Leading to this path:

```
\x1b[H\x1b[2J/Users/jakubmikulas/.oh-my-zsh/custom/plugins/zsh-syntax-highlighting
```

PROTIP: printing these characters is achieved with:
```golang
fmt.Printf("%+q\n", sr) // print back the unicode
```

screenshot_2023-02-14_at_10 30 39


cmd/installer/installer.go
```diff
@@ -364,7 +365,12 @@ func installZshCompletionScript(installDir string) error {
 	if err != nil {
 		return errors.WithStack(err)
 	}
-	fpath := strings.TrimSpace(string(out))
+	// There could be an extra output, like control characters in the output
+	// This regex captures everything before the first slash
+	// - "?" in .*? makes it non-greedy
+	re := regexp.MustCompile("^.*?[^/]*")
+	fpath := re.ReplaceAllLiteralString(string(out), "")
```
MarkusZoppelt, line 372: Is there also a fix that does not require regex?

JackuB: I tried and I haven't found it.
Seems like no native go function will strip it. Using `unicode` and filtering letters won't do, because there are valid/printable/ascii letters like H & J in the string as well ¯\\\_(ツ)\_/¯

MarkusZoppelt (approved):

PR #625, opened 2023-02-14
fix: display origin URL parameter correctly

MarkusZoppelt:
URL parameters get garbled when passed as an argument to `url.JoinPath()` so we add the parameter(s) after the function call with `url.RawQuery = values.Encode()`


JackuB (approved):
I'd be happier if this used a query string builder instead of a string concat, but approving, since it gets the job done

MarkusZoppelt:
> I'd be happier if this used a query string builder instead of a string concat

@JackuB I changed it to use `url.Values{}`. Do you think this is better?

internal/cmd/remote-run/remote_run.go
```diff
@@ -318,13 +318,25 @@ func (c *runRemoteCmd) run() error {
...
 			return err
 		}
+		path = strings.TrimPrefix(path, "https://")
+		path = strings.TrimPrefix(path, "http://")
+
+		values := url.Values{}
+		values.Add("origin", "cli")
+
+		url := url.URL{
+			Scheme:   "https",
```
JackuB, line 332: Upgrading the scheme to HTTPS is a well-intentioned behavior

> The road to hell is paved with good intentions, but I'd recommend avoiding it

Would `url.Parse(c.opts.Server)` instead work here? Then you should be able to add a query param there and serialize the URL again, without affecting anything else in the URL (that could be set by the user to something strange)

MarkusZoppelt: Thanks, I found a simpler way :)

PR #631, opened 2023-02-20
add example for `jazzer.js`

MarkusZoppelt:
### Motivation & Context
This is the first step in preparation for the jazzer.js integration.

### Description
For the brave:
Example can be started and tested with the following commands inside the `examples/nodejs` directory:

Install dependencies:
    
    npm install --save-dev @jazzer.js/core @jazzer.js/jest-runner

Fuzzing run:
    
    npm run test

Coverage:

    npm run coverage


examples/javascript/package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "javascript-example",
+  "scripts": {
+    "test": "jest",
+    "fuzz": "JAZZER_FUZZ=1 jest",
+    "coverage": "jest --coverage"
+  },
+  "devDependencies": {
+    "@jazzer.js/jest-runner": "^1.3.0"
```
kyakdan, line 9:
```suggestion
    "@jazzer.js/core": "^1.3.0",
    "@jazzer.js/jest-runner": "^1.3.0",
    "@types/jest": "^29.4.0",
    "jest": "^29.4.3"
```
To avoid IDE warnings about unresolved jest functions and the Jazzer.js core package not being installed (the example uses the fuzzed data provider from the core package)

examples/javascript/FuzzTestCase.fuzz.js
```diff
@@ -0,0 +1,17 @@
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const { FuzzedDataProvider } = require("@jazzer.js/core");
+const exploreMe = require("./ExploreMe");
+
+/**
+ * @param { Buffer } fuzzerInputData
```
kyakdan, line 6: You usualy put this docstring right before the function with this parameter. This can be confusing since it is before the Jest's `describe` bock

examples/javascript/ExploreMe.js
```diff
@@ -0,0 +1,11 @@
+module.exports = function exploreMe(a, b, c) {
```
kyakdan, line 1: A docstring here with parameter types would be helpful

examples/javascript/.gitignore
```diff
@@ -0,0 +1,130 @@
+# Logs
```
kyakdan, line 1: This is quite long. It would be better to clean this file and limit it to the uses cases in the example project

kyakdan (commented):
Greate to see Jazzer.js supported by cifuzz!

examples/nodejs/FuzzTestCase.fuzz.js
```diff
@@ -0,0 +1,14 @@
+// eslint-disable-next-line @typescript-eslint/no-var-requires
```
jochil, line 1: as there is no eslint in this project we can drop this, or?

examples/nodejs/FuzzTestCase.fuzz.js
```diff
@@ -0,0 +1,14 @@
...
+const { FuzzedDataProvider } = require("@jazzer.js/core");
+const exploreMe = require("./ExploreMe");
+
+describe("My describe", () => {
+    it.fuzz("My fuzz test", (jazzerBuffer) => {
```
jochil, line 6:
```suggestion
    test.fuzz("My fuzz test", jazzerBuffer => {
```
`it` is usually used in jest to form a sentence like `it("should be return the square root of a", ()=>{ ... })` 
As there is no expected outcome when running fuzz tests we can just use `test.fuzz` which seems a better fit semantically 

Also the parentheses on `(jazzerBuffer)` are not necessary with only one parameter

examples/nodejs/FuzzTestCase.fuzz.js
```diff
@@ -0,0 +1,14 @@
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const { FuzzedDataProvider } = require("@jazzer.js/core");
+const exploreMe = require("./ExploreMe");
+
+describe("My describe", () => {
```
jochil, line 5:
```suggestion
describe("My fuzz tests", () => {
```
or just remove it... it is not necessary

examples/nodejs/.gitignore
```diff
@@ -0,0 +1,6 @@
+*.log
+
+coverage
+*.lcov
+
+node_modules/
```
jochil, line 6: we should add `.cifuzz-corpus` as well and maybe even `FuzzTestCase.fuzz` if we want it more convenient when working with the example

MarkusZoppelt: Good idea 👍🏼

jochil (approved):

PR #632, opened 2023-02-20
add `init` for jazzer.js

MarkusZoppelt:
### Motivation & Context
We start jazzer.js integration with the first usable command: `init`.
This command will detect JavaScript projects and print instructions on how to proceed.

### Description
This PR adds instructions and support for `cifuzz init` for JavaScript projects.


internal/config/config.go
```diff
@@ -19,16 +19,18 @@ import (
...
 
 const (
-	BuildSystemBazel  string = "bazel"
-	BuildSystemCMake  string = "cmake"
-	BuildSystemMaven  string = "maven"
-	BuildSystemGradle string = "gradle"
-	BuildSystemOther  string = "other"
+	BuildSystemBazel      string = "bazel"
+	BuildSystemCMake      string = "cmake"
+	BuildSystemJavaScript string = "javascript"
```
mlsandnerCI, line 24: Is JavaScript really the build system in this case?

MarkusZoppelt: What do you think if we either call it `nodejs` or do two types of "build systems":

* `node-npm` and
* `node-yarn`

mlsandnerCI: Both options sound fine to me but I'm not that into the topic to know if it makes sense to separate the two 🤔

MarkusZoppelt: cc @kyakdan @bertschneider

jochil: I gave this a thought too when looking at the example PR... npm and yarn are not build systems, they are package managers and they are so close to eachother that distinguishing them in our code does not make sense in my opinion (all ends up in the same `package.json`).
Build systems in the JS ecosystem are stuff like webpack or esbuild but they are not relevant for what we are doing here.
Another possibility would have been nodejs, as we are using JVM at some places.... also feels odd, but might the best option here.

Typescript/Javascript should be something we only use when creating the stubs

internal/cmd/init/instructions/javascript
```diff
@@ -0,0 +1,35 @@
+To use jazzer.js, add a dev-dependency to @jazzer.js/jest-runner to your
+project. To do so, execute the following command:
+
+    npm install --save-dev @jazzer.js/jest-runner
```
jochil, line 4: we should add the `yarn add -D` command as well

internal/cmd/init/instructions/javascript
```diff
@@ -0,0 +1,35 @@
...
+
+For Jest to pick up the custom fuzz test runner, it has to be added to the Jest
+configuration in the package.json or jest.config.js:
```
jochil, line 7: I think we should go with showing an example of `jest.config.js`  as this is usually the preferred way to configure jest

internal/config/config.go
MarkusZoppelt: @jochil What do you suggest?
To leave it at `examples/javascript` or change it to `examples/nodejs`?

MarkusZoppelt: I guess we can also rename 
`BuildSystemJavaScript` -> `BuildSysyemNode`

JackuB: I'm voting on keeping it as vague as `BuildSystemJavaScript` - it captures the intent.

And, most likely, we might have a better naming idea once we get to the `cifuzz bundle` for JS. We don't know whether we will be distinguishing them by build systems, JS-TS(-CS?!), npm/yarn/pnpm/..., es-modules/rollup/webpack/...?

JackuB: Aaaalso, this discussion might be rendered irrelevant if we rework the feature detection mechanism and abandon this BuildSystem switch https://code-intelligence.atlassian.net/browse/CLI-893

internal/cmd/init/instructions/javascript
```diff
@@ -0,0 +1,59 @@
...
+          {
+            "runner": "@jazzer.js/jest-runner",
+            "displayName": {
+              "name": "Jazzer.js",
+              "color": "cyan"
+            },
+            "testMatch": ["<rootDir>/**/*.fuzz.js"]
+          }
+        ]
+      }
```
jochil, line 38: I would remove this... it makes the instructions longer and we should not encourage this at all. just `jest.config.js`

internal/cmd/init/instructions/javascript
```diff
@@ -0,0 +1,59 @@
...
+     	{
+     	  "runner": "@jazzer.js/jest-runner",
+     	  "displayName": {
+            "name": "Jazzer.js",
+     		"color": "cyan"
+     	  },
+          "testMatch": ["<rootDir>/**/*.fuzz.js"]
+     	}
+      ]
+    };
```
jochil, line 57: indentation is a little bit messed up here

internal/config/config.go
jochil: I think I would prefer `BuildSystemNodejs` or `BuildSystemNpm`... the thing is that we expose `build-system` to the user (eg. in the cifuzz.yaml). And IMO 'Javascript" is the one that might be more confusing.

MarkusZoppelt: Done. Renamed to `BuildSystemNodeJS`

internal/cmd/init/instructions/javascript
```diff
@@ -0,0 +1,44 @@
...
+    {
+      "scripts": {
+        "test": "jest",
+        "fuzz": "JAZZER_FUZZ=1 jest",
+        "coverage": "jest --coverage"
+      },
+      "devDependencies": {
+        "@jazzer.js/core": "^1.3.0",
+        "@jazzer.js/jest-runner": "^1.3.0",
+        "jest": "^29.4.3"
```
jochil: This `devDependencies` block not necessary as it will be added by the commands above

I even think we can remove the whole `package.json` part.... the `test` and `coverage` script is something the user should have from his jest setup and `fuzz` is making `cifuzz run` somehow obsolete

Maybe we can just go with
"In order to work with your jest setup you have to add the following lines to your `jest.config.js`" 
or "To integrate fuzz testing into your jest setup you have to add the following lines to your `jest.config.js`" 

and then only the "jest" part

MarkusZoppelt: I agree about the `devDependencies` block but I am unsure about removing `scripts`... As long as we don't finally know how we'll do the `cifuzz run` command, it would be nice to at least see what 'jazzer.js + jest' already provide.

We can always remove that block later, once `cifuzz run` is ready.

jochil: I am pretty sure we will run `npx jazzer FuzzTarget` or `JAZZER_FUZZ=1 jest --some-args`... there are no other options.

> We can always remove that block later, once cifuzz run is ready.
We  have to remember that... and I would prefer a small, concise instruction block

jochil (approved):

PR #633, opened 2023-02-22
docs: Troubleshooting cifuzz headers not found

mlsandnerCI:
### Motivation & Context
A customer ran into a problem where the cifuzz headers could not be found when running a fuzz test but discovered that it worked when including it directly in the `CMakeLists.txt`. Since we could not discover what the cause of this behavior is, we decided to add a troubleshooting page with the workaround.

### Description
A new `Troubleshooting` page was created with the workaround included.


MarkusZoppelt:
If this is important information, we *could* also add it to our README.md, e.g., before [Contributing](https://github.com/CodeIntelligenceTesting/cifuzz#contributing).

docs/Getting-Started.md
```diff
@@ -95,3 +95,8 @@ and Linkedin.
 Also, watch [Going Beyond Unit Testing | How to Uncover Blind Spots in your
 Java Code with Fuzzing](https://www.youtube.com/watch?v=8yECb-p3cQI) on
 YouTube.
+
+## Troubleshooting
+
+If problems occur when installing or running cifuzz, you can check the [Troubleshooting](Troubleshooting.md)
+for possible solutions.
```
MarkusZoppelt, line 102:
```suggestion
If you encounter problems installing or running cifuzz, you can check [Troubleshooting](Troubleshooting.md)
for possible solutions.
```

MarkusZoppelt (approved):

jochil:
I think it is not necessary to put in into the readme or contributing guide as it usually should cover only edge cases

jochil:
> > I think it is not necessary to put in into the readme or contributing guide as it usually should cover only edge cases
> 
> So no link to it anywhere?

"Getting started" was ok for me... but I guess the way it is done, it is even ok in the readme.

PR #635, opened 2023-02-22
feat(installer): wait for 'Enter' after installation on Windows

MarkusZoppelt:
### Motivation & Context
When running on Windows, the installer is run in a new console window and the installer closes immediately and users won't see the installation notes.

### Description
Now, Windows users need to press `Enter` to close the window.
Result: installer cmd window doesn’t close automatically on Windows.


florianGla (approved):
LGTM

mlsandnerCI (approved):

PR #636, opened 2023-02-23
feat(cmake): add dependency arguments to add_fuzz_test function

florianGla:
... this makes it possible to add fuzz test dependencies right in the add_fuzz_test function, eg:

add_fuzz_test(my_fuzz_test
    SOURCES my_fuzz_test.cpp
    DEPENDENCIES ${FUZZ_TEST_DEPENDENCIES}
    INCLUDE_DIRS ${FUZZ_TEST_INCLUDE_DIRS}
)

These proposed changes are provided by a customer and do not affect backwards compatibility for existing projects.


florianGla:
This is a small convenience feature, which is currently not discoverable by the user. I am not sure if we should add a more complicated c++ example project or open a new section in our docs.

jochil:
IMO we should keep the examples simple. Maybe we can mention it here: https://github.com/CodeIntelligenceTesting/cifuzz/blob/main/docs/How-To-Write-A-Fuzz-Test.md#cmake

PR #637, opened 2023-02-23
Persist selected project choice

MarkusZoppelt:
### Motivation & Context
It is a nice QoL improvement to offer users the possibility to persist their choice after selecting a project during finding upload instead of just printing an echo statement.

### Description
The introduced change will ask users after they have selected a project during `cifuzz run` if they want to persist their choice. If they consent, we will append the server and project name to their cifuzz.yaml file.


jochil:
Isn't this something we should do for `remote-run` as well?

pkg/dialog/dialog.go
```diff
@@ -95,3 +97,33 @@ func ReadSecret(message string, file *os.File) (string, error) {
...
+			return errors.WithStack(err)
+		}
+		defer f.Close()
+
+		_, err = f.WriteString(text)
+		if err != nil {
+			return errors.WithStack(err)
+		}
+		log.Notef("Your choice has been persisted in cifuzz.yaml.")
+	}
```
jochil, line 127: It would be nice to separate the actual changing from the yaml file from the dialog itself. But for now I think this is fine

jochil (approved):

pkg/dialog/dialog.go
MarkusZoppelt: I'd say we extract a config-updating function once we have multiple use cases / occurences where we want to update the config for users.

PR #642, opened 2023-02-24
refactor: add output path as pflag

mlsandnerCI:
### Motivation & Context
I noticed when writing a test for the bundle command that the `OutputPath` Flag behaved differently than the other flags e.g. `ProjectDir` and will be overwritten if it is not set as an argument for the command but in an bundler options struct.

### Description
`OutputPath` is now a flag configured like the others to achieve the same behavior with every flag. Nothing has been changed in the way a user can use this flag.


internal/cmdutils/pflag.go
```diff
@@ -154,6 +154,14 @@ func AddInteractiveFlag(cmd *cobra.Command) func() {
 	}
 }
 
+func AddOutputPathFlag(cmd *cobra.Command) func() {
+	cmd.Flags().StringP("output", "o", "",
+		"Optional path of the created bundle artifact or coverage report.")
```
jochil, line 159:
```suggestion
		"Optional path of the file(s) created by this command  (eg. coverage report, bundle artifact, ...)
```
wdyt? This way we don't have to change the message when using it in another command.

jochil (approved):

jochil:
I think we should close this then and not merge it... I just saw that we made `output` available to the `cifuzz.yaml` (the mapstructure tag)` and this IMO doesn't make sense. You can not define one output for all commands

PR #645, opened 2023-02-27
Add email contact and docs link to issue template

JackuB:
This PR is a part of our work on introducing a more defined contribution model for external users.

Using GH's issue config file: https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository

However, `mailto:` links are not allowed, but you can get around with a link shortener.


jochil:
> However, mailto: links are not allowed, but you can get around with a link shortener.

Can you add this as a comment in the yml file, too?

MarkusZoppelt (changes requested):
```
    url: https://tinyurl.com/nkxuwh9v
```

Link shorteners leave a very bad taste in my mouth.
Users cannot see where the link is ultimately going.

I would rather have the email link not clickable than a shortener link.

JackuB:
> I would rather have the email link not clickable than a shortener link.

That is pretty much the state. Text in the issue contains the email, user is able to copy-paste it and not click the link.

jochil (approved):

MarkusZoppelt (approved):

Marcono1234:
Are GitHub issues not supposed to be used at all for this repository, not even for bug reports? Because that is what the issue templates and the README might currently suggest.

Using e-mail seems quite inconvenient and non-transparent for reporting issues to me. And I would assume you will also get duplicate reports because there is no way for users to search for existing reports.

PR #647, opened 2023-02-27
feat(finding): print more details if available

MarkusZoppelt:
Prints more external details (if available) in the output of the `finding` command.


internal/cmd/finding/finding_cmd.go
```diff
@@ -136,6 +164,68 @@ func (cmd *findingCmd) printFinding(f *finding.Finding) error {
...
+		pterm.Println(pterm.Blue("Description:"))
+		fmt.Println(f.MoreDetails.Description)
+	}
+	if f.MoreDetails.Mitigation != "" {
+		pterm.Println(pterm.Blue("\nMitigation:"))
+		fmt.Println(f.MoreDetails.Mitigation)
+	}
+}
+
+func getColorFunctionForSeverity(severity float32) func(a ...interface{}) string {
```
jochil, line 222: WDYT about something like a `PrintSeverity` function... in that case we can reuse it in multiple places and you don't have to call two functions with somehow the same parameter. So instead of
```
colorFunc := getColorFunctionForSeverity(f.MoreDetails.Severity.Score)
colorFunc(fmt.Sprintf("%.1f", f.MoreDetails.Severity.Score)),
```
just
```
PrintServerity(f.MoreDetails.Severity.Score)
```

JackuB: The former is more flexible and allows you to color other things with severity color - e.g. the whole line in a table output. I don't think either is better, so unless you feel strongly about one, I'd be okay keeping it.

jochil:
> One thing I noticed (but might not be relevant for now): When I have a finding found while not having the `error-details.json` it keeps missing the additional info even after adding the json file. I understand why... but maybe we should update our findings as soon as the user has access to the file

@MarkusZoppelt some comment on this?

PR #648, opened 2023-02-27
chore: add JVM finding dirs to .gitignore

MarkusZoppelt:


jochil:
wouldn't this be useful for maven too? It should create the findings in the same location

MarkusZoppelt:
> wouldn't this be useful for maven too? It should create the findings in the same location

@jochil Ah, I was wondering why it was not necessary for maven: We had a line in the root `.gitignore` that included that directory. I moved it to the maven example.

jochil (approved):

PR #651, opened 2023-02-28
Token creation / SaaS connection messaging

HenrichN:
### Motivation & Context
Add A/B testing capabilities for messaging while connecting to our SaaS.

### Description
Different messages are included and depending on the message shown different UTM parameters are set for the token creation URL


internal/cmd/run/run.go
```diff
@@ -987,6 +987,56 @@ Authenticate with the CI Fuzz Server (%s) to get more insights.`, server)
...
+func showServerConnectionMessage(server string) *url.Values {
+	messageA := "Do you want to persist your findings?\n" + "Authenticate with the CI Fuzz Server " + server + " to get more insights."
+	messageB := `Code Intelligence provides you a full history for all your findings and allows you to postpone work on findings and completely ignore them to keep the output clean and focused.
+
+	With a free authentication you receive detailed information and solution tips for each findings in your console.
+	
+	All your finding data stay only with us at Code Intelligence and will never be shared
+	
+	Do you want to continue with authentication now?`
+
```
MarkusZoppelt, line 999: Thank you for contribution!

We usually use backticks (`) with linebreaks after < 80 chars for long strings that we show to users so they get printed on smaller terminal sessions.

E.g., long string messages should look like this (note no tabs after newlines):

```go
	messageA := fmt.Sprintf(`Do you want to persist your findings?
Authenticate with the CI Fuzz Server %s to get more insights.`, server)
	
    messageB := `Code Intelligence provides you a full history for all your findings and allows
you to postpone work on findings and completely ignore them to keep the output
clean and focused.
[...]
```

However, I propose that we store the messages in a list. That way, we get better readability and can access the messages easily by index number coming from the number generator:

```suggestion
	messages := []string{
		fmt.Sprintf(`Do you want to persist your findings?
Authenticate with the CI Fuzz Server %s to get more insights.`, server),

		`Code Intelligence provides you a full history for all your findings and allows
you to postpone work on findings and completely ignore them to keep the output
clean and focused.

With a free authentication you receive detailed information and solution tips for
each findings in your console.
All your finding data stay only with us at Code Intelligence and will never be shared.
	
Do you want to continue with authentication now?`,

	}

```

Later: e.g.:

```go
n := computeStablePseudoRandomNumber()
log.Notef(messages[n])
```

internal/cmd/run/run.go
```diff
@@ -987,6 +987,56 @@ Authenticate with the CI Fuzz Server (%s) to get more insights.`, server)
...
+	additionalParams := url.Values{}
+	if computeStablePseudoRandomNumber() == 0 {
+		additionalParams.Add("utm_campaign", "A")
+		log.Notef(messageA)
+	} else {
+		additionalParams.Add("utm_campaign", "B")
+		log.Notef(messageB)
+	}
```
MarkusZoppelt, line 1007: With the messages list (map?), maybe this can also be removed to a single line where you access the item directly.

internal/cmd/run/run.go
```diff
@@ -987,6 +987,56 @@ Authenticate with the CI Fuzz Server (%s) to get more insights.`, server)
...
+// To avoid that a user sees a different message each time
+// we compute a stable "random" number
+func computeStablePseudoRandomNumber() int {
+	// Path name for the executable
+	path, err := os.Executable()
+
+	if err != nil {
+		// As a fallback, return "true" random number
+		return rand.Intn(2)
+	}
```
MarkusZoppelt, line 1021: Not a fan of recovering here. In the (highly unlikely) case that this should fail, we should handle and report the error to the caller.

internal/cmd/run/run.go
```diff
@@ -987,6 +987,56 @@ Authenticate with the CI Fuzz Server (%s) to get more insights.`, server)
...
+
+	file, err := os.Stat(path)
+	if err != nil {
+		// As a fallback, return "true" random number
+		return rand.Intn(2)
+	}
```
MarkusZoppelt, line 1027: see other comment

internal/cmd/run/run.go
```diff
@@ -987,6 +987,56 @@ Authenticate with the CI Fuzz Server (%s) to get more insights.`, server)
 	return wishToAuthenticate == "Yes", nil
 }
 
+func showServerConnectionMessage(server string) *url.Values {
```
jochil, line 990: I think the `showServerConnectionMessage` and `computeStablePseudoRandomNumber` functions would be better placed in a separate package... as they might be used in other commands and for not polluting the run command even more. WDYT about `dialog` or `messaging`

internal/cmd/run/run.go
```diff
@@ -987,6 +987,56 @@ Authenticate with the CI Fuzz Server (%s) to get more insights.`, server)
...
+
+	// We are using minute instead of hour
+	// to avoid a "bias" for people trying
+	// things out late in the evening
+	minute := file.ModTime().Minute()
+	if minute < 30 {
+		return 0
+	} else {
+		return 1
+	}
```
MarkusZoppelt, line 1037: Maybe we can change this to a modulo check, e.g. `something % len(messages)`.

That way, we can handle more than two messages.

internal/cmd/run/run.go
```diff
@@ -987,6 +987,56 @@ Authenticate with the CI Fuzz Server (%s) to get more insights.`, server)
...
+		additionalParams.Add("utm_campaign", "B")
+		log.Notef(messageB)
+	}
+
+	return &additionalParams
+}
+
+// To avoid that a user sees a different message each time
+// we compute a stable "random" number
+func computeStablePseudoRandomNumber() int {
```
MarkusZoppelt, line 1014: The name of this function in convoluted and contradictory.

How about `pickNumberForMessagingIndex`?

MarkusZoppelt (changes requested):

jochil (commented):
What do you think about another approach: Instead of distributing the functionality over multiple places we can have a package that assigns the user to a group (and saves it internally) and then asking the package for messages and additional url params?
It might be less messy if we continue with this kind of stuff

MarkusZoppelt (approved):

pkg/messaging/messaging.go
```diff
@@ -0,0 +1,70 @@
...
+		{
+			"utm_source":   []string{"cli"},
+			"utm_campaign": []string{"login-message"},
+			"utm_term":     []string{"a"}},
+		{
+			"utm_source":   []string{"cli"},
+			"utm_campaign": []string{"login-message"},
+			"utm_term":     []string{"b"},
+		},
+	}
```
jochil, line 40: can we put this in one struct to not rely on some indexes being in the right order? Something like
```
[]struct{
  message string
  urlParams []url.Values
}{
{...},
{...}
}
```

pkg/messaging/messaging.go
```diff
@@ -0,0 +1,70 @@
...
+		messageIndex = 0
+	}
+
+	log.Notef(messages[messageIndex])
+	return &additionalParams[messageIndex]
+}
+
+// To avoid that a user sees a different message each time
+// we compute a stable "random" number
+func PickNumberForMessagingIndex(numberOfMessages int) (int, error) {
```
jochil, line 53: can be private as it isn't use outside the packge

jochil (approved):

PR #652, opened 2023-02-28
feature(bundle): '--add' flag to add external files to archive

mlsandnerCI:
### Motivation & Context
Customer had a complicated setup because they needed to include specific files in their bundle, so they bundled, unpacked the bundled, added files and bundled it again.

### Description
A new flag `--add` has been implemented for the `bundle` command to allow adding external files or directories to the bundle. They will be stored in the `work_dir` directory in the archive. A target path can be specified with `--add <source>;<target>`


internal/bundler/bundler.go
```diff
@@ -186,3 +214,29 @@ func prepareSeeds(seedCorpusDirs []string, archiveSeedsDir string, archiveWriter
...
+			strings.Count(arg, ":") == 1 &&
+			filepath.VolumeName(arg) != ""
+
+	// If there is no target dir specified, the target path will be "work_dir"
+	// Includes windows path with volume name e.g. C:\path\to\file or C:path\to\file
+	if !strings.Contains(arg, ":") || isWindowsPathWithoutTarget {
+		return []string{arg, filepath.Join(archiveWorkDirPath, filepath.Base(arg))}, nil
+	}
+
+	regexPattern := regexp.MustCompile(`(?P<source>\S+[/|\\]?\S+):(?P<target>\S+)`)
```
JackuB, line 230: Would a `strings.Split(arg, ":")` work here instead of the regex? This would also allow you to lose the `Count :` check above.

jochil: I replaced the regex with a split and also changed the separator to `;` to keep it simple, especially for the windows cases

internal/bundler/bundler.go
```diff
@@ -186,3 +210,29 @@ func prepareSeeds(seedCorpusDirs []string, archiveSeedsDir string, archiveWriter
...
+	if len(parts) == 1 {
+		// if there is no ; separator just use the work_dir
+		// handles "test.txt"
+		source = parts[0]
+		target = filepath.Join(archiveWorkDirPath, filepath.Base(arg))
+	} else {
+		// handles test.txt;test2.txt
+		source = parts[0]
+		target = parts[1]
+	}
```
MarkusZoppelt, line 227: This also handles malformed arguments, e.g.

`test.txt;foo;test2.txt`

Maybe we should check with an `else if len(parts) == 2` and show an error to inform that the argument is probably malformed or misstyped.

internal/bundler/bundler.go
```diff
@@ -186,3 +210,29 @@ func prepareSeeds(seedCorpusDirs []string, archiveSeedsDir string, archiveWriter
...
+
+	if source == "" || target == "" {
+		return "", "", errors.New("could not parse '--add' argument")
+	}
+
+	if filepath.IsAbs(target) {
+		return "", "", errors.New("target path of '--add' argument should not be absolute, but relative to the created archive")
+	}
```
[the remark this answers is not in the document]
jochil: As it is still an error message I would keep it short. What do you think about: `"when using '--add source;target, target has to be a relative path"`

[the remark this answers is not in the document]
MarkusZoppelt: Perfect 👍🏼

internal/bundler/bundler.go
jochil: I will add it to the existing check `errors.New("could not parse '--add' argument")`

MarkusZoppelt (approved):

internal/bundler/bundler.go
mlsandnerCI: I didn't think about changing the separator, I thought that was fixed 😅 Definitely would have been the easier way than trying to find a regex that works...

PR #658, opened 2023-03-02
Improve CI/CD detection

JackuB:
This PR creates a new package for detecting (and identifying) CI/CD runs. This will help us with not requiring interactive terminal when not appropriate.
Also replaces all occurences of CI envvar checks with this package.


MarkusZoppelt (approved):
I like it!

pkg/detect_ci/detect_ci.go
```diff
@@ -0,0 +1,191 @@
...
+The ISC License
+
+Copyright (c) npm, Inc.
+
+Permission to use, copy, modify, and/or distribute this software for any
+purpose with or without fee is hereby granted, provided that the above
+copyright notice and this permission notice appear in all copies.
+*/
+
+package detect_ci
```
jochil, line 13: We should avoid package names with underscore... see: https://google.github.io/styleguide/go/decisions#underscores

pkg/detect_ci/detect_ci.go
```diff
@@ -0,0 +1,191 @@
...
+
+import "os"
+
+func GetCIName() string {
```
jochil, line 17: According to the styleguide it can be `CIName`: https://google.github.io/styleguide/go/decisions#getters

pkg/detect_ci/detect_ci.go
```diff
@@ -0,0 +1,191 @@
...
+	if os.Getenv("GERRIT_PROJECT") != "" {
```
jochil, line 18: Semantically it would be more correct to use `LookupEnv` (https://pkg.go.dev/os#LookupEnv) as you want to check for the presence of the env, not that the value is actually empty

pkg/detect_ci/detect_ci_test.go
```diff
@@ -0,0 +1,30 @@
...
+// Some samples
+var envvarsAndExpected = [][]string{
+	{"CI", "custom"},
+	{"GERRIT_PROJECT", "gerrit"},
+	{"GITHUB_ACTIONS", "github-actions"},
+	{"TRAVIS", "travis-ci"},
+}
+
+func TestIsCI(t *testing.T) {
+	os.Clearenv()
```
jochil, line 19: you can use `t.Setenv` (https://pkg.go.dev/testing#B.Setenv) in tests... it restores the original values of the environment after the test

pkg/detect_ci/detect_ci.go
JackuB: @jochil we do have a few of those, is it worth changing there as well?

Screenshot 2023-03-07 at 13 12 48

pkg/detect_ci/detect_ci.go
JackuB: I don't see that it would improve things:

> LookupEnv retrieves the value of the environment variable named
by the key. If the variable is present in the environment the
value (which may be empty) is returned and the boolean is true.
Otherwise the returned value will be empty and the boolean will
be false.

LookupEnv is also returning a value AND bool. Since we don't care about a difference between empty or unset values it would probably only made the code longer.

pkg/detect_ci/detect_ci_test.go
JackuB: Hmm, but it will still keep the env values that were present. So running this particular test, without `Clearenv` on say CircleCI would fail?

jochil: I am not 100% sure... maybe trying it out. 
Messing around with the OS env can cause side effects, especially when running the tests in parallel

pkg/detect_ci/detect_ci.go
jochil: > Since we don't care about a difference between empty or unset values...

Are you sure that all the CI systems are on the same page? If one is just setting the env without a value we would not detect it properly

pkg/detect_ci/detect_ci.go
jochil: IMO we should do... I think @MarkusZoppelt has already his editor/linter complaining about these names

PR #659, opened 2023-03-06
feat: print `*` characters as visual feedback when users enter secrets

MarkusZoppelt:
### Motivation & Context
Previously, we used `term.ReadPassword()` for reading secrets from stdin. However, that function gives no visual feedback to users.

### Description
This PR adds a function that prints `*` characters as a visual indicator, e.g., when users paster their API access token into the terminal.


JackuB (approved):
Tried breaking it locally and it works fine.

Limitation that is _possibly_ fine is that it's ignoring control characters now. Anything from backspace to `ctrl+c`

MarkusZoppelt:
> Limitation that is _possibly_ fine is that it's ignoring control characters now. Anything from backspace to `ctrl+c`

@JackuB Ugh, that is unfortunately a pretty strong limitation... we should catch at least `Bksp` and `Ctrl-C` manually... or even rethink if we **_really_** want this feature.

It also introduces a lot of hacky logic that might break in some cases.

**How do you feel about not hiding/obfuscating API keys at all?**

What do @jochil, @fmeum and @HenrichN think?

jochil:
> > Limitation that is _possibly_ fine is that it's ignoring control characters now. Anything from backspace to `ctrl+c`
> 
> @JackuB Ugh, that is unfortunately a pretty strong limitation... we should catch at least `Bksp` and `Ctrl-C` manually... or even rethink if we **_really_** want this feature.
> 
> It also introduces a lot of hacky logic that might break in some cases.
> 
> **How do you feel about not hiding/obfuscating API keys at all?**
> 
> What do @jochil, @fmeum and @HenrichN think? 

IMO not worth it. Handling all the possible cases is a tedious task and it is more worst when a terminal don't behave like expected than not showing "typing" signs (which is also default behavior)

HenrichN:
@MarkusZoppelt It is a feature which has been requested multiple times from different points of view (Product and Customer Success) so I would like to create a solution. I also cross-checked the Github CLI, it also prints * when a token is inserted. I would propose we add a manual detection for Backspace and Control-C. If we get constant complaints and it looks like that we spent too much time in maintaining it, we can remove the future again. But I would like to see first how much effort it really creates for us. On another note: Did you get a chance to test this in a Windows Terminal and Power Shell?

MarkusZoppelt:
Thanks everyone for the help on this!

[cli/cli](https://github.com/cli/cli) is using https://github.com/go-survey/survey and that seems to work wonderfully!

jochil (approved):

PR #666, opened 2023-03-07
fix(ci): use bazel 6.0 as macos pipeline broke with the release of 6.1

jochil:
### Motivation & Context
Integration tests broke on macos with newest bazel version

### Description
use 6.0


.github/actions/vm_setup/action.yml
```diff
@@ -10,7 +10,7 @@ runs:
       - name: Setup Go environment
         uses: actions/setup-go@v3
         with:
-          go-version: "1.19"
+          go-version: "^1.19"
```
JackuB, line 13: Is this intended?

JackuB (approved):

.github/actions/vm_setup/action.yml
jochil: no, good catch :+1:

PR #668, opened 2023-03-07
Windows/use clang instead of msvc

florianGla:
Running fuzz tests shall work with CMake + clang/clang-cl. MSVC is replaced completely with clang-cl.
This is still work in progress...


internal/build/build.go
```diff
@@ -40,7 +40,7 @@ func CommonBuildEnv() ([]string, error) {
 		return nil, err
 	}
 
-	// On Windows, our preferred compiler is MSVC, which can't easily be run
+	// On Windows, our preferred compiler is clang-cl, which can't easily be run
 	// from an arbitrary terminal as it requires about a dozen environment
 	// variables to be set correctly. Thus, we assume users to run cifuzz from
 	// a developer command prompt anyway and thus don't need to set the
```
mlsandnerCI, line 46: Is this complete comment still relevant with clang-cl?

florianGla: Which part would you leave out? You have to run clang-cl from the a developer command prompt, because many environment variables are required, which are not set in the regular command prompt. There is a executable, which sets those variables, but it is located in the visual studio folder and we need at least the environment variable, which stores the location to the visual studio folder

PR #669, opened 2023-03-07
Renaming packages to avoid underscores

JackuB:
Renaming cifuzz packages to avoid underscores in names https://go.dev/blog/package-names


pkg/finding/finding.go
```diff
@@ -71,10 +71,10 @@ type ErrorDetails struct {
...
 
 const (
-	SeverityLevel_CRITICAL SeverityLevel = "CRITICAL"
-	SeverityLevel_HIGH     SeverityLevel = "HIGH"
-	SeverityLevel_MEDIUM   SeverityLevel = "MEDIUM"
-	SeverityLevel_LOW      SeverityLevel = "LOW"
+	SeverityLevelCritical SeverityLevel = "CRITICAL"
+	SeverityLevelHigh     SeverityLevel = "HIGH"
+	SeverityLevelMedium   SeverityLevel = "MEDIUM"
+	SeverityLevelLow      SeverityLevel = "LOW"
```
MarkusZoppelt, line 77: same here

pkg/runner/libfuzzer/integrationtest/asan_test.go
```diff
@@ -1,4 +1,4 @@
-package integration_tests
+package integrationtest
```
MarkusZoppelt, line 1: not sure if this (and everywhere else) shouldn't be `integrationstests` (plural)

JackuB: I found that Golang _prefers_ singular name for packages:
- https://go.dev/blog/package-names : Golang using simple and singular packages
- https://rakyll.org/style-packages/

It makes sense when you think about it as "all those things have a single goal of testing an integration".

MarkusZoppelt: but then even the directory (top-level) will be called `integrationtest` and it looks kinda strange, since there are numerous inside...

mlsandnerCI (approved):

pkg/finding/finding.go
```diff
@@ -50,11 +50,11 @@ type ErrorType string
 // to parse JSON-marshalled reports as protobuf reports which use an
 // enum for this field.
 const (
-	ErrorType_UNKNOWN_ERROR     ErrorType = "UNKNOWN_ERROR"
-	ErrorType_COMPILATION_ERROR ErrorType = "COMPILATION_ERROR"
-	ErrorType_CRASH             ErrorType = "CRASH"
-	ErrorType_WARNING           ErrorType = "WARNING"
-	ErrorType_RUNTIME_ERROR     ErrorType = "RUNTIME_ERROR"
```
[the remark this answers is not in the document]
JackuB: Brought back Error into an error of ErrorType Unknown Error

PR #654, opened 2023-03-01
Introduce additional E2E tests

JackuB:
This PR introduces a new level of tests. I'm calling the tests E2E for now, but maybe we need a more fitting name to distinguish them from Integration/System testing. **Is there a better name? I'm thinking something with behavior/matrix/zoo could be nice. I'm open to suggestions?**
Goal of these tests will be to capture wide variety of scenarios and provide high-level assertions.

These tests will currently run at the release time, and as a nightly verification.
We will be expanding these as we will work on cifuzz - first major addition will be Jazzer.js coverage, beyond unit and integration tests.

PR #670, opened 2023-03-07
CI: update pipelines to incorporate review comments

JackuB:
Follow ups from https://github.com/CodeIntelligenceTesting/cifuzz/pull/654


mlsandnerCI (changes requested):

JackuB:
@mlsandnerCI I had to add an extra step, because we don't want to run `make install`, when testing a `cifuzz` executable coming from a GH Release.

JackuB:
@mlsandnerCI update the e2e runner naming to use `testCase` and `testCaseRun` instead of the "subtests" lingo

mlsandnerCI (approved):

jochil (approved):
thanks for incorporating my late feedback :slightly_smiling_face:

PR #671, opened 2023-03-07
fix(fmt): exclude .git from fmt checks to ignore go files

jochil:
### Motivation & Context
`make fmt` was checking go files located in `.git` which can contain invalid/broken go files fetched from other revisions

### Description
just ignore `.git` for fmt checks


MarkusZoppelt:
Nice, thank you for discovering a fix!

Is there also an option so that we can re-enable running `go fmt ./...` without the error?

My guess is there is not, since we would need to rewrite git history... right?

MarkusZoppelt (approved):

jochil:
> Nice, thank you for discovering a fix!
> 
> Is there also an option so that we can re-enable running `go fmt ./...` without the error?
> 
> My guess is there is not, since we would need to rewrite git history... right?

IIRC `go fmt` is not working with `./...` as it is using package identifier like `go fmt code-intelligence.com/...` whereas `gofmt ./...` would work but runs into the same problems as right now. If you want you use `go fmt` you usually use it like `go fmt $(go list ./...)` which should also avoid the problem

... but nevertheless we are not using any of them directly, as we go wth https://github.com/incu6us/goimports-reviser to have the imports sorted nicely.

PR #672, opened 2023-03-08
Fix installer builder on Windows

florianGla:
We use a symbolic link from our share directory to our cmake modules. The builder copies all files and folders in our share directory and resolves symbolic links. On Windows this symbolic link is converted to a text file containing the path to the folder it points to. In that case we have to read this file and use this path instead.


fmeum (commented):
@florianGla Windows has support for symlinks in Developer Mode, which I think we can assume  contributors have enabled (if we document this requirement). Git should check out symlinks as such on Windows with symlinks enabled and if it doesn't, I think we should preferably fix the git config instead.

florianGla:
> @florianGla Windows has support for symlinks in Developer Mode, which I think we can assume contributors have enabled (if we document this requirement). Git should check out symlinks as such on Windows with symlinks enabled and if it doesn't, I think we should preferably fix the git config instead.

Thank you, I didn't know that. I will try it on my machine and update the contributing docs (and git config) instead.

PR #673, opened 2023-03-08
fix(bundle): repair `--add ` and add regression test

jochil:
### Motivation & Context
the `--add` flag is broken due to a swap of parameters

### Description
This change fixes the `--add` flag for bundle by swapping parameters and adding a regresssion/integration test to avoid that error in the future


internal/bundler/bundler.go
```diff
@@ -116,16 +116,16 @@ func (b *Bundler) Bundle() error {
...
+			err = archiveWriter.WriteDir(target, source)
 			if err != nil {
 				return err
 			}
 		} else {
-			err = archiveWriter.WriteFile(source, target)
+			err = archiveWriter.WriteFile(target, source)
 			if err != nil {
 				return err
 			}
```
MarkusZoppelt, line 131: Maybe add a comment to explain the order of the arguments

MarkusZoppelt (approved):

internal/bundler/bundler.go
jochil: It is how the `WriteFile` function was written... so it should be clear from the function declaration

PR #675, opened 2023-03-08
Add git symlink support on Windows

florianGla:
Symlinks on Windows are not resolved by default by git. This causes problems in our installer, because our cmake scripts are included via a symlink. This issue is resolved if “Developer Mode“ is activated on Windows and symlink support is activated for git. We should explain this in our contributing guidelines.


CONTRIBUTING.md
```diff
@@ -85,3 +85,20 @@ cd examples/cmake
...
+We use symlinks in our repository. Developer mode needs to be
+activated in order to create symlinks without admin privileges.
+It can be activated in the system settings under:
+```
+Settings > Update & Security > For developers > Developer Mode
+```
+Moreover either symlink support has to be activated in your git
+config with:
+```
+git config --global core.symlinks true
```
fmeum, line 99: Instead of asking users to do this, couldn't we add this setting to a `.gitconfig` file in the root of the repo, leaving just the first paragraph of this?

florianGla: Done

jochil (approved):

PR #678, opened 2023-03-10
docs: Recommend `target_*` CMake functions with `add_fuzz_test`

fmeum:
### Motivation & Context

The `target_*` functions are idiomatic CMake and can be used with `add_fuzz_test` just like with `add_executable`. We should recommend this standard style in our docs.

### Description
Recommend `target_*` over `add_fuzz_test`'s custom keyword arguments.


fmeum:
@jochil I don't think we have any dedicated reference documentation for the CMake macros yet. It seems like we would mostly need to add it to document the new arguments which at the same time we don't really want to advertise broadly. Not sure how to handle that part, so I only updated the tutorial for now.

jochil:
> @jochil I don't think we have any dedicated reference documentation for the CMake macros yet. It seems like we would mostly need to add it to document the new arguments which at the same time we don't really want to advertise broadly. Not sure how to handle that part, so I only updated the tutorial for now.

Could you think about something else we can document regarding the CMake integration. I am afraid that we will not document this feature unless we do it now

jochil:
@fmeum How should we continue here?

fmeum:
@jochil Sorry, I only just got back to this PR.

In order to unblock it, I wrote up a short reference doc for the CMake integration. Please take a look.

jochil:
@fmeum thanks... looks very good :+1:

PR #684, opened 2023-03-15
feat(Makefile): use $(RM) to support remove on all platforms

MarkusZoppelt:
Sets a OS-dependent var for `RM` so that we can use the correct remove command on all supported platforms via `make`.


jochil:
For me this starts looking somehow hacky... maybe we should think about changing our task runner soon (must not be part of this task). First things coming to mind would be a python script or just plain go

MarkusZoppelt:
> For me this starts looking somehow hacky... maybe we should think about changing our task runner soon (must not be part of this task). First things coming to mind would be a python script or just plain go

It is hacky, but is pretty commonly used when building portable Makefiles.
See also ["Working with Nonportable Tools"](https://www.oreilly.com/openbook/make3/book/ch07.pdf).

jochil (approved):
Ok, I see

PR #688, opened 2023-03-16
feat: small improvements on verbose output

jochil:
### Motivation & Context
When getting logs from clients/users it is sometimes unclear which cifuzz version was used

### Description
We show the version on top of our command output


internal/build/cmake/cmake.go
```diff
@@ -398,13 +399,16 @@ func (b *Builder) fuzzTestsInfoDir() (string, error) {
...
 		return fuzzTestsDir, nil
 	}
 	// The path to the info file for multi-configuration CMake generators (e.g. MSBuild).
 	fuzzTestsDir = filepath.Join(buildDir, cmakeBuildConfiguration, ".cifuzz", "fuzz_tests")
+	log.Debugf("Searching for test info file in %s", fuzzTestsDir)
 	if fileutil.IsDir(fuzzTestsDir) {
 		return fuzzTestsDir, nil
 	}
-	return "", os.ErrNotExist
+	log.Warn("Not found test info file")
```
MarkusZoppelt, line 412:
```suggestion
	log.Warn("Did not find test info file")
```

MarkusZoppelt (approved):

PR #694, opened 2023-03-20
init: ask users if they want to use CI App

MarkusZoppelt:
- feat(init): ask users to configure CI app (server) for their project
- chore(examples): update cifuzz.yaml


internal/config/config.go
```diff
@@ -64,6 +65,14 @@ func CreateProjectConfig(configDir string) (string, error) {
...
 	}
 
+	// write server if set
+	if server != "" {
+		_, err = fmt.Fprintf(f, "server: %s\n", server)
+		if err != nil {
+			return "", errors.WithStack(err)
+		}
+	}
+
```
JackuB, line 75: Since yesterday, I'm still thinking whether we are accidentally introducing an unexpected behavior here. I'm still going back and forth whether this is something we actually want or not.

Should we considering always filling the `server` key by a default value, instead of omitting it?

jochil: this is actually what I took from yesterday... there shouldn't be the case where the server is not defined

HenrichN: But we would still need some way to deal with cifuzz.yamls which are older and where the field is not filled, right? Here I would propose that we still use app.code-intelligence.com in case the value is not set.

jochil: yes, `app.code-intelligence.com` should be the default but set and handled as early in the code as possible not in a function deep down in the logic

MarkusZoppelt:
We reached consensus that not setting `server` at all will always default to `app.code-intelligence.com` and custom values need to be present in a project `cifuzz.yaml` to override that setting. This makes the changes from this PR no longer necessary.

PR #695, opened 2023-03-20
Use gotestfmt to improve test output

JackuB:
Added https://github.com/GoTestTools/gotestfmt to pipelines to improve test output readability from this:

screenshot_2023-03-20_at_18 20 06

to much better:

Screenshot 2023-03-22 at 12 31 47

The `-hide all` option makes test silent when they are passing and only print errors. Otherwise you need to scroll the list of tests to find failures:

Screenshot 2023-03-22 at 12 31 57


.github/workflows/pipeline_tests.yml
```diff
@@ -78,6 +78,9 @@ jobs:
         with:
           os: ${{ matrix.os }}
 
+      - name: Set up gotestfmt
+        run: go install github.com/gotesttools/gotestfmt/v2/cmd/gotestfmt@latest
```
MarkusZoppelt, line 82: Should we use the GitHub action instead?

https://github.com/GoTestTools/gotestfmt#github-actions

PR #696, opened 2023-03-20
Use clang instead of MSVC on Windows

florianGla:
msvc is replaced by clang-cl

- there is still one msvc comment in the launcher and one code block for msvc in the replayer, which I think should not be deleted
- instrumenting shared libraries is not currently supported


.github/actions/vm_setup/action.yml
```diff
@@ -56,9 +56,9 @@ runs:
           brew install lcov
         shell: bash
 
-      # Add cl.exe to the PATH on Windows for the replayer integration test.
+      # Action for configuring Developer Command Prompt for Microsoft Visual C++
       # Pinned to the commit of the v1.10.0 tag on 2022-05-05.
-      - name: Setup - Windows
+      - name: ClangCL - Windows
         uses: ilammy/msvc-dev-cmd@7315a94840631165970262a99c72cfb48a65d25d
```
jochil, line 62: we are still using the msvc actions? This seems a little bit confusing

florianGla: msvc-dev-cmd
[GitHub Action](https://github.com/features/actions) for configuring Developer Command Prompt for Microsoft Visual C++.

jochil: wdyt about adding a comment to avoid confusion?

internal/build/cmake/cmake.go
```diff
@@ -148,12 +148,15 @@ func (b *Builder) Configure() error {
 	}
 
 	cacheArgs := []string{
-		"-DCMAKE_BUILD_TYPE=" + cmakeBuildConfiguration,
 		"-DCIFUZZ_ENGINE=libfuzzer",
 		"-DCIFUZZ_SANITIZERS=" + strings.Join(b.Sanitizers, ";"),
 		"-DCIFUZZ_TESTING:BOOL=ON",
 	}
 	if runtime.GOOS != "windows" {
+		// CMAKE_BUILD_TYPE is ignored when building with Visual Studio.
```
fmeum, line 158: "with MSBuild" may be more precise here.

internal/build/cmake/cmake.go
```diff
@@ -148,12 +148,15 @@ func (b *Builder) Configure() error {
...
+		// The config has to be only specified in the build step with
```
fmeum, line 159: Do you really mean "has to be only" or rather "only has to be", that is, does the build fail if we specify this or is it just ignored?

tools/cmake/cmake_test.go
```diff
@@ -265,6 +259,7 @@ func TestIntegration_FuzzTestBinaryLaunchesCIFuzz(t *testing.T) {
 	cmakeVariables := map[string]string{
 		"CIFUZZ_ENGINE":     "libfuzzer",
 		"CIFUZZ_SANITIZERS": "address",
+		"CIFUZZ_TESTING":    "ON",
```
fmeum, line 262: Is this change needed? It does seem somewhat out of place.

tools/cmake/modules/cifuzz-functions.cmake
```diff
@@ -35,8 +35,12 @@ function(enable_fuzz_testing)
           # TODO(fmeum): Remove once ASan has been stabilized and clang_rt.asan_dynamic-x86_64.dll is available in the
           #  default PATH, e.g. in System32.
           # https://stackoverflow.com/a/66532115/297261
-          /MTd
+          /MT
```
fmeum, line 38: We used to use `/MTd` to link the debug version of the standard library. Does this no longer work with clang-cl or is there some other reason not to do it?

tools/cmake/modules/cifuzz-functions.cmake
```diff
@@ -35,8 +35,12 @@ function(enable_fuzz_testing)
...
+          -D_DISABLE_VECTOR_ANNOTATION
```
fmeum, line 39: Could you add a comment to this option that explain why it is needed and whether it could be removed at some point?

tools/cmake/modules/cifuzz-functions.cmake
```diff
@@ -35,8 +35,12 @@ function(enable_fuzz_testing)
...
       )
+
+      # Add Visual Studio directory for clang runtime libraries to link directories 
+      link_directories("$ENV{VSINSTALLDIR}\\VC\\Tools\\Llvm\\x64\\lib\\clang\\${CMAKE_CXX_COMPILER_VERSION}\\lib\\windows")
```
fmeum, line 43: Is this relative directory stable across relevant VS versions?

tools/cmake/modules/cifuzz-functions.cmake
```diff
@@ -59,47 +63,45 @@ function(enable_fuzz_testing)
...
+          -fsanitize-address-use-after-scope
+          # Disable source fortification, which is currently not supported
+          # in combination with ASan, see https://github.com/google/sanitizers/issues/247
+          # Note that this does not override a user-specified -D_FORTIFY_SOURCE,
+          # because the flags we add here come before user-specified flags.
+          # It's still useful to disable source fortification which the
+          # toolchain or distribution might have enabled by default.
+          -U_FORTIFY_SOURCE
+      )
+      if(NOT WIN32)
```
fmeum, line 88: This if condition can be flipped.

tools/cmake/modules/cifuzz-functions.cmake
```diff
@@ -231,12 +232,22 @@ function(add_fuzz_test name)
...
-    if((NOT MSVC) AND ((address IN_LIST CIFUZZ_SANITIZERS) OR (undefined IN_LIST CIFUZZ_SANITIZERS)))
+    if((address IN_LIST CIFUZZ_SANITIZERS) OR (undefined IN_LIST CIFUZZ_SANITIZERS))
       # The macOS linker doesn't support --wrap, so we fall back to a different strategy that doesn't require any linker
       # flags.
       # See src/dumper.c for details.
       if(NOT APPLE)
-        target_link_options("${name}" PRIVATE -Wl,--wrap=__sanitizer_set_death_callback)
+        if(WIN32)
+          # ALTERNATENAME replaces the function call to _ZN11__sanitizer6PrintfEPKcz with ?Printf@fuzzer@@YAXPEBDZZ.
+          # ?Printf@fuzzer@@YAXPEBDZZ is the mangled version of __sanitizer::Printf(const char *format, ...) on Windows,
```
fmeum, line 242: `...of fuzzer:Printf...`

tools/cmake/modules/cifuzz-functions.cmake
```diff
@@ -231,12 +232,22 @@ function(add_fuzz_test name)
...
     target_sources("${name}" PRIVATE "${_launcher_src}")
-    if((NOT MSVC) AND ((address IN_LIST CIFUZZ_SANITIZERS) OR (undefined IN_LIST CIFUZZ_SANITIZERS)))
+    if((address IN_LIST CIFUZZ_SANITIZERS) OR (undefined IN_LIST CIFUZZ_SANITIZERS))
       # The macOS linker doesn't support --wrap, so we fall back to a different strategy that doesn't require any linker
       # flags.
       # See src/dumper.c for details.
       if(NOT APPLE)
-        target_link_options("${name}" PRIVATE -Wl,--wrap=__sanitizer_set_death_callback)
+        if(WIN32)
+          # ALTERNATENAME replaces the function call to _ZN11__sanitizer6PrintfEPKcz with ?Printf@fuzzer@@YAXPEBDZZ.
```
fmeum, line 241: You could add that `_ZN11__sanitizer6PrintfEPKcz` is the gcc-style mangled name of `__sanitizer:Printf(const char *format, ...)`.

internal/build/cmake/cmake.go
```diff
@@ -123,7 +123,9 @@ func (b *Builder) BuildDir() (string, error) {
 				return "", errors.WithStack(err)
 			}
 		}
-		hashString := hex.EncodeToString(hash.Sum(nil))
+		// Use only the first 10 characters in order to prevent errors on
+		// Windows, which cannot handle long file paths.
+		hashString := hex.EncodeToString(hash.Sum(nil))[:10]
```
fmeum, line 128: 10 hex characters give you 40 bits of entropy. Since every character matters on Windows, how about using https://pkg.go.dev/encoding/base32 to cut this down to 8 characters?

fmeum (approved):
While I was a bit skeptical about the idea of ripping out MSVC completely, this PR provides considerable gains in terms of code and feature consistency across all three platforms. Well done and bye bye MSVC :-)

tools/cmake/modules/cifuzz-functions.cmake
florianGla: Yes, unfortunately it is not supported with clang-cl
`clang-cl : error : invalid argument '/MTd' not allowed with '-fsanitize=address'`

tools/cmake/modules/cifuzz-functions.cmake
florianGla: To the best of my knowledge this is stable. We use a different VS version in the github runners.

tools/cmake/modules/cifuzz-functions.cmake
florianGla: Done

tools/cmake/modules/cifuzz-functions.cmake
florianGla: Done

internal/build/cmake/cmake.go
florianGla: Done

PR #697, opened 2023-03-21
Windows clang coverage

florianGla:
Add support for clang-cl coverage on Windows


mlsandnerCI (approved):

internal/cmd/coverage/coverage_test.go
```diff
@@ -55,6 +60,10 @@ func TestClangMissing(t *testing.T) {
 }
 
 func TestCMakeMissing(t *testing.T) {
+	if runtime.GOOS == "windows" {
```
fmeum, line 63: Aren't we using `CMake` on Windows? It seems possible that users have VS installed but didn't install CMake with it.

internal/cmd/coverage/llvm/llvm.go
```diff
@@ -417,8 +417,18 @@ func (cov *CoverageGenerator) generateHTMLReport() (string, error) {
...
+
+	// genHTML is a perl script, which has to be started like "perl /path/to/genhtml args..." on Windows
+	if runtime.GOOS == "windows" {
+		perl, err := runfiles.Finder.PerlPath()
+		if err != nil {
+			return "", err
+		}
+		args = append([]string{genHTML}, args...)
+		cmd = exec.Command(perl, args...)
+	}
```
fmeum, line 430: Adding an `else` here and having just `var cmd exec.Cmd` above the if would make the flow more readable.

internal/cmd/coverage/llvm/llvm.go
```diff
@@ -528,7 +538,12 @@ func (cov *CoverageGenerator) indexedProfilePath() string {
 }
 
 func (cov *CoverageGenerator) executableName() string {
-	return filepath.Base(cov.buildResult.Executable)
+	executable := cov.buildResult.Executable
+	// Remove .exe file extension on Windows
```
fmeum, line 542: Could you explain why we want to do this? I could see Windows users expecting this extension to mentally classify something as an executable.

internal/cmd/run/run.go
```diff
@@ -697,15 +697,17 @@ func (c *runCmd) checkDependencies() error {
...
-			dependencies.CMake,
-		}
 		switch runtime.GOOS {
 		case "linux", "darwin":
-			deps = append(deps, dependencies.Clang)
-			deps = append(deps, dependencies.LLVMSymbolizer)
+			deps = []dependencies.Key{
+				dependencies.Clang,
+				dependencies.CMake,
+				dependencies.LLVMSymbolizer,
```
fmeum, line 705: Do you happen to know why we don't need the symbolizer on Windows?

fmeum (approved):

pkg/runfiles/finder.go
```diff
@@ -58,6 +58,22 @@ func (f RunfilesFinderImpl) LLVMSymbolizerPath() (string, error) {
 }
 
 func (f RunfilesFinderImpl) GenHTMLPath() (string, error) {
+	if runtime.GOOS == "windows" {
+		path := os.Getenv("path")
```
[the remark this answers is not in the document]
florianGla: On Windows exec.LookPath looks for files with specific file extensions (the default ones are {".com", ".exe", ".bat", ".cmd"}). Genhtml is a perl script without a file extension. I tried to implement a simplified version of exec.LookPath, which should be enough for our specific use case.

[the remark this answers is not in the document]
fmeum: Okay, sounds reasonable.

Could you pass `PATH` to `Getenv` instead of `path`?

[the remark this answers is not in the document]
florianGla: Done

internal/cmd/coverage/llvm/llvm.go
florianGla: If we don't remove the file extension, the html reports gets stored as `my_fuzz_test.exe/index.html` and the lcov report as `my_fuzz_test.exe.coverage.lcov`. I think it is ok to do the removal in `executableName`, but I am open for other suggestions.

internal/build/cmake/cmake.go
```diff
@@ -417,3 +423,20 @@ func (b *Builder) fuzzTestsInfoDir() (string, error) {
 	log.Warn("Did not find test info file")
 	return "", errors.WithStack(os.ErrNotExist)
 }
+
+func isSystemLibrary(dep string) bool {
+	wellKnownSystemLibraries := map[string][]*regexp.Regexp{
```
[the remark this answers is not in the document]
florianGla: Done

internal/cmd/coverage/llvm/llvm.go
fmeum: Ah, in that case removing it does make sense.

README.md
```diff
@@ -269,6 +271,7 @@ Run `cifuzz` in `Developer PowerShell for VS 2022` inside of `Windows Terminal`.
 **Windows**
 
 * C/C++ projects are only supported with CMake and fuzz tests cannot depend on shared libraries.
+* Continuous code coverage is not supported for C/C++ projects.
```
fmeum, line 274: Do you happen to know why? What kind of errors do we get with this?

internal/cmd/coverage/llvm/llvm.go
```diff
@@ -416,17 +416,22 @@ func (cov *CoverageGenerator) generateHTMLReport() (string, error) {
...
+		// "perl /path/to/genhtml args..." on Windows
+		args = append([]string{genHTML}, args...)
 		perl, err := runfiles.Finder.PerlPath()
 		if err != nil {
 			return "", err
 		}
-		args = append([]string{genHTML}, args...)
 		cmd = exec.Command(perl, args...)
+	} else {
+		// Prefix flag is not working on Windows
```
fmeum, line 432: Do you have an idea why it isn't working?

pkg/runfiles/finder.go
```diff
@@ -192,6 +193,32 @@ func (f RunfilesFinderImpl) findFollowSymlinks(relativePath string) (string, err
...
+		}
+
+		path := os.Getenv("PATH")
+		for _, dir := range filepath.SplitList(path) {
+			// Only look for llvm tools in visual studio path
+			if !strings.HasPrefix(dir, visualStudioPath) {
+				continue
+			}
+
+			path := filepath.Join(dir, name+".exe")
```
fmeum, line 209: This shadows the variable in the outer scope. Inline into the call on the next line to prevent this.

fmeum (approved):

README.md
florianGla: I have not tried to use it. The llvm doc mentions: `Support for Linux may be mostly complete but requires testing, and support for Windows may require more extensive changes: please get involved if you are interested in porting this feature.`
Moreover we would need we need to parse the symbols of the binary to check whether it has been built with the required compile-time flags (-mllvm -runtime-counter-relocation). I don't know how to do it on Windows.

fmeum: I think that the Go `debug` package would support this, but we could investigate that in a follow-up. That message seems as if it were slightly out of date.

internal/cmd/coverage/llvm/llvm.go
fmeum: I can't find that line in https://github.com/linux-test-project/lcov/blob/master/bin/genhtml, could you point to me to it?

README.md
florianGla: That is interesting. I will create a ticket for that.

internal/cmd/coverage/llvm/llvm.go
florianGla: We use the genhtml script, which is provided by the lcov chocolatey package. I compared both scripts and yours appears to be newer and contains many differences. I am happy to check if that one solves the issue.

florianGla: Chocolatey ships lcov 1.15 and the most recent version is 1.16. I haven't found a way to install 1.16 on Windows. Is it ok, if I add a comment to reevaluate the --prefix flag, once the chocolatey package got updated?

fmeum: Yes, that's fine, thanks for checking. I just want to ensure that every comment that says X doesn't work on Y mentions at least some condition under which we should reevaluate the state of X on Y.

PR #702, opened 2023-03-22
feat: integrate gradle plugin into cifuzz

jochil:
### Motivation & Context
Preparing the use of the new gradle plugin in cifuzz

### Description
- removing init script
- update instructions
- include plugin
- adding a simple tool for upgrading the gradle version in all places


examples/gradle-multi/testsuite/build.gradle.kts
```diff
@@ -0,0 +1,10 @@
+plugins {
+    id("java-library")
+    id("com.code-intelligence.cifuzz") version "1.0.0"
+}
```
MarkusZoppelt, line 4: I assume it is intentional, that the cifuzz plugin is only listed here instead of also being included in `examples/gradle-multi/app/build.gradle.kts`?

tools/updater/updater.go
MarkusZoppelt, line 1: I am curious how this updater works. Is there a way to test it?

MarkusZoppelt (approved):

examples/gradle-multi/testsuite/build.gradle.kts
jochil: yes this is intentional as the plugin only needs to be included in the package with the fuzz tests

tools/updater/updater.go
jochil: you can change the version number in this file and just run it with `go run tools/updater/updater.go`
I just want to try out if this kind of tool will actually help us before implementing it in a more "maintainable" way

PR #703, opened 2023-03-23
refactor: constants for libfuzzer and jazzer options

mlsandnerCI:
### Motivation & Context
To prevent misspelling mistakes and define possible options for jazzer and libfuzzer in place. This is more pre-work since we don't use the jazzer options in multiple places right now (what we expected when creating the ticket) but expect to do it later on.

### Description
Jazzer and Libfuzzer Flags have been set as constants and moved into a `options` package.


MarkusZoppelt:
I really like putting options in a dedicated package!

However, I think we can further improve the readability of using the options.
Especially since most of them are string that look like `--foo=` and need to be suffixed with an argument.

Maybe we can build a function that automatically sets the argument as a suffix. Otherwise, occurrences like `options.JazzerAutoFuzzFlag+r.AutofuzzTarget` look strange, at least imho. Readers might think: "why is this being concatenated?"

How about expressing it like:

    options.JazzerAutoFuzzFlag(r.AutofuzzTarget)

or something more generic like

    options.Make(JazzerAutoFuzzFlag, AutofuzzTarget)

wdyt?

MarkusZoppelt (approved):
LGTM 👍🏼

PR #708, opened 2023-03-24
fix(installer): do not print cifuzz version multiple times

jochil:
... only in verbose mode

image

### Motivation & Context
Before it was printed 3 time in a row


Makefile
```diff
@@ -75,7 +75,7 @@ deps/test:
 .PHONY: install
 install:
 	go run tools/builder/builder.go --version $(version)
-	go run -tags installer cmd/installer/installer.go
+	go run -tags installer cmd/installer/installer.go --verbose
```
MarkusZoppelt, line 78: Why `--verbose`?
`make install` is pretty chatty now:
Screenshot 2023-03-26 at 21 35 20

mlsandnerCI (approved):

PR #710, opened 2023-03-27
Validate that build system is supported

florianGla:
On Windows we currently do not support Bazel and "other" build systems. It should be checked if the build system is supported by the target os before starting the build process.


internal/config/config.go
```diff
@@ -36,6 +38,10 @@ var buildSystemTypes = []string{
 	BuildSystemOther,
 }
 
+var unsupportedBuildSystems = map[string][]string{
```
mlsandnerCI, line 41: Would it make more sense to have a list of supported build systems instead of unsupported ones? Generally I would say whitelisting is more "secure" than blacklisting.

florianGla: I can do that. I thought that it would be more repetitive to list all build systems for linux and macos.

florianGla: I have changed it. I think you were right.

jochil:
WDYT about having this check for all commands? Now it is only covering `run` and `coverage` but `bundle` and `remote-run` might need it as well and maybe it does even make sense for `init` and `create`... why setting up a project on a system where you can not build it?

florianGla:
> WDYT about having this check for all commands? Now it is only covering `run` and `coverage` but `bundle` and `remote-run` might need it as well and maybe it does even make sense for `init` and `create`... why setting up a project on a system where you can not build it?

Good idea. I will do that.

jochil:
> > WDYT about having this check for all commands? Now it is only covering `run` and `coverage` but `bundle` and `remote-run` might need it as well and maybe it does even make sense for `init` and `create`... why setting up a project on a system where you can not build it?
> 
> Good idea. I will do that.

Let's see what @HenrichN thinks about it too

jochil (approved):

PR #713, opened 2023-03-28
Recommend win terminal for better font rendering

florianGla:
The default terminal on Windows does not render font colors and glyphs correctly. We should recommend to install Windows Terminal (https://github.com/microsoft/terminal) for the best user experience.


README.md
```diff
@@ -1,4 +1,4 @@
-<div align="center">
+<div align="menter">
```
MarkusZoppelt, line 1: typo?

MarkusZoppelt (approved):

README.md
florianGla: Good catch :)

PR #714, opened 2023-03-28
improve user dialogs

MarkusZoppelt:


pkg/messaging/messaging.go
```diff
@@ -20,7 +19,8 @@ Authenticate with the CI Fuzz Server %s to get more insights.`, server),
 			additionalParams: url.Values{
 				"utm_source":   []string{"cli"},
 				"utm_campaign": []string{"login-message"},
-				"utm_term":     []string{"a"}},
+				"utm_term":     []string{"a"},
```
JackuB, line 22: Just a formatting change not relevant to the PR... but what is the meaning of the â̴͕̤͈̜̤̙̜̈͑̂̓̀̄̅͋̉̏̒̍͛̃͠ͅ?

JackuB (approved):

pkg/messaging/messaging.go
MarkusZoppelt: I removed the last line of the 'b' message because it is redundant with the message we show in the dialog.

MarkusZoppelt: I removed the formatting changes from the PR so that the change becomes more obvious.

PR #716, opened 2023-03-28
fix(run): only persist `project` in cifuzz.yaml after confirm

MarkusZoppelt:
Only persist `project` in cifuzz.yaml after user confirms their choice to persist after `cifuzz run|remote-run`


HenrichN (approved):
LGTM

PR #723, opened 2023-04-03
Provide context for messaging

HenrichN:
Enable showing different messages for a connection to CI App depending on the command.

### Motivation & Context
This change enables to tell the user more explicitly what the additional features are for each command.


pkg/messaging/messaging.go
```diff
@@ -8,22 +8,30 @@ import (
...
+
+func ShowServerConnectionMessage(server string, context MessagingContext) *url.Values {
+
+	messagesAndParams := map[string][]struct {
 		message          string
 		additionalParams url.Values
 	}{
-		{
-			message: fmt.Sprintf(`Do you want to persist your findings?
+		"Run": {
```
MarkusZoppelt, line 24: Since you already created the `MessagingContext` type, I would use it here:

```suggestion
	messagesAndParams := map[MessagingContext][]struct {
		message          string
		additionalParams url.Values
	}{
		Run: {
```

pkg/messaging/messaging.go
```diff
@@ -32,21 +40,49 @@ for each finding in your console.
...
+				additionalParams: url.Values{
+					"utm_source":   []string{"cli"},
+					"utm_campaign": []string{"finding-message"},
+					"utm_term":     []string{"b"},
+				},
 			},
 		},
 	}
 
+	messageAndParam, entryPresent := messagesAndParams[string(context)]
```
MarkusZoppelt, line 73: ... then we also don't need the cast here:

```suggestion
	messageAndParam, entryPresent := messagesAndParams[context]
```

pkg/messaging/messaging.go
```diff
@@ -32,21 +40,49 @@ for each finding in your console.
...
-				"utm_campaign": []string{"login-message"},
-				"utm_term":     []string{"b"},
+				additionalParams: url.Values{
+					"utm_source":   []string{"cli"},
+					"utm_campaign": []string{"run-message"},
+					"utm_term":     []string{"b"},
+				},
+			},
+		},
+		"Finding": {
```
MarkusZoppelt, line 50:
```suggestion
		Finding: {
```

pkg/messaging/messaging.go
```diff
@@ -8,22 +8,30 @@ import (
...
-func ShowServerConnectionMessage(server string) *url.Values {
+type MessagingContext string
 
-	messagesAndParams := []struct {
+const (
+	Run     MessagingContext = "Run"
+	Finding MessagingContext = "Finding"
+)
+
+func ShowServerConnectionMessage(server string, context MessagingContext) *url.Values {
```
MarkusZoppelt, line 18: This function gets pretty big, if we add new entries to the map in the future...

I would propose that we initialise the `messagesAndParams` map outside of the function to avoid initialising the map every time the function is called.

The "clean" new function can then be concise like this:

```go
func ShowServerConnectionMessage(server string, context MessagingContext) *url.Values {
	messageAndParam, entryPresent := messagesAndParams[context]

	if !entryPresent {
		return &url.Values{}
	}

	messageIndex, err := pickNumberForMessagingIndex(len(messageAndParam))
	if err != nil {
		messageIndex = 0
	}

	log.Notef(messageAndParam[messageIndex].message)
	return &messageAndParam[messageIndex].additionalParams
}
```

HenrichN:
@MarkusZoppelt Thanks for the helpful comments, I addressed all of them.

pkg/messaging/messaging.go
```diff
@@ -29,24 +34,53 @@ them to keep the output clean and focused.
...
-				"utm_campaign": []string{"login-message"},
+				"utm_campaign": []string{"run-message"},
+				"utm_term":     []string{"b"},
+			},
+		},
+	},
+	Finding: {
+		{
+			message: `Authenticate with CI Fuzz to get more insights
+on your findings and persist them for a full history.`,
```
MarkusZoppelt, line 50: I think we need to rephrase some of these to CI App, right?
```suggestion
			message: `Authenticate with CI App to get more insights
on your findings and persist them for a full history.`,
```

pkg/messaging/messaging.go
```diff
@@ -1,25 +1,30 @@
...
+var messagesAndParams = map[MessagingContext][]struct {
+	message          string
+	additionalParams url.Values
+}{
+	Run: {
 		{
-			message: fmt.Sprintf(`Do you want to persist your findings?
-Authenticate with the CI Fuzz Server %s to get more insights.`, server),
+			message: `Do you want to persist your findings?
+Authenticate with the CI Fuzz server to get more insights.`,
```
MarkusZoppelt, line 24:
```suggestion
Authenticate with CI App to get more insights.`,
```

MarkusZoppelt (approved):
Thank you!

pkg/messaging/messaging.go
HenrichN: 🙈 Good catch!

PR #724, opened 2023-04-03
fix: adjust regex for java version

mlsandnerCI:
### Motivation & Context
Our version extraction for java couldn't handle [the "new" version type scheme of java](https://openjdk.org/jeps/223) with four numbers (i.e. 18.0.0.1).
### Description
We adjusted the regex to accept the new format and added a test case for it (plus some minor refactoring).


pkg/dependencies/version.go
```diff
@@ -24,7 +24,7 @@ var (
 	clangRegex = regexp.MustCompile(`(?m)clang version (?P<version>\d+\.\d+(\.\d+)?)`)
 	cmakeRegex = regexp.MustCompile(`(?m)cmake version (?P<version>\d+\.\d+(\.\d+)?)`)
 	llvmRegex  = regexp.MustCompile(`(?m)LLVM version (?P<version>\d+\.\d+(\.\d+)?)`)
-	javaRegex  = regexp.MustCompile(`(?m)version \"(?P<version>\d+(\.\d+\.\d+)?)(_\d+)?\"`)
+	javaRegex  = regexp.MustCompile(`(?m)version "(?P<version>\d+(\.\d+\.\d+)*)\.*\d*?(_\d+)?"`)
```
jochil, line 27:
```suggestion
	javaRegex  = regexp.MustCompile(`(?m)version "(?P<version>\d+(\.\d+\.\d+)*)([_\.]\d+)?"`)
```
IMO this is easier to understand (as far as this is possible for regex)
So its read basically as `semver` optionally followed by `_number` or `.number`

jochil (approved):

pkg/dependencies/version.go
mlsandnerCI: Cool, thanks!

PR #725, opened 2023-04-04
update and improve config

MarkusZoppelt:
- fix: don't print a newline in installer note
- feat(examples): update cifuzz.yaml configs
- feat: add server and project flags to init cmd
- refactor: validate and normalise URLs in API func
- feat: update config with project
- feat(config): ensure project entry in config


jochil (changes requested):
When I run `cifuzz init` the created `cifuzz.yaml` contains `server: https://app.code-intelligence.com`.
IIRC  we agreed that the default config file should be an empty one (in the sense that there are no "active" fields) and that we are not explicitly defining the default value

Also I think the last two commits can be one

jochil:
> IIRC we agreed that the default config file should be an empty one (in the sense that there are no "active" fields) 

IMO we should have a test for this... as this PR would otherwise have introduced a regression

MarkusZoppelt:
@jochil Thank you for the feedback! I added your requested changes.

internal/cmd/init/init.go
```diff
@@ -56,20 +65,37 @@ func New() *cobra.Command {
...
 
 	log.Debugf("Creating config file in directory: %s", opts.Dir)
-	configpath, err := config.CreateProjectConfig(opts.Dir)
+
+	// if the user is using the default server, we don't want to save it in the config file
+	if strings.Contains(opts.Server, "app.code-intelligence.com") {
+		// an empty string will be interpreted as a comment in the config file
+		// in the next step
+		opts.Server = ""
+	}
```
jochil, line 104: can we move this to the `CreateProjectConfig` function... this way we can make sure the default value is never persisted to the config file, even when used in another command for example

MarkusZoppelt, line 104: Note: We do this because cmdutils.AddServerFlag automatically sets the default server, but we do not want to write it to the config.yaml

internal/config/config_test.go
```diff
@@ -47,6 +47,34 @@ func TestCreateProjectConfig(t *testing.T) {
 	assert.NoError(t, err)
 	assert.NotEmpty(t, content)
 	assert.Contains(t, string(content), "Configuration for")
+	assert.Contains(t, string(content), "#server: https://app.code-intelligence.com")
+	assert.Contains(t, string(content), "#project: ")
+}
+
+// Should create a config file with a commented out server if the server value
+// is empty
+func TestCreateProjectConfig_WithServerAndProject(t *testing.T) {
```
jochil, line 56: I was looking for a test covering the case that the `config.yaml` is technically empty (only comments) when not using `--server` and `--project`

MarkusZoppelt: It's the test above

jochil: sorry... I misread it. Thanks!

jochil (approved):

PR #726, opened 2023-04-04
Fix various spelling issues

HenrichN:
* FuzzTest -> Fuzz Test
* Checkout -> check out
* Link to docs.code-intelligence.com


MarkusZoppelt (approved):

integration-tests/cmake/cmake_test.go
```diff
@@ -333,7 +333,7 @@ func testRun(t *testing.T, cifuzzRunner *shared.CIFuzzRunner) {
 			},
 			{
 				SourceFile:  "src/parser/parser_fuzz_test.cpp",
-				Line:        30,
+				Line:        29,
```
florianGla, line 336: This is really strange. I don't know if we should change the expected stack frame. The changes introduced in this PR should not change the libfuzzer output.  Maybe we should investigate this issue?

HenrichN: That's a good idea because I just changed the expected stackframe in the tests  😂 Can you perhaps help out because I think I can't find the root cause.

florianGla: I'm sorry, now I get it. You made the stub one line shorter, which causes parser_fuzz_test.cpp to be on line shorter. So the line of the reported error has to be reduced by one.

PR #727, opened 2023-04-04
test(e2e): finding command

JackuB:
Introduces basic regression testing for a `cifuzz finding` command behavior. Especially for the difference of logged in / not logged in user.

Screenshot 2023-04-06 at 17 17 32


e2e-tests/e2e-test-runner.go
```diff
@@ -10,10 +10,15 @@ import (
...
 
 	"code-intelligence.com/cifuzz/integration-tests/shared"
+	"code-intelligence.com/cifuzz/internal/cmdutils/login"
+	"code-intelligence.com/cifuzz/internal/tokenstorage"
 	"code-intelligence.com/cifuzz/pkg/cicheck"
 	"code-intelligence.com/cifuzz/util/fileutil"
 )
 
+const ciServerToUseForE2ETests = "https://app.code-intelligence.com/"
+const envvarWithE2EUserToken = "E2E_TEST_CIFUZZ_API_TOKEN"
```
MarkusZoppelt, line 20:
```suggestion
const envVarWithE2EUserToken = "E2E_TEST_CIFUZZ_API_TOKEN"
```

e2e-tests/e2e-test-runner.go
```diff
@@ -10,10 +10,15 @@ import (
 	"testing"
 
 	"code-intelligence.com/cifuzz/integration-tests/shared"
+	"code-intelligence.com/cifuzz/internal/cmdutils/login"
+	"code-intelligence.com/cifuzz/internal/tokenstorage"
 	"code-intelligence.com/cifuzz/pkg/cicheck"
 	"code-intelligence.com/cifuzz/util/fileutil"
 )
 
+const ciServerToUseForE2ETests = "https://app.code-intelligence.com/"
```
MarkusZoppelt, line 19: I vote for using the staging server
```suggestion
const CIServerToUseForE2ETests = "https://app.staging.code-intelligence.com/"
```

e2e-tests/e2e-test-runner.go
```diff
@@ -84,6 +128,10 @@ func runTest(t *testing.T, testCase *TestCase) {
 
 	for index, testCaseRun := range testCaseRuns {
 		t.Run(fmt.Sprintf("[%d/%d] cifuzz %s %s", index+1, len(testCaseRuns), testCaseRun.command, testCaseRun.args), func(t *testing.T) {
+			fmt.Println("Running test:", testCase.Description)
+			fmt.Println("Command:", "cifuzz", testCaseRun.command, testCaseRun.args)
+			fmt.Println(" ")
```
MarkusZoppelt, line 133: you can leave the function call empty `fmt.Println()` if you want just a newline or append `\n` to the previous call

.github/workflows/pipeline_e2e_matrix_nightly.yml
```diff
@@ -4,6 +4,7 @@ on:
   schedule:
     - cron: "0 0 * * *"
   workflow_dispatch:
+  push: # WIP for testing pipelines
```
MarkusZoppelt, line 7: /reminder if this needs to be removed before merge

internal/cmd/finding/finding_cmd.go
```diff
@@ -62,6 +64,11 @@ func newWithOptions(opts *options) *cobra.Command {
 		},
 		RunE: func(c *cobra.Command, args []string) error {
 			opts.Interactive = viper.GetBool("interactive")
+			// Command should not be interactive when stdin is not a terminal.
+			// TODO: Should this be global? Set on a Viper level for all commands?
+			if !term.IsTerminal(int(os.Stdin.Fd())) {
+				opts.Interactive = false
+			}
```
MarkusZoppelt, line 71: For reference, this is how we do it in the `create` & `remoterun` cmd:

```go
if opts.Interactive {
    opts.Interactive = term.IsTerminal(int(os.Stdin.Fd())) && term.IsTerminal(int(os.Stdout.Fd()))
}
```

e2e-tests/e2e-test-runner.go
jochil: honestly I vote for not using any remote server at all... but if we use one, we should use `staging` and talk to the web team about it

e2e-tests/e2e-test-runner.go
```diff
@@ -44,7 +57,24 @@ type testCaseRunOptions struct {
...
 // RunTests Runs all test cases generated from the input combinations
 func RunTests(t *testing.T, testCases []TestCase) {
-	for _, testCase := range testCases {
+	// Convenience option for local testing. Grab local token, backup the existing one and restore it after the test.
+	if !cicheck.IsCIEnvironment() && os.Getenv(envvarWithE2EUserToken) == "" {
+		fmt.Println("E2E_TEST_CIFUZZ_API_TOKEN envvar is not set. Trying to use the default one, since this is not a CI/CD run.")
+		token := login.GetToken(ciServerToUseForE2ETests)
+		if token != "" {
+			fmt.Println("Found local token with login.GetToken, going to use it for the tests.")
+			os.Setenv(envvarWithE2EUserToken, token)
```
jochil, line 66: to minimize potential side effects using `t.Setenv`

e2e-tests/e2e-test-runner.go
```diff
@@ -44,7 +57,24 @@ type testCaseRunOptions struct {
...
+			tokenFilePath := tokenstorage.GetTokenFilePath()
+			if _, err := os.Stat(tokenFilePath); err == nil {
+				fmt.Println("Backing up existing access token file")
+				_ = os.Rename(tokenFilePath, tokenFilePath+".bak")
+				defer func() {
+					_ = os.Rename(tokenFilePath+".bak", tokenFilePath)
+				}()
```
jochil, line 73: using `t.Cleanup` would be a more idiomatic way of doing this

e2e-tests/e2e-test-runner.go
```diff
@@ -44,7 +57,24 @@ type testCaseRunOptions struct {
 
 // RunTests Runs all test cases generated from the input combinations
 func RunTests(t *testing.T, testCases []TestCase) {
-	for _, testCase := range testCases {
+	// Convenience option for local testing. Grab local token, backup the existing one and restore it after the test.
```
jochil, line 60: Honestly I am not sure if it so convenient when something went wrong during the test I have to figure out why I am not logged in anymore

e2e-tests/e2e-test-runner.go
```diff
@@ -70,6 +98,22 @@ func runTest(t *testing.T, testCase *TestCase) {
...
+			fmt.Println("You are trying to test LoggedIn behavior, you need to set " + envvarWithE2EUserToken + " envvar.")
+			t.FailNow()
+		}
+		testCase.Environment = append(testCase.Environment, "CIFUZZ_API_TOKEN="+os.Getenv(envvarWithE2EUserToken))
+	}
+
+	if testCase.CIUser == InvalidTokenCIUser {
+		testCase.Environment = append(testCase.Environment, "CIFUZZ_API_TOKEN=invalid")
+	}
+
```
jochil, line 116: is there any reason why not using `if...elseif` or `switch` here... I guess it would be more readable

e2e-tests/finding/finding_test.go
```diff
@@ -0,0 +1,87 @@
...
+		},
+	},
+	{
+		Description:  "finding command in a project without findings",
+		Command:      "finding",
+		SampleFolder: []string{"project-with-empty-cifuzz-yaml"},
+		Assert: func(t *testing.T, output e2e.CommandOutput) {
+			assert.EqualValues(t, 0, output.ExitCode)
+			assert.Equal(t, output.Stdout, "")
+			assert.Contains(t, output.Stderr, "This project doesn't have any findings yet")
```
jochil, line 29: There is a lot of repeating code for the same assertions... I made a quick draft for reducing this: https://github.com/CodeIntelligenceTesting/cifuzz/commit/5a92a4d715531fa27fcfae662923e1ea42c41606

WDYT?

e2e-tests/e2e-test-runner.go
JackuB: Keeping it empty. If I add newline to the previous Println statement, linter complains about it

e2e-tests/e2e-test-runner.go
JackuB: > honestly I vote for not using any remote server at all

I'm not totally opposed to it, yet do we check that "full-circle" cifuzz<->API interaction anywhere else currently?

---

> I vote for using the staging server

> but if we use one, we should use staging and talk to the web team about it

Why would you use a staging?

e2e-tests/finding/finding_test.go
JackuB: Added this to the PR

e2e-tests/e2e-test-runner.go
JackuB: I want you to be able to run these tests locally, without much setup 🤷 using the `cleanup` should ensure that tokens are persisted

e2e-tests/e2e-test-runner.go
jochil: > I'm not totally opposed to it, yet do we check that "full-circle" cifuzz<->API interaction anywhere else currently?

I am not sure about this... maybe in the core repo. @simonresch might know
But I agree we need some real e2e/integration tests, but I would prefer them setup independently from this repo, as this is only one part of the whole product. And even if we test against a real (prod) server it is not really covering the server part as there is no actual reporting / logging /... with the server team involved
Also it would make our pipeline useless if there is a downtime

> Why would you use a staging?

To make sure we are not causing any trouble in production, like distorted usage data or DDoSing ourself... should not happen, but you never know. Also having test data (accounts, tokens, projects...) in our prod setup just feels wrong

I am aware that there are reasons for using prod (less downtime, matches what the user is actually using). But I am not sure about if it is worth doing this

e2e-tests/e2e-test-runner.go
jochil: IMO tests that have side effects on the local system are prone for hard to debug problems... for example running them in parallel will most probably cause race conditions access the single file

AFAIK in tests it is not guaranteed that `defer` functions are called... for example when a panic occurs. To make sure the file is renamed back I think you have to use `t.Cleanup`

e2e-tests/e2e-test-runner.go
simonresch: > I am not sure about this... maybe in the core repo.

We currently have tests in the core repository that create a bundle with cifuzz and then start bundle execution via `cictl`. The test asserts certain findings and coverage information. https://gitlab.code-intelligence.com/code-intelligence/core/-/blob/master/backend/scripts/smoke_test/core.clj#L728
I'm happy to add more test cases there.

The limitation of the test in the core repository currently is that you would not be able to run it easily for every commit or PR, since we use a fixed `cifuzz` release which is manually updated.

JackuB: Switched to staging

jochil (approved):
lgtm

PR #736, opened 2023-04-06
Unquarantine windows integration tests

florianGla:


JackuB:
@florianGla I can unquarantine tests as part of https://github.com/CodeIntelligenceTesting/cifuzz/pull/740 - WDYT? The parallel execution probably solves similar issues as this PR. Then we could close this PR

florianGla:
> @florianGla I can unquarantine tests as part of #740 - WDYT? The parallel execution probably solves similar issues as this PR. Then we could close this PR

Sure, that would be the best solution.

PR #739, opened 2023-04-11
Fix broken integration tests

florianGla:
The integration tests now work on windows.


pkg/runfiles/finder_test.go
```diff
@@ -24,38 +26,49 @@ func TestMain(m *testing.M) {
 	m.Run()
 }
 
-func TestLlvmToolPath(t *testing.T) {
+func TestIntegration_LlvmToolPath(t *testing.T) {
```
MarkusZoppelt, line 29:
```suggestion
func TestIntegration_LLVMToolPath(t *testing.T) {
```

internal/cmd/coverage/llvm/llvm_test.go
```diff
@@ -92,8 +101,6 @@ func TestLLVM(t *testing.T) {
 				assert.DirExists(t, reportPath)
 				assert.FileExists(t, filepath.Join(reportPath, "index.html"))
 			}
-			assert.Contains(t, bOut.String(), "src/explore_me.cpp")
-			assert.Contains(t, bOut.String(), "my_fuzz_test.cpp")
```
MarkusZoppelt, line 96: why were these removed?

MarkusZoppelt (approved):

internal/cmd/coverage/llvm/llvm_test.go
florianGla: They check if those lines appear in the build output of the makefile. I changed the build system to cmake, in order to be able to test it on Windows. Those checks cannot be done for cmake projects

PR #740, opened 2023-04-12
Run Integration tests in parallel containers

JackuB:
giphy

This PR introduces a scheduled GH Action to generate testing buckets to speed up our PR checks! Goal was to get test time under 10 minutes, and we are almost there. On a good day, tests take ~8 minutes, on a bad one ~12.

- Ubuntu and macOS containers are getting 3 containers each
- Windows tests run on 6 containers
- Tests are balanced using a few .txt files, generated by judging how long each package test takes from `go test ./... -json` output. These times are updated via a scheduled GH Action that automatically opens a PR with new tests timing each week.
- Tests that are added between these balancing runs will be automatically added to the fastest (1st) container.
- CPP error tests are extremely slow on Windows currently, thus they are skipped. We can discuss this point, possibly this will require a separate PR to dig deeper.
- "VM Setup" action is currently unpredictable, taking anywhere from less than 1 minute to over 5 minutes. This is making it hard for us to stay under 10 minutes.


.github/workflows/create_integration_test_buckets.yml
```diff
@@ -0,0 +1,45 @@
...
+
+jobs:
+  bucket-generator:
+    name: bucket generator
+    runs-on: ubuntu-22.04
+
+    steps:
+      - uses: actions/checkout@v3
+      - name: Setup VM for cifuzz
+        uses: ./.github/actions/vm_setup
```
MarkusZoppelt, line 15:
```suggestion
        uses: .github/actions/vm_setup
```

jochil:
One thing I was asking myself about the buckets... Why are we defining the number of buckets we want to have, instead of the max time and let the algorithm choose the amount of buckets needed? Of course we need to respect the upper limit of buckets... but the parameter we are optimizing for is time, or?

.github/workflows/create_integration_test_buckets.yml
```diff
@@ -0,0 +1,46 @@
+name: Generate Integration Test Buckets
+
+on:
+  schedule:
+    - cron: "0 1 * * 3" # Every Wed at 1AM
```
jochil, line 5: doesn't this mean that all packages/tests that are added in between a week will not be executed during the tests?
Ok, you can trigger it manually but this is of course easy to forget... and will end in a second PR

WDYT about creating the buckets more often but saving them in some location and overwrite them regulary?

jochil: I am also not the biggest fan of having generated data in the repository

JackuB: > I am also not the biggest fan of having generated data in the repository

Yes, _and_ these are generated, but capturing a state of a test's performance and load _at the time_. Keeping these related to the commits will allow you inspect or revert changes in test times etc.

Putting them in an S3 or something is an option, since it will avoid the PR step - do you think that's worth it?

---

> Doesn't this mean that all packages/tests that are added in between a week will not be executed during the tests?

Plan is to use `go list` to check if everything is included 👌

JackuB:
>  Why are we defining the number of buckets we want to have, instead of the max time and let the algorithm choose the amount of buckets needed? Of course we need to respect the upper limit of buckets... but the parameter we are optimizing for is time, or?

We do have an upper concurrency limit (20, including all other jobs running in this action), splitting it further will make it slower, as there are steps that benefit from concurrency.
Also, I think defining the Actions' matrix size dynamically is not supported by GH Actions. I can imagine a workaround, but it would introduce quite some complexity.

.github/workflows/create_integration_test_buckets.yml
JackuB: @MarkusZoppelt FYI it doesn't work without the prefix, and I had to revert it

JackuB: Screenshot 2023-04-14 at 17 37 47

.github/workflows/create_integration_test_buckets.yml
jochil: > Plan is to use go list to check if everything is included ok_hand

sounds good... what is the planned behavior when a package is missing? Executing the diff in a separate task?

> Putting them in an S3 or something is an option, since it will avoid the PR step - do you think that's worth it?

Maybe we can start with the PR approach and see how much we annoyed of it?

JackuB: > what is the planned behavior when a package is missing?
New packages are assigned to the fastest (first) bucket. And they'll be picked up next time buckets are regenerated.

integration-tests/errors/cpp/errors_test.go
```diff
@@ -22,6 +23,11 @@ func TestIntegration_CPPErrors(t *testing.T) {
 		t.Skip()
 	}
 
+	// These tests run 20+ minutes on Windows
+	if runtime.GOOS == "windows" {
+		t.Skip()
+	}
```
jochil, line 29: Do we have a task to tackle this in the backlog? IMO it is important to have these tests on windows as well

jochil (approved):

MarkusZoppelt (approved):

integration-tests/errors/cpp/errors_test.go
JackuB: We have it now 👍 CLI-1009

PR #743, opened 2023-04-13
test: remove sequential limitation

JackuB:
Testing whether we still need this option


jochil:
To actually run tests in parallel you have to use `t.Parallel` in the tests and make sure that `GOMAXPROCS` is set to a value bigger 1 or setting `parallel` to some number that makes sense

IMO it would also be helpful to run the tests multiple times (in a loop or with `-count`) to give us more confidence that there are no side effects... concurrency can be tricky.

JackuB:
Resolved with https://github.com/CodeIntelligenceTesting/cifuzz/pull/740

PR #744, opened 2023-04-13
fix: reduce request timout to 10 secs

MarkusZoppelt:
Since our CI pipeline runs take a really long time to complete, we want to see if a reduced request timeout brings measurable improvements.


jochil:
IMO we can close this one, because:
1. Most probably it is not the reason for slow tests... local http requests against a mock server (both in go) should be **blazingly fast ™**
2. adjusting a user relevant parameter just to make our tests faster smells bad... we choose the timeout for a reason

MarkusZoppelt:
> * Most probably it is not the reason for slow tests... local http requests against a mock server (both in go) should be **blazingly fast ™**
> * adjusting a user relevant parameter just to make our tests faster smells bad... we choose the timeout for a reason

@jochil I agree with both points. I was just curious if maybe the requests were somehow wonky on the GitHub CI, but that seems to be happening unrelated to these changes.

PR #746, opened 2023-04-14
fix(jazzer): don't append generated corpus dir

MarkusZoppelt:
For JVM projects, we get a wrong count of corpus entries (it’s an overflow since we substract two unsigned numbers to calculate the delta).
9a8ea1a9-ca59-4338-b993-a2cf397f6ec8

From the `-v` invocation from cifuzz, we can see that corpus-entries get added twice:

```
INFO: Seed: 451139807
INFO: Loaded 1 modules   (512 inline 8-bit counters): 512 [0x1701e0000, 0x1701e0200),
INFO: Loaded 1 PC tables (512 PCs): 512 [0x13125a400,0x13125c400),
INFO:        4 files found in cifuzz/examples/gradle/.cifuzz-corpus/com.example.FuzzTestCase
INFO:        1 files found in cifuzz/examples/gradle/build/resources/test/com/example/FuzzTestCaseInputs
INFO:        4 files found in cifuzz/examples/gradle/.cifuzz-corpus/com.example.FuzzTestCase
INFO:        1 files found in cifuzz/examples/gradle/src/test/resources/com/example/FuzzTestCaseInputs
INFO: -max_len is not provided; libFuzzer will not generate inputs larger than 4096 bytes
INFO: Instrumented com.example.ExploreMe (took 4 ms, size +86%)
INFO: seed corpus: files: 10 min: 2b max: 18b total: 88b rss: 1021Mb
Initializing fuzzer with 10 seed inputs
```

Jazzer needs to come up with the same paths anyway, so there should be no need for cifuzz to provide the default paths.

**Solution:** not adding the corpus directory arguments from cifuzz, relying on Jazzer to set them instead.


pkg/runner/jazzer/jazzer_runner.go
```diff
@@ -108,9 +108,6 @@ func (r *Runner) Run(ctx context.Context) error {
 		args = append(args, options.LibFuzzerRSSLimitFlag("3000"))
 	}
 
-	// Tell libfuzzer which corpus directory it should use
-	args = append(args, r.GeneratedCorpusDir)
-
 	// Add any additional corpus directories as further positional arguments
 	args = append(args, r.SeedCorpusDirs...)
```
fmeum, line 112: I would expect this to still cause duplication as I think the inputs directory is passed as a seed dir.

MarkusZoppelt: Hm, then I'm not entirely sure what is happening. The change eliminated what caused the overflow and the corpus entry counts seems correct afterwards.

fmeum: Then it's probably fine. I saw two duplicated paths, but this change could only really get rid of one.

I assume the other is less problematic as Jazzer won't emit findings into it, but it may still affect the INFO output.

fmeum: Then it's probably fine. I saw two duplicated paths, but this change could only really get rid of one.

I assume the other is less problematic as Jazzer won't emit findings into it, but it may still affect the INFO output.

MarkusZoppelt: > I would expect this to still cause duplication as I think the inputs directory is passed as a seed dir.

We also checked `run` and now only add user-defined corpus dirs.

fmeum (approved):

PR #747, opened 2023-04-14
fix: add `shift_exponent` to error ID matchers

MarkusZoppelt:
### Motivation & Context
`shift exponent` errors were not picked up consistently and matched with error IDs, e.g., in the `cmake` example:

```
⚠️ unable to find matching error id for given finding: undefined behavior: shift exponent 32 is too large for 32-bit type 'int'
```

Libfuzzer output was:

```
explore_me.cpp:13:11: runtime error: shift exponent 32 is too large for 32-bit type 'int'
```

### Description

Added `shift_exponent` to error ID matchers.


pkg/parser/errorid/errorid.go
```diff
@@ -56,6 +56,11 @@ var matchers = []matcher{
 	{id: "use_of_uninitialized_value", substrings: []string{"use-of-uninitialized-value"}},
 
 	// more global issues, should be at the end so they do not overwrite more explicit ones
+	{
+		id:         "undefined behavior",
```
jochil, line 60:
```suggestion
		id:         "undefined_behavior",
```
all the other IDs are like this

pkg/parser/errorid/errorid.go
```diff
@@ -56,6 +56,11 @@ var matchers = []matcher{
 	{id: "use_of_uninitialized_value", substrings: []string{"use-of-uninitialized-value"}},
 
 	// more global issues, should be at the end so they do not overwrite more explicit ones
+	{
+		id:         "undefined_behavior",
+		substrings: []string{"undefined behavior", "undefined behaviour"},
+		regexs:     []*regexp.Regexp{regexp.MustCompile(`undefined behaviou?r: .*`)},
```
jochil, line 62: It is not necessary to match for substrings and regex if they are doing essential the same

jochil (changes requested):
can you please update the package unit tests and the error integration tests with a matching example?

pkg/parser/errorid/errorid.go
```diff
@@ -56,6 +56,11 @@ var matchers = []matcher{
 	{id: "use_of_uninitialized_value", substrings: []string{"use-of-uninitialized-value"}},
 
 	// more global issues, should be at the end so they do not overwrite more explicit ones
+	{
+		id:         "undefined_behavior",
+		substrings: []string{"UndefinedBehaviorSanitizer"},
+		regexs:     []*regexp.Regexp{regexp.MustCompile(`undefined behaviou?r`)},
+	},
```
jochil, line 63: the strings we are parsing are coming from our libfuzzer parser, which usually works on the line above the stacktrace... so in this specific case on something like `runtime error: shift exponent 32 is too large for 32-bit type 'int'`... you can take a look at the `out-of-bounds` error, which is also a undefined behavior bug. I have the impression that just looking for `undefined behavior` is a little bit to generic, espeically when we have a specific entry for invalid shifts in our `error_details.json`: https://gitlab.code-intelligence.com/code-intelligence/core/-/blob/master/pkg/fuzzing/error_details/error-details.json#L281

MarkusZoppelt: I added a specific `shift_exponent` error check and leave the "generic" `undefined_behavior` check there, just in case.

MarkusZoppelt: Would you rather remove the undefined behavior check completely and only check for exact matches?

MarkusZoppelt: I removed the check. We can add it back if there is a clear need for it.

jochil (approved):
lgtm... thanks

pkg/parser/errorid/errorid.go
jochil: I am fine with removing it for now.

PR #749, opened 2023-04-17
fix(summary): split file path into two parts

florianGla:
Source file lines in the lcov coverage report have the following form on Windows:
SF:C:\Users\...\my_fuzz_test.cpp

We split this path using colon as the delimiter. In order to get it to work with absolute paths on Windows, we should only split the line into two parts. Otherwise the path is split into {"C", "Users\...\my_fuzz_test.cpp"}.


florianGla:
> Could you write a quick unit test for this so we are sure that it this scenario always works?

Done.

mlsandnerCI (approved):

internal/cmd/coverage/summary/lcov.go
```diff
@@ -43,7 +43,7 @@ func ParseLcov(in io.Reader) *CoverageSummary {
 	// with `man geninfo`
 	scanner := bufio.NewScanner(in)
 	for scanner.Scan() {
-		parts := strings.Split(scanner.Text(), ":")
+		parts := strings.SplitN(scanner.Text(), ":", 2)
```
jochil, line 46: sorry for being late to the party, but I think this is a use case for https://pkg.go.dev/path/filepath#VolumeName

florianGla: I'm not sure if that package would simplify the problem. It think splitting it into exactly two parts is the right way to go. We want to split those lines `SF:C:\Users\...\my_fuzz_test.cpp` into `SF` and `C:\Users\...\my_fuzz_test.cpp`.
We could split all colons and get three parts and check if the second part is a volume name and then merge the second and third part. But I think my solution is simpler. What solution did you have in mind?

jochil: A sorry... my fault. I thought we are splitting for the "C" part of the example. So nevermind :)

PR #750, opened 2023-04-17
feat: allow jazzer parameters for maven & gradle

mlsandnerCI:
### Motivation & Context
Right now we support adding jazzer parameters via `--engine-arg `in the `run` command. This PR is expanding this feature to `coverage. `bundle` and `remote-run` already allow setting these.

### Description
The flag `--engine-arg` is now usable for the `coverage` command for maven & gradle. The docs have been updated with the usage and minor refactoring was included to make code easier to understand.


internal/cmd/coverage/maven/maven.go
```diff
@@ -40,15 +43,20 @@ func (cov *CoverageGenerator) BuildFuzzTestForCoverage() error {
...
-		// We still want to generate the coverage report, so use this flag:
-		"-Dmaven.test.failure.ignore=true",
-		"-Djazzer.hooks=false",
+	// Maven tests fail if fuzz tests fail, so we ignore the error here,
+	// so we can still generate the coverage report
+	mavenTestArgs := []string{"-Dmaven.test.failure.ignore=true"}
+
+	// Flags for jazzer
+	mavenTestArgs = append(mavenTestArgs, "-Djazzer.hooks=false")
+	mavenTestArgs = append(mavenTestArgs, cov.EngineArgs...)
```
MarkusZoppelt, line 52: Maybe do this in a single `append`, like below for the cifuzz flags.

docs/Configuration.md
```diff
@@ -74,15 +74,32 @@ dict: path/to/dictionary.dct
...
+
+#### Example
+```yaml
+engine-args:
+ - -Djazzer.instrumentation_includes=com.** 
+```
+
+Specific for **Gradle**, the following code has to be added to the `build.gradle` file 
+in the project to allow using system properties via `-D` flags. 
+```
```
MarkusZoppelt, line 97: Nice to have...

```suggestion
```groovy
```

MarkusZoppelt (commented):
Nice! Can we also add a small test with some additional jazzer parameters to see if this works?

internal/cmd/coverage/maven/maven.go
mlsandnerCI: Sorry, can't combine a string slice and single string in one append.

MarkusZoppelt (approved):
LGTM

PR #752, opened 2023-04-20
normalize URLs and remove trailing slashes

MarkusZoppelt:
To avoid future parsing problems, we remove trailing slashes from server URLs.

There might already be cases where this might have caused problems, because we store the server URL as the key in `access_tokens.json`.

This fix avoids having, e.g., both

* `https://app.code-intelligence.com` and
* `https://app.code-intelligence.com/`

in the `access_tokens.json` file.


internal/api/api.go
```diff
@@ -326,11 +327,13 @@ func ValidateAndNormalizeServerURL(server string) (string, error) {
 		server = "https://" + server
 	}
 
-	// normalize server URL
-	url, err := url.JoinPath(server)
+	// normalize server URL by removing trailing slash
+	url, err := url.JoinPath(server, "")
```
JackuB, line 331: Will this change render all access_token files with trailing slash as invalid?

MarkusZoppelt: Yes 🙈
Any suggestions?

internal/tokenstorage/tokenstorage.go
```diff
@@ -67,8 +68,20 @@ func Set(target, token string) error {
...
+	if token, ok := accessTokens[target]; ok {
+		return token
+	}
+	if token, ok := accessTokens[strings.TrimSuffix(target, "/")]; ok {
+		return token
+	}
+	if token, ok := accessTokens[target+"/"]; ok {
+		return token
+	}
+	return ""
```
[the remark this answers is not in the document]
MarkusZoppelt: Done!

JackuB (approved):

PR #757, opened 2023-04-20
implement `create` for NodeJS projects

MarkusZoppelt:
Requires `CIFUZZ_PRERELEASE=true` environment flag


MarkusZoppelt:
@bertschneider Could you please take a look at the test stubs (`pkg/stubs/test.fuzz.{js|ts}.tmpl`) and check if they look good? Thank you!

pkg/stubs/stubs.go
```diff
@@ -38,37 +44,43 @@ func Create(path string, testType config.FuzzTestType) error {
...
+				if testType == config.Kotlin {
+					packageName = strings.TrimSuffix(packageName, ";")
+				}
+				content = []byte(strings.Replace(string(content), "__PACKAGE__", packageName, 1))
 			}
-			content = []byte(strings.Replace(string(content), "__PACKAGE__", packageName, 1))
+		}
+	case config.JavaScript, config.TypeScript:
+		content = javaScriptStub
+		if testType == config.TypeScript {
```
jochil, line 76: using the if block here is even longer than having just
```
case config.TypeScript:
  content = typeScriptStub
```
and IMO it is less readable

pkg/stubs/test.fuzz.js.tmpl
```diff
@@ -0,0 +1,25 @@
+const { FuzzedDataProvider } = require("@jazzer.js/core");
+// const target = require("./target");
+
+describe("My describe", () => {
+  it.fuzz("My async fuzz test", async (data) => {
+   	// Call the functions you want to test with the provided data and optionally
+   	// assert that the results are as expected.
+
+   	// For example:
+   	// await target.asyncFuzzMe(data);
```
jochil, line 10: We should show how you can use the `FuzzedDataProvider` to get the needed data types... most of the function will not just take a `Buffer`

pkg/stubs/test.fuzz.ts.tmpl
jochil, line 1: This is basically the exact same file as the javascript stub... IMO we should  use TypeScript features to highlight the differences

pkg/stubs/test.fuzz.js.tmpl
```diff
@@ -0,0 +1,25 @@
+const { FuzzedDataProvider } = require("@jazzer.js/core");
```
jochil, line 1: why requiring it when not using it?

pkg/stubs/test.fuzz.js.tmpl
```diff
@@ -0,0 +1,25 @@
...
+describe("My describe", () => {
+  it.fuzz("My async fuzz test", async (data) => {
+   	// Call the functions you want to test with the provided data and optionally
+   	// assert that the results are as expected.
+
+   	// For example:
+   	// await target.asyncFuzzMe(data);
+  });
+
+  // Alternatively, you can use a callback to access the data provider.
```
jochil, line 13: IMO this does not make sense... the callback is not used to access the data provider.
1. data is not a "data provider"... it is just a Buffer
2. the callback is used to handle the async behavior

From my side we can skip the `done` part... doing it with `async/await` is way more common nowadays

pkg/stubs/test.fuzz.js.tmpl
```diff
@@ -0,0 +1,25 @@
+const { FuzzedDataProvider } = require("@jazzer.js/core");
+// const target = require("./target");
+
+describe("My describe", () => {
+  it.fuzz("My async fuzz test", async (data) => {
```
jochil, line 5: I think I mentioned it before, but I am not a fan of using `it` here, as it makes no sense semantically... it should be used to describe expected behavior, like in `it('should hide ... when ...", () => {}` and fuzz tests are the opposite of expected behavior :wink: 

In general I think we could make the stubs looking more like our example https://github.com/CodeIntelligenceTesting/cifuzz/blob/main/examples/nodejs/FuzzTestCase.fuzz.js

internal/cmd/create/create_test.go
```diff
@@ -75,6 +75,42 @@ func TestOkGradle(t *testing.T) {
...
+	defer cleanup()
+
+	args := []string{
+		"javascript",
+		"--output",
+		filepath.Join(testDir, "myTest.fuzz.js"),
+	}
+
+	_, err := cmdutils.ExecuteCommand(t, New(), os.Stdin, args...)
+	assert.NoError(t, err)
```
jochil, line 93: why not testing if the file exists?

jochil, line 93: I guess using `require.NoError` is fine here... we usually do it that way, because when the function fails unexpectedly there is no point in running more assertions

pkg/stubs/test.fuzz.js.tmpl
MarkusZoppelt: I know, it is just that I was following the [jazzer.js docs here](https://github.com/CodeIntelligenceTesting/jazzer.js/blob/main/docs/jest-integration.md#async-jest-fuzz-tests) (and also the info from the jira ticket) and they exclusively use `it`.

pkg/stubs/test.fuzz.js.tmpl
MarkusZoppelt: Was following the example from here: https://github.com/CodeIntelligenceTesting/jazzer.js/blob/main/docs/jest-integration.md#async-jest-fuzz-tests

I too think that a single/simpler stub makes more sense, so we may drop the alternative...

internal/cmd/create/create_test.go
MarkusZoppelt: The way it is now it is in line with all other tests for create (Gradle, Maven, etc.)

If we do it differently, we should change all of the other tests as well.

pkg/stubs/test.fuzz.js.tmpl
bertschneider: Agreed. You could use `test` as well, if that makes more sense here.

pkg/stubs/test.fuzz.js.tmpl
bertschneider: I would also tend to remove the callback example. It was just added to support all available Jest test pattern, so that existing tests can "easily" be changed to fuzz tests. It will probably not be used much.

pkg/stubs/test.fuzz.js.tmpl
MarkusZoppelt: @bertschneider Would it make sense then changing your examples too so that are more aligned with ours?

pkg/stubs/test.fuzz.ts.tmpl
```diff
@@ -0,0 +1,25 @@
+const { FuzzedDataProvider } = require("@jazzer.js/core");
```
bertschneider, line 1: Currently, TypeScript fuzz tests need to be transpiled to JavaScript before executing them with Jest. We are working on supporting direct execution of TypeScript tests with Jest / ts-jest.

MarkusZoppelt: Does that mean we should not yet include TypeScript support (creating a ts stub, etc.)?

pkg/stubs/test.fuzz.js.tmpl
bertschneider: We could do so, yes. I don't have strong feelings regarding `it` or `test` in the examples to begin with.

pkg/stubs/test.fuzz.ts.tmpl
bertschneider: That's probably quite project specific. I would tend to leave it out until no major hurdles can be expected anymore. We want to work on appropriate support short-term, though.

MarkusZoppelt: Hm, we'll leave it out then. Either way, we have CLI-874 to revisit this. cc @jochil

internal/cmd/create/create_test.go
MarkusZoppelt: Done

pkg/stubs/test.fuzz.ts.tmpl
MarkusZoppelt: Resolving because of [this thread](https://github.com/CodeIntelligenceTesting/cifuzz/pull/757#discussion_r1173740641).

pkg/stubs/test.fuzz.js.tmpl
```diff
@@ -0,0 +1,13 @@
...
+test.fuzz("My async fuzz test", async (data) => {
+	const provider = new FuzzedDataProvider(data);
+ 	// Call the functions you want to test with the provided data and optionally
+ 	// assert that the results are as expected.
+
+ 	// For example:
+ 	// await target.asyncFuzzMe(
+	//   provider.consumeNumber(),
+	//   provider.consumeString(8));
+});
```
florianGla, line 13: In all other stubs the FuzzedDataProvider is commented out. It is optional to use it and we should not force the user. Why do we use await here and not in our nodejs example project. The stub should be similar to our provided examples.

florianGla:
Maybe we should start to write an integration test for init and create. What do you think?

pkg/stubs/test.fuzz.js.tmpl
jochil: +1 to commenting out the `FuzzedDataProvider`

I would be fine removing the `async/await` stuff here, as it works like you would expect it to work... IMO it is well known.

jochil (approved):

PR #763, opened 2023-04-24
exclude third party code from being instrumented by jazzer

florianGla:
By default third party code should be excluded from being instrumented by jazzer. This PR tries to generate glob patterns for instrumentation_includes, which only match local classes.


MarkusZoppelt (commented):
Thanks! Would you mind helping me understand some of the changes by clarifying:

1) Does this PR play well with the changes and comments in #766?

2) Aren't some of these changes already included in #750, e.g., passing `-Djazzer.instrumentation_includes=com.**`?

florianGla:
@MarkusZoppelt It turns out that this PR will be probably closed because a simplified approach can be added to jazzer directly.

PR #764, opened 2023-04-25
fix: don't show empty MoreDetails table in finding details

MarkusZoppelt:
### Motivation & Context
After the error ID refactoring, the `f.MoreDetails` struct always contains data (the `id`), therefore the check in the `finding <specific_finding>` cmd was not working correctly.

### Description
These changes check if users are logged in before attempting to print more details and then check if `MoreDetails` include a name and severity before printing a table with more extensive information.


internal/cmd/finding/finding_cmd.go
```diff
@@ -205,8 +205,11 @@ func (cmd *findingCmd) printFinding(f *finding.Finding) error {
 		if err != nil {
 			return err
 		}
-		PrintMoreDetails(f)
 
+		// only show more details if the user is authenticated
+		if auth, err := auth.GetAuthStatus(cmd.opts.Server); err == nil && auth {
```
mlsandnerCI, line 210: You are not returning the error when it fails, is this intended?

mlsandnerCI: The readability is fine, I just wasn't sure if non-nil should be ignored or returned 👍

mlsandnerCI (approved):

PR #770, opened 2023-04-28
fix(bundle): make sure that bundles created on windows will run on linux workers (only jazzer)

jochil:
- explicitly use forward slashes as requested from the `zip.FileHeader` struct / zip spec
- use forward slashes in `runtime_paths` inside `bundle.yaml`


florianGla (approved):
LGTM

PR #772, opened 2023-05-02
Gradle Integration: Do not attempt to call 'gradle testClasses'

jjohannes:
### Motivation & Context

See: https://github.com/CodeIntelligenceTesting/cifuzz-gradle-plugin/pull/7

### Description

This removes a legacy from before the Gradle plugin. It was called to compile the test code. However, `gradle cifuzzPrintTestClasspath`, which we call next, will take care of compiling the code. This will select the correct compile task(s) depending on which test source set was selected etc.

`testClasses` might also not exist in all projects. Especially Android projects usually do not have it.

### DIFF
--- a/internal/build/gradle/gradle.go
+++ b/internal/build/gradle/gradle.go
@@ -92,18 +92,6 @@ func (b *Builder) Build(targetClass string) (*build.Result, error) {
 		flags = append(flags, "--parallel")
 	}
 
-	args := append([]string{"testClasses"}, flags...)
-	cmd, err := buildGradleCommand(b.ProjectDir, args)
-	if err != nil {
-		return nil, err
-	}
-
-	log.Debugf("Command: %s", cmd.String())
-	_, err = cmd.Output()
-	if err != nil {
-		return nil, cmdutils.WrapExecError(errors.WithStack(err), cmd)
-	}
-
 	deps, err := b.getDependencies()
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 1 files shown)


internal/build/gradle/gradle.go
```diff
@@ -92,18 +92,6 @@ func (b *Builder) Build(targetClass string) (*build.Result, error) {
 		flags = append(flags, "--parallel")
 	}
 
-	args := append([]string{"testClasses"}, flags...)
```
jochil, line 95: removing this causes a linter error as `flags` is not used anymore. 
Does it makes sense to add the `--parallel` flag to all gradle command calls? If so we can move the block L90-93 it to the `buildGradleCommand` function. WDYT?

jjohannes: I would say it does not make sense to add `--parallel`. This is something the user should configure for her build if needed. It has no effect when only running tasks for one (sub)project anyway. Only when you build/test multiple subprojects in parallel. (And soon Gradle will be parallel-by-default anyway.)

I think we should remove the `flags = append(flags, "--parallel")` line.

jjohannes: IIUC the _flags_ code was now unused anyway. I just missed it. Removed it now.

jochil: ok, perfect... I guess we are fine then

jochil (approved):

PR #774, opened 2023-05-04
ci(pr): fix PR pipeline for pull requests created from forks

jochil:
... by replacing our manual git-based solution with an existing github action which uses the github api


.github/workflows/pipeline_pr.yml
```diff
@@ -6,41 +6,17 @@ on:
...
-        run: |
-          git fetch origin "${HEAD_REF}"
-          if git log "origin/${BASE_REF}..origin/${HEAD_REF}" | grep -B4 -i -e "^\s\+fixup"; then
-            exit 1
-          fi
+      - name: Check Commit Messages
+        uses: gsactions/commit-message-checker@v2
+        with:
+          pattern: '^(?!WIP|xx|fixup).*'
+          error: 'A commit message should not start with (XXX|fixup|WIP)'
```
florianGla, line 19: Why are `xx` in pattern lowercase and in error uppercase?

florianGla (approved):

.github/workflows/pipeline_pr.yml
jochil: should be `XXX` (as it was before) in both cases... fixed

PR #778, opened 2023-05-05
Android support for local tests

jochil:
... by upgrading to v1.1.0 of the `cifuzz gradle plugin`: https://plugins.gradle.org/plugin/com.code-intelligence.cifuzz


README.md
```diff
@@ -257,6 +258,20 @@ choco install microsoft-openjdk gradle
...
+<details>
+ <summary>Android</summary>
+
+**Info:** Currently cifuzz is **not** supporting fuzz tests running in an
+emulator or on a device, it is still possible to run local tests. 
+You can find more information and an example at
+the [cifuzz-gradle-plugin](https://github.com/CodeIntelligenceTesting/cifuzz-gradle-plugin)
+repository.
+
+* [Gradle](https://gradle.org/install/) >= 8.0
```
jjohannes, line 270:
```suggestion
* [Gradle](https://gradle.org/install/) >= 7.5 
```

If you use Androind `7.4` Gradle 7.5 is fine. With newer Android versions, you also need newer Gradle versions. But that's because of the Android plugin (and it will tell the user) not because of us.

So I think `7.5` is correct here. See also: https://github.com/CodeIntelligenceTesting/cifuzz-gradle-plugin#compatibility

PBetzler:
Testing the new version on the android example in the gradle-plugin repository works, but when creating a new android project like [here](https://github.com/PBetzler/Android-Automotive-Fuzzing-Example) it fails to detect the tests.
Is this a (test-) setup issue?

PBetzler:
You were right, there was a junit4 dependency in the build.gradle. It has been removed, but the error persists.

The log output is attached at the end.
 
[failing.log](https://github.com/CodeIntelligenceTesting/cifuzz/files/11408154/failing.log)

jjohannes:
I am afraid we still have a bug. The difference I can see is that the example we test with has the test written in Kotlin. And your example is in pure Java. Of course that should work as well.

jjohannes:
This was bothering me. Here is the attempt to fix it: https://github.com/CodeIntelligenceTesting/cifuzz-gradle-plugin/pull/9

jochil:
I released the bugfix as `1.1.1` of the gradle plugin and updated this PR accordingly

PBetzler (approved):

PR #779, opened 2023-05-05
ci: skip commit message check for merge group

jochil:
... as it is not necessary (was executed in every PR before the merge group) and is also not supported by the used action.

It still has to be present to get triggered for the required status checks


.github/workflows/pipeline_pr.yml
```diff
@@ -13,6 +13,8 @@ jobs:
 
     steps:
       - name: Check Commit Messages
+        # we can skip this check on merge groups (it is also not supported by the used action)
+        if: ${{ github.event_name != 'merge_group' }}
```
MarkusZoppelt, line 17: Can't we just remove the entry on line 4?

```yaml
on:
  merge_group:
```

MarkusZoppelt (approved):

PR #780, opened 2023-05-08
test: integration test for java errors

mlsandnerCI:


integration-tests/errors/java/java_errors_test.go
```diff
@@ -0,0 +1,136 @@
...
+	testdataTmp := shared.CopyTestdataDir(t, "java_cpp")
+	t.Cleanup(func() { fileutil.Cleanup(testdataTmp) })
+
+	// This testdata is in a separate project due to external dependencies that
+	// need to be downloaded for compilation and slow down every testcase.
+	testdataLDAPAndSQLTmp := shared.CopyCustomTestdataDir(t, "testdata-sql-ldap", "java_cpp_sql_ldap")
+	t.Cleanup(func() { fileutil.Cleanup(testdataLDAPAndSQLTmp) })
+
+	installDir := shared.InstallCIFuzzInTemp(t)
+	t.Cleanup(func() { fileutil.Cleanup(installDir) })
```
MarkusZoppelt, line 31: We stopped removing/cleaning up the cifuzz install dir in tmp to avoid side-effects because other (concurrent) tests might need it, see #789

mlsandnerCI: Good catch 👍

PR #781, opened 2023-05-08
fix(bundle): check if gradlew exists

mlsandnerCI:
### Description
The same check for gradlew we do in `run` we now have in `bundle` so if a wrapper is present but gradle is not installed, the command still works. This also influences `remote-run`.


internal/bundler/jazzer_bundler.go
```diff
@@ -177,6 +177,16 @@ func (b *jazzerBundler) checkDependencies() error {
 	case config.BuildSystemMaven:
 		deps = []dependencies.Key{dependencies.Java, dependencies.Maven}
 	case config.BuildSystemGradle:
+		// First check if gradle wrapper exists and check for gradle in path otherwise
+		wrapper, err := gradle.FindGradleWrapper(b.opts.ProjectDir)
+		if err != nil && !errors.Is(err, os.ErrNotExist) {
+			return err
+		}
```
jochil, line 184: I think we should not do it that way... using the `wrapper` is optional. As long as gradle is installed in the min. needed version we are fine.
In the same package is a ` GetGradleCommand` function. This checks if the wrapper is available and if not it uses `gradle` from the `PATH`. Maybe we can use this function instead and make `FindGradleWrapper` private to avoid using it this way in future. WDYT?

pkg/dependencies/deps_test.go
jochil, line 1: WDYT about adding a test for gradle here? Otherwise the only case using the `projectDir` is not covered

pkg/dependencies/definitions.go
```diff
@@ -166,7 +168,17 @@ var deps = Dependencies{
 		GetVersion: func(dep *Dependency) (*semver.Version, error) {
 			return semver.NewVersion("0.0.0")
 		},
-		Installed: func(dep *Dependency) bool {
+		Installed: func(dep *Dependency, projectDir string) bool {
```
jochil, line 171: As we have a case (`create` command) where `""` is passed as `projectDir`, should we handle that and return false, or are we assuming `""` equals to the current working directory?

pkg/dependencies/definitions.go
```diff
@@ -21,7 +23,7 @@ var deps = Dependencies{
 		Key:        Bazel,
 		MinVersion: getMinVersionBazel(),
 		GetVersion: bazelVersion,
-		Installed: func(dep *Dependency) bool {
+		Installed: func(dep *Dependency, _ string) bool {
```
jochil, line 26: Is there a reason (eg. linter) for using `_` here? I would guess it is more readable using `projectDir`.

jochil (approved):
I like it... more clean and better encapsulated than before :+1:

pkg/dependencies/definitions.go
mlsandnerCI: Since it is not being used except for the special gradle case thought it was just clearer to use `_`.

pkg/dependencies/definitions.go
mlsandnerCI: In the case of `projectDir == ""` I would just skip the wrapper check and default back to needing Gradle itself to be installed. What do you think would be the benefit of returning `false` in this case?

pkg/dependencies/deps_test.go
mlsandnerCI: Added a test 👍

pkg/dependencies/definitions.go
jochil: just skipping the wrapper check is perfect :+1:

pkg/dependencies/definitions.go
jochil: but when you have to work on one of the functions that are **not** the gradle case you have to figure out what `_` actually means and if it affects what you are doing

pkg/dependencies/deps_test.go
jochil: thanks

pkg/dependencies/definitions.go
mlsandnerCI: Good point!

PR #785, opened 2023-05-09
fix(installer): keep window open after installation on windows

MarkusZoppelt:
### Motivation & Context
After the introduction of `cicheck.IsCIEnvironment`, the if condition for the `Press 'Enter' to terminate the installer` message was wrong.
Moreover, the installer only printed the message and waited for input after a successful installation.

### Description
We now wait for `Enter` on every possible exit of the installer.


cmd/installer/installer.go
```diff
@@ -43,7 +43,10 @@ func main() {
 	err := flags.Parse(os.Args)
 	if err != nil {
 		log.Error(errors.WithStack(err))
-		os.Exit(1)
+		os.Exit(func() int {
+			showConfirmDialogOnWindows()
+			return 1
+		}())
```
jochil, line 49: whats about declaring a variable, assign the function to it and reuse it instead of creating the function multiple times?
For example
```
exitFunc := func() int {
      showConfirmDialogOnWindows()
			return 1
}

os.Exit(exitFunc())
```

jochil: or more readable
```
exit := func() {
  showConfirmDialogOnWindows()
  os.Exit(1)
}

exit()
```

MarkusZoppelt: very nice, thank you! Done.

jochil (approved):

PR #787, opened 2023-05-10
chore: remove `CIFUZZ_BUNDLE_ON_UNSUPPORTED_PLATFORMS`

MarkusZoppelt:
Since the new env `CIFUZZ_ALLOW_UNSUPPORTED_PLATFORMS` has been released, we can remove the old `CIFUZZ_BUNDLE_ON_UNSUPPORTED_PLATFORMS`.

Also: Replace `os.Setenv` with `t.Setenv` in tests throughout the project.


integration-tests/shared/bundle.go
```diff
@@ -27,9 +27,7 @@ func TestBundleLibFuzzer(t *testing.T, dir string, cifuzz string, cifuzzEnv []st
 
 	// Make the bundle command not fail on unsupported platforms to be
 	// able to test it on all platforms
-	err := os.Setenv("CIFUZZ_BUNDLE_ON_UNSUPPORTED_PLATFORMS", "1") // TODO: remove me when CIFUZZ_ALLOW_UNSUPPORTED_PLATFORMS is released
-	require.NoError(t, err)
-	err = os.Setenv(config.AllowUnsupportedPlatformsEnv, "1")
+	err := os.Setenv(config.AllowUnsupportedPlatformsEnv, "1")
```
jochil, line 30: Why not resetting it like before? Fiddling around with ENVs in Tests can easily have side effects, especially if the are not restored to the initial values... or am I missing something?

Alternative can be `t.Setenv`

internal/cmd/bundle/bundle_test.go
```diff
@@ -32,12 +32,7 @@ func TestMain(m *testing.M) {
 
 	// Make the bundle command not fail on unsupported platforms to be
 	// able to test it on all platforms
-
-	err := os.Setenv("CIFUZZ_BUNDLE_ON_UNSUPPORTED_PLATFORMS", "1") // TODO: remove me when CIFUZZ_ALLOW_UNSUPPORTED_PLATFORMS is released
-	if err != nil {
-		panic(err)
-	}
-	err = os.Setenv(config.AllowUnsupportedPlatformsEnv, "1")
+	err := os.Setenv(config.AllowUnsupportedPlatformsEnv, "1")
```
jochil, line 35: same here with the resetting of the ENV

integration-tests/shared/bundle.go
MarkusZoppelt: I added a commit to replace all occurrences of `os.Setenv` with `t.Setenv` and/or reset env vars in tests throughout the project.

jochil (approved):

PR #790, opened 2023-05-12
feat: create additional verbose log file in bundle

mlsandnerCI:
### Motivation & Context
When debugging failed bundle processes, CS needs to know with what versions the bundle has been build and what steps occurred.

### Description
- a general `VerboseSecondOutput` possibility has been added to the log process to allow writing the verbose output to a second writer (even if the command wasn't called in verbose mode) 
- the `bundle` and `remote-run` command use the `VerboseSecondOutput` to create a `full-build-*.log` file during the bundling process in the logs dir and is being added as a `build.log` file to the archive
- `TestRunBundle` checks that the log file is added to the archive
- code for logging in `bundle.go` has been grouped and refactored to make the code easier to read
- a `logging` package has been added to `cmdutils` that holds functions necessary for the verbose log + the original build log has been moved in there 
- the color of the maven build output has been disabled for easier to read logs


internal/cmd/bundle/bundle.go
```diff
@@ -237,3 +233,28 @@ on the build system. This can be overridden with a docker-image flag.
 
 	return cmd
 }
+
+func SetUpBundleLogging(cmd *cobra.Command, opts *bundler.Opts) error {
+	var err error
+
+	opts.BundleBuildLogFile, err = log.SetUpVerboseSecondOutputToFile(opts.ProjectDir, opts.FuzzTests)
```
MarkusZoppelt, line 240: Can we rename this and other occurrences of `[..]SecondOutput[..]` to something else?
Maybe `LogFileOutput` or `FileOutput`? _Second_ might imply there is a _First_ or _Primary_ output variable somewhere else.

internal/cmd/remoterun/remoterun.go
```diff
@@ -110,14 +111,20 @@ variable or by running 'cifuzz login' first.
 			// were bound to the flags of other commands before.
 			bindFlags()
 
+			err := bundle.SetUpBundleLogging(cmd, &opts.Opts)
+			if err != nil {
+				log.Errorf(err, "Failed to setup logging: %v", err.Error())
+				return cmdutils.WrapSilentError(err)
+			}
```
MarkusZoppelt, line 118: Why do you wrap the error here and not in bundle?

pkg/log/helper.go
```diff
@@ -0,0 +1,51 @@
...
+	return strings.ReplaceAll(logSuffix, string(os.PathSeparator), "_")
+}
+
+func CreateLogDir(projectDir string) (string, error) {
+	logDir := filepath.Join(projectDir, ".cifuzz-build", "logs")
+	// create logs dir if it doesn't exist
+	err := os.MkdirAll(logDir, 0755)
+	if err != nil {
+		return "", err
+	}
```
MarkusZoppelt, line 48: This will also create a `.cifuzz-build` directory and not only the `logs` directory as the name of the function suggests.
I think this will be fine, but we need to double-check if this causes side-issues... checking if the parent directory already exists would be more safe maybe?

pkg/log/helper.go
```diff
@@ -0,0 +1,51 @@
...
+		logSuffix = fuzzTestNames[0]
+	}
+	// Make sure that calling fuzz tests in subdirs don't mess up the build log path
+	return strings.ReplaceAll(logSuffix, string(os.PathSeparator), "_")
+}
+
+func CreateLogDir(projectDir string) (string, error) {
+	logDir := filepath.Join(projectDir, ".cifuzz-build", "logs")
+	// create logs dir if it doesn't exist
+	err := os.MkdirAll(logDir, 0755)
```
MarkusZoppelt, line 45: I would suggest using `0700` permissions since logs might contain sensitive information

pkg/log/helper.go
```diff
@@ -0,0 +1,51 @@
...
+// to a separate log file for a command.
+func SetUpVerboseSecondOutputToFile(projectDir string, fuzzTests []string) (string, error) {
+	logDir, err := CreateLogDir(projectDir)
+	if err != nil {
+		return "", err
+	}
+	logSuffix := SuffixForLog(fuzzTests)
+	logFile := filepath.Join(logDir, fmt.Sprintf("%s.log", logSuffix))
+
+	VerboseSecondOutput, err = os.OpenFile(logFile, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0755)
```
MarkusZoppelt, line 20: I would suggest using `0644` to allow read and write permissions for the owner and read-only permissions for others

pkg/log/log.go
```diff
@@ -57,12 +64,35 @@ func log(style pterm.Style, icon string, a ...any) {
...
+	if VerboseSecondOutput != nil {
+		verboseLogToSecondOutput(icon, a...)
+	}
+}
+
+func verboseLogToSecondOutput(icon string, a ...any) {
+	if VerboseSecondOutput == nil {
+		// TODO: this would be a programming error, how to handle it?
+		return
+	}
```
MarkusZoppelt, line 87: Will `verboseLogToSecondOutput` ever get called when `VerboseSecondOutput` is nil?

pkg/log/helper.go
mlsandnerCI: But doesn't `MkdirAll()` only create the ones that don't already exist? So if `.cifuzz-build` exists, we only create `logs`, if none exist, it creates both and if both already exist, it does nothing?

pkg/log/helper.go
mlsandnerCI: Thanks!

pkg/log/helper.go
mlsandnerCI: Thanks, will do!

pkg/log/log.go
mlsandnerCI: Nope, you are right 👍  This is code I forgot to delete in my refactoring process.

internal/cmd/remoterun/remoterun.go
mlsandnerCI: Good catch, I will add it in the bundle 👍

pkg/log/helper.go
MarkusZoppelt: Yes, should be fine

internal/cmd/bundle/bundle.go
MarkusZoppelt: The name is not ideal, but as long as it is explained somewhere, it should be fine.

MarkusZoppelt: Maybe change the name to `SecondaryOutputToFile` to use "Secondary" instead of "Second" for better clarity and meaning.

```go
// SetUpVerboseSecondaryOutputToFile configures logging the verbose output
// to a separate log file for a command.
func SetUpVerboseSecondaryOutputToFile(projectDir string, fuzzTests []string) (string, error) {
```

```go
// VerboseSecondaryOutput captures the complete verbose output
// of the primary output, even when the command is not called
// in verbose mode. It provides a secondary output option.

var VerboseSecondaryOutput io.Writer
```

mlsandnerCI: That's great, thanks!

MarkusZoppelt (approved):
Thanks for the changes 👍🏼

internal/cmd/bundle/bundle.go
```diff
@@ -233,3 +232,36 @@ on the build system. This can be overridden with a docker-image flag.
 
 	return cmd
 }
+
+// SetUpBundleLogging configures the verbose log and build log file for the bundle command.
+func SetUpBundleLogging(cmd *cobra.Command, opts *bundler.Opts) error {
```
mlsandnerCI, line 237: @jochil I couldn't move this to the `cmdutils` package due to an import cycle.

PR #797, opened 2023-05-23
refactor(run): improve synced/not synced messages

mlsandnerCI:
### Description
Minor adjustments of our messages when we are synced/not synced during `run`. "Remote Fuzzing Server"/"CI Fuzz Server" have been replaced with "CI Sense".


integration-tests/shared/run_not_authenticated.go
```diff
@@ -55,5 +55,5 @@ func TestRunNotAuthenticated(t *testing.T, dir string, cifuzz string, args ...st
 	out, err := cmd.CombinedOutput()
 	require.NoError(t, err)
 
-	assert.Contains(t, string(out), "You are not authenticated with a remote fuzzing server.")
+	assert.Contains(t, string(out), "You are not authenticated")
```
mlsandnerCI, line 58: Not happy that we are testing against a hard coded string and that I had to adjust it just because I wanted to change "remote fuzzing server" to "CI Sense"... But it also doesn't make sense to put it into a const somewhere if we only use this string once so adjusted the test to only checks for the basic messages "You are authenticated" and "You are not authenticated".

florianGla (approved):

PR #800, opened 2023-05-25
fix: calculate number of new corpus entries correctly and make sure they don't underflow.

MarkusZoppelt:
### Motivation & Context
Ever since #746, the corpus entry count was wrongly calculated.

### Description
Revert the changes from 6bf86a2 and make sure that the new corpus entry variable does not underflow.


internal/cmd/run/reporthandler/reporthandler.go
```diff
@@ -273,6 +273,14 @@ func (h *ReportHandler) PrintFinalMetrics(numCorpusEntries uint) error {
 	totalCorpusEntries := numCorpusEntries
 	newCorpusEntries := totalCorpusEntries - h.numSeedsAtInit
 
+	// if the new corpus entries are negative: set them to 0 to avoid a uint
+	// underflow. this can happen if the total corpus entries are 0 (e.g. when
+	// running with `--engine-arg=-runs=10`) and cifuzz picks up new seeds at
+	// init during subsequent runs.
+	if newCorpusEntries > totalCorpusEntries {
```
mlsandnerCI, line 280: The comment says `if the new corpus entries are negative` but you are checking if the `newCorpusEntries` are larger than the `totalCorpusEntries`?

MarkusZoppelt: Thanks, the comment was confusing so I rephrased it.

internal/cmd/run/reporthandler/reporthandler.go
```diff
@@ -273,6 +273,15 @@ func (h *ReportHandler) PrintFinalMetrics(numCorpusEntries uint) error {
 	totalCorpusEntries := numCorpusEntries
 	newCorpusEntries := totalCorpusEntries - h.numSeedsAtInit
 
+	// If the number of new corpus entries exceeds the total corpus entries, it
+	// indicates an unexpected scenario where the total corpus entries are zero
+	// (e.g., when running with `--engine-arg=-runs=10`) and cifuzz discovers new
+	// seeds during subsequent runs. To avoid any issues related to unsigned
+	// integers, we set the new corpus entries to 0 in such cases.
```
jochil, line 280: I would expect it to be a valid scenario to have more new than existing corpus entries. For example the previous run was very short or was not able to make a lot progress, or am I missing something.

MarkusZoppelt: True, that might be the case, but this is not what we are checking here.

We need to check that the delta is indeed not a negative value. I am pretty confident that the code itself is correct because it fixes the issue, passes the newly introduced test and also helped [here](https://codeintelligence.slack.com/archives/C03R639E2JJ/p1685466220604149?thread_ts=1685465732.305329&cid=C03R639E2JJ).

Maybe the comment is still confusing...

jochil: Ok, but wouldn't we (in the outlined scenario & with the current implementation) set `newCorpusEntries` to `0`... this would be wrong, or?
Couldn't we just do the following?
```
if newCorpusEntries < 0 {
    newCorpusEntries = 0
}
```

MarkusZoppelt: That wouldn't work because newCorpusEntries is of type `uint`.
(same as `numCorpusEntries` which gets passed all the way down to this function)
That's why it's causing an underflow.

jochil: Ok, I get this now thanks

internal/cmd/run/reporthandler/reporthandler.go
```diff
@@ -273,6 +273,15 @@ func (h *ReportHandler) PrintFinalMetrics(numCorpusEntries uint) error {
 	totalCorpusEntries := numCorpusEntries
```
jochil, line 273: I think this assignment is unnecessary

jochil (approved):

PR #810, opened 2023-05-31
fix: missing = in coverage --help

JochenHiller:
Using `cifuzz coverage --help` shows how to generate HTML only:

```
HTML
    cifuzz coverage --output coverage-report <fuzz test>
```

But the text was missing an equal: correct is `--output=coverage-report`

### DIFF
--- a/internal/cmd/coverage/coverage.go
+++ b/internal/cmd/coverage/coverage.go
@@ -126,7 +126,7 @@ or a lcov trace file.
     cifuzz coverage <fuzz test>
 
 ` + pterm.Style{pterm.Reset, pterm.Bold}.Sprint("HTML") + `
-    cifuzz coverage --output coverage-report <fuzz test>
+    cifuzz coverage --output=coverage-report <fuzz test>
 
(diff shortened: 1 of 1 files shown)


MarkusZoppelt:
@JochenHiller Thank you for your contribution!

I just tried running without the equal sign, so I don't think this is necessary:

```
cifuzz coverage --output foobar -C examples/maven com.example.FuzzTestCase

[...]

✅ Created coverage HTML report: foobar
To view the report, open this URI in a browser:

   file:///Users/mz/code/cifuzz/examples/maven/foobar/index.html

```

Could you maybe provide an example where a missing `=` causes a problem? Thanks again!

JochenHiller:
My fault:

I did run coverage with wrong options `-output report` which did not work (due to missing `-`).
Then I changed to  `-output=report` which did work, but folder name was `utput=report`. I just saw that later when checking coverage results.

When using correct option either `--output` or `-o` it does work as described.

Sorry for inconvenience, next time I will test more thoroughly first.
I will close PR as well.

PR #816, opened 2023-06-01
Fix slow input cpp test on Windows

florianGla:
Note that the FuzzedDataProvider cannot be used in combination with <windows.h>. Specific input data generated by libfuzzer is not needed in order to test that asan reports a slow input finding.


internal/cmd/finding/finding_cmd.go
MarkusZoppelt, line 1: Why are changes to this file necessary to this PR?

florianGla: I noticed that sometimes not all stacktrace information is available. Without this change the finding command sometimes crashes.

MarkusZoppelt (approved):

PR #819, opened 2023-06-02
feat: '--style' flag to choose cifuzz style

mlsandnerCI:
### Motivation & Context
Make output not look weird on all kinds of terminals (especially windows).

### Description
Flag `--style` has been added to the root command with the following valid arguments: 

`--style=plain` or `--plain` to disable all colored output (EVEN IN LIBFUZZER REPORTS!) and icons:
Bildschirmfoto 2023-06-06 um 13 34 13

`--style=color` to have colored output but no icons:
Bildschirmfoto 2023-06-06 um 13 33 59

`--style=pretty`to have colored output and icons (this is default):
Bildschirmfoto 2023-06-06 um 13 33 35


pkg/log/log_test.go
```diff
@@ -16,7 +16,7 @@ var testOut io.ReadWriter
 func TestMain(m *testing.M) {
 	testOut = bytes.NewBuffer([]byte{})
 	Output = testOut
-	disableColor = true
+	viper.Set("color", "off")
```
MarkusZoppelt, line 19: This tests only previous behavior. Should we also test if the new flag works if set to each of the three new values?

pkg/log/log.go
```diff
@@ -171,3 +185,7 @@ func Printf(format string, a ...any) {
 func Print(a ...any) {
 	log(pterm.Style{pterm.FgDefault}, "", a...)
 }
+
+func Finding(a ...any) {
+	log(pterm.Style{pterm.FgDefault}, "💥", a...)
+}
```
MarkusZoppelt, line 191: This is only called once in the reporthandler, correct?

mlsandnerCI: That is true but this way the logic in the log pkg is used and I do not have to add a switch/case in the report handler to check for the color flag.

pkg/log/log_test.go
mlsandnerCI: Will do, thanks!

pkg/log/log.go
MarkusZoppelt: I see. I think it's ok.

MarkusZoppelt (approved):

PR #825, opened 2023-06-12
fix(jazzer): Do not hard-code `-rss_limit_mb`

fmeum:
Jazzer automatically sets `-rss_limit_mb` to a reasonable default at runtime, so there is no need to supply a hard-coded value on the command line.


jochil (approved):
thanks for taking care of this

PR #827, opened 2023-06-12
refactor: cleanup buildlog spinner functions

mlsandnerCI:
### Motivation & Context
I noticed when including these in the container command that the recurring if{} blocks made the code difficult to read and they were the same anyway.

### Description
Grouped "build in progress" functions together.


MarkusZoppelt (approved):
Nice cleanup!

PR #831, opened 2023-06-13
feat:  run fuzz tests inside container

jochil:
- [x] e2e tests for execute command
- [ ] e2e for container run - should be implemented once https://github.com/CodeIntelligenceTesting/cifuzz/pull/838 is merged
- [x] move container code from bundler
- [x] handle verbose output from container - streaming them out?


e2e-tests/execute/execute_test.go
```diff
@@ -0,0 +1,68 @@
...
+	// 	Description:  "execute command with a fuzz test argument in a folder with bundle contents runs the fuzz test",
+	// 	Command:      "execute",
+	// 	Args:         []string{"com.example.FuzzTestCase"},
+	// 	Environment:  []string{"CIFUZZ_PRERELEASE=true"},
+	// 	SampleFolder: []string{"folder-with-unpacked-bundle"},
+	// 	Assert: func(t *testing.T, output e2e.CommandOutput) {
+	// 		// TODO: should fail! Execute doesn't respect the libfuzzer findings today
+	// 		output.Success().ErrorContains("Security Issue: Remote Code Execution in exploreMe (com.example.ExploreMe:19)")
+	// 	},
+	// },
```
MarkusZoppelt, line 53: should we let that comment stay this way?

e2e-tests/samples/folder-with-unpacked-bundle/bundle.yaml
```diff
@@ -0,0 +1,25 @@
+run_environment:
+    docker: openjdk:latest
```
MarkusZoppelt, line 2: we should go for `eclipse-temurin` wherever possible

internal/cmd/container/run/run.go
```diff
@@ -1,24 +1,174 @@
...
-func newWithOptions() *cobra.Command {
+func newWithOptions(opts *containerRunOpts) *cobra.Command {
+	var bindFlags func()
+
 	cmd := &cobra.Command{
-		Use:   "run",
-		Short: "Build and run a Fuzz Test container image locally",
+		Use:               "run",
+		Short:             "Build and run a Fuzz Test container image locally",
+		Long:              "", //TODO
```
MarkusZoppelt, line 50: We should add a long description, especially since this is totally new for users

internal/cmd/container/run/run.go
```diff
@@ -1,24 +1,174 @@
...
+	containerID, err := c.buildContainerFromImage()
+	if err != nil {
+		logging.StopBuildProgressSpinnerOnError(log.ContainerBuildInProgressErrorMsg)
+		return err
+	}
+
+	logging.StopBuildProgressSpinnerOnSuccess(log.ContainerBuildInProgressSuccessMsg, false)
+
+	logging.StartBuildProgressSpinner(log.ContainerRunInProgressMsg)
+	// TODO: make me nice
```
MarkusZoppelt, line 123: is it?

internal/cmd/execute/execute.go
```diff
@@ -0,0 +1,192 @@
...
+	*cobra.Command
+	opts *executeOpts
+}
+
+func New() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "execute",
+		Short:   "tbd",
+		Long:    `tbd`,
+		Example: "tdb",
```
MarkusZoppelt, line 35: shia

internal/container/image.go
```diff
@@ -0,0 +1,198 @@
...
+		return errors.WithStack(err)
+	}
+
+	return nil
+}
+
+func copyCifuzz(buildContextDir string) error {
+	// Add the CIFuzz binaries to the bundle if the version is "dev".
+	// TODO: this should work even if internal doesn't exist
+	exists, err := fileutil.Exists("../../build/bin")
```
MarkusZoppelt, line 180: We need a ticket for this

MarkusZoppelt: CLI-1100

e2e-tests/execute/execute_test.go
JackuB: FWIW this should get addressed once https://github.com/CodeIntelligenceTesting/cifuzz/pull/838 is also merged. As that will allow us to pull Java dependency for a single test case running in a container

internal/cmd/container/run/run.go
jochil: i guess nice enough for now

e2e-tests/samples/folder-with-unpacked-bundle/bundle.yaml
jochil: done :+1:

jochil:
> Calling the command with `-v` is not working for Markus and me. Bildschirmfoto 2023-06-21 um 16 03 05
> 
> Apparently the `jsonmessage.DisplayJSONMessagesStream()` function fails.

I had the same error... it happens when the cifuzz binaries are not available. It worked after using `make build` before

jochil:
> > I had the same error... it happens when the cifuzz binaries are not available. It worked after using `make build` before
> 
> Maybe we can wrap this error in a suitable error message? Or prevent this somehow? I would like for the other teams who want to try out the command to not run into the same errors and then wonder what went wrong. Wrapping/preventing this would also mean less support work for us.

IMO we should soon find a better overall solution for including the right executable into the image... in the meeting yesterday it became clear that this is an unnecessary hurdle to get started with container.
So maybe it is ok for this PR

MarkusZoppelt (approved):

PR #839, opened 2023-06-15
feat(login): improve invalid token handling

MarkusZoppelt:
### Motivation & Context

When tokens are invalidated (deleted) on CI Sense, users need to manually edit the token file to remove them. Users should be able to remove these tokens via the CLI directly.

### Description
This change prompts users for a new token if the token in the `access_tokens.json` file looks invalid.


internal/cmdutils/auth/auth.go
```diff
@@ -24,10 +28,21 @@ func GetAuthStatus(server string) (bool, error) {
 	err := login.CheckValidToken(&apiClient, token)
 	if err != nil {
 		log.Warnf(`Failed to authenticate with the configured API access token.
```
mlsandnerCI, line 30:
```suggestion
		log.Warn(`Failed to authenticate with the configured API access token.
```

internal/cmdutils/auth/auth.go
```diff
@@ -24,10 +28,21 @@ func GetAuthStatus(server string) (bool, error) {
...
-It's possible that the token has been revoked. Please try again after
-removing the token from %s.`, tokenstorage.GetTokenFilePath())
+It's possible that the token has been revoked.`)
 
-		return false, err
+		if viper.GetBool("interactive") && term.IsTerminal(int(os.Stdin.Fd())) {
+			tryAgain, err := dialog.Confirm("Do you want to log in again?", true)
```
mlsandnerCI, line 34: The two parts you changed look very similar, maybe you can move this logic into a function? Generally I would wait for a third occurrence but this has a dialog/error message directed at the user and I assume we always want them to be the same so it would make sense to not have them in two places.

MarkusZoppelt: Yes, that's what I was thinking about, too..
I will spend some more time thinking about a clever solution. If I don't find one, I will leave it as is.

MarkusZoppelt: We included a refactor of the `auth` and `login` package. Now it should be cleaner.

internal/cmdutils/auth/auth.go
```diff
@@ -23,11 +27,22 @@ func GetAuthStatus(server string) (bool, error) {
...
 	err := login.CheckValidToken(&apiClient, token)
 	if err != nil {
-		log.Warnf(`Failed to authenticate with the configured API access token.
-It's possible that the token has been revoked. Please try again after
-removing the token from %s.`, tokenstorage.GetTokenFilePath())
+		log.Warn(`Failed to authenticate with the configured API access token.
+It's possible that the token has been revoked.`)
 
-		return false, err
+		if viper.GetBool("interactive") && term.IsTerminal(int(os.Stdin.Fd())) {
```
jochil, line 129: same here with `c.InOrStdin()`

jochil (approved):

jochil:
I know they the E2E behavior tests are not required but it looks like a "valid" fail: https://github.com/CodeIntelligenceTesting/cifuzz/actions/runs/5331837609

WDYT?

MarkusZoppelt:
> I know they the E2E behavior tests are not required but it looks like a "valid" fail: https://github.com/CodeIntelligenceTesting/cifuzz/actions/runs/5331837609
> 
> WDYT?

Should get fixed by #846. (can confirm it works locally)
As soon that PR gets merged, we can try the pipelines again...

internal/cmdutils/auth/auth.go
MarkusZoppelt: After moving all of this into `auth.go` (not a cobra cmd...) I didn't find an easy solution for this.
Maybe we best revisit this in another PR / ticket

jochil: It's not a blocker for me

PR #844, opened 2023-06-19
fix(jazzer): recognize jazzer parameter in fuzz test identification

mlsandnerCI:
### Description

Adjusted the regex pattern to recognize `@FuzzTest` and `@FuzzTest(...)` and included the case in the unit test. Since it now handled different cases, the test was refactored to use a test case structure.


jochil (approved):
nice tests :+1:

PR #847, opened 2023-06-20
fix(coverage): support new (2.0) version of genhtml

mlsandnerCI:
### Motivation & Context
Apparently using a newer version of genhtml/lcov breaks our current implementation  

### Description
We removed the `--prefix` flag for the genhtml command. Tests show that we do not need it to run the command and not using it resolves issue between the different genhtml versions


florianGla (commented):
We should probably also remove the prefix flag for bazel.

jochil:
> We should probably also remove the prefix flag for bazel.

good point... done :+1:

florianGla (approved):

PR #851, opened 2023-06-21
Update jazzer to version 0.19.0

florianGla:


florianGla:
> I think `pkg/messaging/instructions/gradlekotlin` is missing the plugin update. There is a small tool that helps updating all occurrences of the gradle plugin: `tools/updater/updater.go`

Good catch, done!

jochil (approved):

PR #852, opened 2023-06-21
Fix cifuzz always sending a timeout flag

br-lewis:
This changes `Runner`'s behavior to only add `-max_total_time` to the Jazzer call when `Timeout` is not 0.

### Motivation & Context
`cifuzz` was always sending `-max_total_time` even if `cifuzz` was not called with a `--timeout` flag. In that case, it would use `-max_total_time=0` which would cause Jazzer to ignore a test's configured time out and run forever.

### Description
Prior to this change, `cifuzz` would always add `-max_total_time` with the value from `Timeout` and, due to Go initializing all variables with their 0 values, if `--timeout` is not added then the `Timeout` that `Runner` uses will be 0. Jazzer forwards those flags directly to libfuzzer with the CLI args taking precedence over defaults. Libfuzzer would apparently treat a 0 as telling it to run forever so the timeouts of the fuzz tests would be ignored and fuzzing would never stop.

This would prevent someone from explicitly setting `--timeout=0` though I'm not sure that's a sensible thing to want to do anyway.


MarkusZoppelt (approved):
Looks good!

Thank you for your contribution @br-lewis!

PR #853, opened 2023-06-21
refactor: optimize cleanup in tests

jochil:
by adding and improving helper methods for creating and cleaning up directories


internal/testutil/testutil.go
```diff
@@ -98,3 +98,12 @@ func CheckOutput(t *testing.T, r io.Reader, s ...string) {
 		require.Contains(t, string(output), str)
 	}
 }
+
+// MkdirTemp wraps oss.MkdirTemp and makes sure that errors are checked and
```
MarkusZoppelt, line 102:
```suggestion
// MkdirTemp wraps os.MkdirTemp and makes sure that errors are checked and
```

integration-tests/errors/java/java_errors_test.go
```diff
@@ -20,12 +20,10 @@ func TestIntegration_JavaErrors(t *testing.T) {
...
 	testdataTmp := shared.CopyTestdataDir(t, "java_cpp")
-	t.Cleanup(func() { fileutil.Cleanup(testdataTmp) })
 
 	// This testdata is in a separate project due to external dependencies that
 	// need to be downloaded for compilation and slow down every testcase.
 	testdataLDAPAndSQLTmp := shared.CopyCustomTestdataDir(t, "testdata-sql-ldap", "java_cpp_sql_ldap")
-	t.Cleanup(func() { fileutil.Cleanup(testdataLDAPAndSQLTmp) })
 
 	installDir := shared.InstallCIFuzzInTemp(t)
 	t.Cleanup(func() { fileutil.Cleanup(installDir) })
```
MarkusZoppelt, line 29: We keep this "old" cleanup because it is our install dir... is this right?

MarkusZoppelt (approved):
Nice!

integration-tests/errors/java/java_errors_test.go
jochil: Yes... I checked `InstallCIFuzzInTemp` but the comment is saying
> The temporary directory should *not* be deleted by the caller because it is shared between tests and deleting it would cause side effects.

so I didn't want to risk more side effects

PR #856, opened 2023-06-22
Include version/os information in verbose mode

adombeck:
Print all the [Basic Information](https://code-intelligence.atlassian.net/wiki/spaces/CI/pages/325877781/Development+how+to+quickly+solve+customer+issues#Basic-Information) in the verbose output. The only info that was missing is the OS and architecture and the gradle version, everything else was already printed.


adombeck:
Closing in favor of https://github.com/CodeIntelligenceTesting/cifuzz/pull/858 (which uses the correct branch prefix)

adombeck:
Closed the wrong PR, I meant to close https://github.com/CodeIntelligenceTesting/cifuzz/pull/857

pkg/dependencies/version.go
```diff
@@ -165,6 +168,20 @@ func javaVersion(dep *Dependency) (*semver.Version, error) {
 	return version, nil
 }
 
+func gradleVersion(dep *Dependency, projectDir string) (*semver.Version, error) {
+	path, err := gradle.GetGradleCommand(projectDir)
```
jochil, line 172: I tested it quickly and I think we can use `os.Getwd` instead of the "real" project dir... what would make the change of the function signatures unnecessary
I can't think of a scenario where at least  `gradlew` is not in the working directory or in one parent directory

jochil:
There a no new or changed tests... maybe we can think about some test cases. Should be easy to implement some with the e2e behavior tests

pkg/dependencies/version.go
adombeck: We support setting the project dir via the `--project-dir` flag to a directory which is not the working directory and not a parent directory of it, so I think we have to actually look in the project dir to not break use cases.

jochil: of course... you're right. Do you know if we have a test case using the `--projectDir` flag?

adombeck: I can't find any test using `--project-dir`

adombeck:
I added a test

jochil:
I made the handling of `--version` a bit simpler... also the unit tests are now :heavy_check_mark:  again

PR #857, opened 2023-06-22
fix(installer): Fix zsh completion script symlink loop

adombeck:
Consider the following scenario:

1. The installer fails to create a symlink to the first fpath directory
2. User follows the instructions to add the completion script to their fpath
3. The user runs a cifuzz installer again
4. The installer creates a symlink from the first fpath to the completion script.

The symlink created in step 4 creates a symlink loop, because the first fpath is the path to the completion script, so we create a symlink to itself. In effect, zsh hangs whenever the user opens a new shell.

This commit avoids creating such a symlink loop.


adombeck:
Closing in favor of https://github.com/CodeIntelligenceTesting/cifuzz/pull/858 (which uses the correct branch prefix)

PR #859, opened 2023-06-23
feat(container): Copy cifuzz executable from container image

adombeck:


.github/workflows/pipeline_release.yml
jochil, line 1: if we want to enforce a formatting style for yaml files, the formatter should be part of the repo and not some abitrary one from the core repo

.github/workflows/workflow_container_image.yml
```diff
@@ -0,0 +1,39 @@
+on:
+  workflow_call:
+    inputs:
+      version:
+        required: true
+        type: string
+      compatibility_check:
+        required: true
+        type: string
```
jochil, line 9: this input is not used... I think it can be removed here

Makefile
```diff
@@ -37,6 +37,19 @@ project := "code-intelligence.com/cifuzz"
 # make version=1.0.0-dev [target]
 version = dev
 
+# Set IMAGE_ID to ghcr.io/codeintelligencetesting/cifuzz if it's not set
+image_id ?= ghcr.io/codeintelligencetesting/cifuzz
```
jochil, line 41: I think the registry URI might be better placed inside pipeline and configured by a github action variable. This way we can change the destination without code changes

.github/workflows/workflow_container_image.yml
```diff
@@ -0,0 +1,39 @@
...
+      contents: read
+      packages: write
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Build and push
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
jochil, line 25: is there any risk of exposing the `GITHUB_TOKEN` when using it as env?

.github/workflows/pipeline_release.yml
```diff
@@ -53,3 +53,11 @@ jobs:
           draft: true
           files: cifuzz_installer_*
           fail_on_unmatched_files: true
+
+  tag_container_image:
```
jochil, line 57: I forgot this when we were talking about the release pipeline: If we place the image upload here, it will be released before the actual release, as we just create a draft release to manually release it later... 

If we want to avoid this (not sure if it is important now). There is also an event for running a pipeline after the "real" release
```
on:
  release:
    types: [published]
```

jochil (commented):
This looks pretty straight forward... I like it. 

What's the behavior if there is already an image inside the registry with the same tag?

.github/workflows/pipeline_release.yml
adombeck: > What's the behavior if there is already an image inside the registry with the same tag?

It gets overwritten

adombeck: > Do you have already an idea on how to include a cifuzz build from source?

What do you mean? If you're talking about the local development workflow, running `make container-image` will create a local container image. When you then execute the `cifuzz` executable that was built by that command (by the `build` target which the `container-image` depends on) it will use that local container image.

.github/workflows/pipeline_release.yml
adombeck: > if we want to enforce a formatting style for yaml files, the formatter should be part of the repo and not some abitrary one from the core repo

I'm in favor of using a formatter. The reason why my IDE formatted that file is that I configured it according to the [core/README.md](https://gitlab.code-intelligence.com/code-intelligence/core/-/blob/master/README.md?plain=1#L71-L85). I don't think it's possible to configure that on a per-project basis in Intellij, so I would have to find another solution to enforce formatting in the core project and not also have the formatter run when working on the cifuzz repo.

Makefile
adombeck: In the GitHub Actions we use the value from the GitHub pre-defined variable `github.repository`: https://github.com/CodeIntelligenceTesting/cifuzz/blob/CLI-1100-cifuzz-in-container-image/.github/workflows/workflow_container_image.yml#L25-L25

This here is just a default value which is helpful when a developer wants to execute the make target on their local machine instead of in the GitHub Action (which is useful for debugging and might also be handy for hot fixes).

.github/workflows/workflow_container_image.yml
adombeck: I don't think so, configuring secrets in the CI environment is pretty standard in my experience. And they are still redacted in the GitHub action logs.

.github/workflows/workflow_container_image.yml
adombeck: fixed, thanks!

jochil:
> > Do you have already an idea on how to include a cifuzz build from source?
> 
> What do you mean? If you're talking about the local development workflow, running `make container-image` will create a local container image. When you then execute the `cifuzz` executable that was built by that command (by the `build` target which the `container-image` depends on) it will use that local container image.

okay, I overlooked the part with default version = "dev"... no everything makes sense

Makefile
jochil: ok... I agree, we should definitely have a default value in the makefile. But for the pipeline part it would be nice to change it from the "outside" as we might use some azure registry in the future

.github/workflows/pipeline_release.yml
jochil: > I forgot this when we were talking about the release pipeline: If we place the image upload here, it will be released before the actual release, as we just create a draft release to manually release it later...
> 
> If we want to avoid this (not sure if it is important now). There is also an event for running a pipeline after the "real" release
> 
> ```
> on:
>   release:
>     types: [published]
> ```

What do you think about this?

.github/workflows/pipeline_release.yml
jochil: so maybe like this: https://github.com/CodeIntelligenceTesting/cifuzz/pull/865 ?

jochil (approved):

.github/workflows/pipeline_release.yml
adombeck: I added a commit where I try to use it but I can't test it without publishing a release. Maybe we should just merge it and check after the next release if it worked?

.github/workflows/pipeline_release.yml
adombeck: Lets follow up on this on #865

Makefile
adombeck: Isn't it easy enough to change the pipeline yaml in case that we change the registry in the future?

PR #861, opened 2023-06-23
Support target methods in java bundles

florianGla:
Java bundles need to define target methods in order to allow the execution of specific fuzz test methods in java classes.


integration-tests/shared/bundle.go
jochil, line 1: I think this commit can just be part of the previous one

internal/build/build.go
```diff
@@ -12,6 +12,8 @@ import (
 type Result struct {
 	// A name which uniquely identifies the fuzz test and is a valid path
 	Name string
+	// Optional name for the fuzz test target method
+	TargetMethod string
```
jochil, line 16: I think it is ok for now, but in general I think it might be confusing having `targetMethod` but "hiding" `TargetClass` inside the `Name`

jochil (approved):

internal/build/build.go
florianGla: Yes, you are right. We should refactor that. The name field is more suitable for c++ fuzz tests.

PR #863, opened 2023-06-23
chore(githooks): allow to push with uncommitted changes

jochil:
there are plenty of reasons you want to push while having files uncommited/untracked, for example:
-  files for testing you created locally that should not be part of the repo and `.gitignore`
- changes you want to continue working on
- ...

it forces you to use `git stash` all the time and/or remove untracked files... while the benefit is close to none. If someone forgets to add something it can be easily done later


githooks/pre-push
```diff
@@ -10,11 +10,6 @@ set -u
 
 GIT_ROOT=$(git rev-parse --show-toplevel)
 
-if [ -n "$(git status --porcelain)" ]; then
-    echo >&2 "Error: The working tree has uncommitted changes."
-    exit 1
-fi
-
```
adombeck, line 17: This commit will not allow pushing with uncommitted changes, because we exit in two other places in the same script if the working tree is unclean. We do that to check if formatting / running `go mod tidy` changes any files. I don't see a way to reliably check if files are correctly formatted which also allows to push with uncommited changes.

adombeck:
> * files for testing you created locally that should not be part of the repo and .gitignore

Would adding those files to your `.git/info/exclude` be an option?

> * changes you want to continue working on

That sounds like a classic `git stash` use case.

>  while the benefit is close to none

Unless you see another way to enforce our formatting, I disagree that the benefit is close to none.

githooks/pre-push
jochil: I left the other ones on intent to make sure that formatting and linting will be enforced... it just just about things I didn't staged/added on purpose, but as they use `git status --porcelain` it will still fail.

I will think about a workaround... for example `make fmt/check`

jochil:
> > * files for testing you created locally that should not be part of the repo and .gitignore
> 
> Would adding those files to your `.git/info/exclude` be an option?

extra steps and (at least in my case) there is no pattern, for example: adding files to a example project, creating a new package, ...
> 
> > * changes you want to continue working on
> 
> That sounds like a classic `git stash` use case.

What is what I usually have to do, but only because you git hook enforces it, not because it fits my workflow
 
> > while the benefit is close to none
> 
> Unless you see another way to enforce our formatting, I disagree that the benefit is close to none.

We are checking it in the push pipeline as well...so nothing unformatted will get merged. 
But my point is not about skipped format/lint checks, it's about blocking `git push` (which purpose is to push commits) blocked by the current state of the stage.
An alternative could be a moving the checks to a pre-commit hook or using checks that will not mess with the git stage

adombeck:
How about we ignore untracked files in the pre-push hook? I created a PR for that: https://github.com/CodeIntelligenceTesting/cifuzz/pull/870

mlsandnerCI (approved):

MarkusZoppelt:
> In general we should still think about using pre-commit hooks instead

Please, let's not use pre-commit hooks. I find them extremely tedious.

PR #865, opened 2023-06-23
style: adding formatter for various file types

jochil:
... because we don't have one by now


MarkusZoppelt (changes requested):
This change deleted all content from our `**/cifuzz.yaml` files.
(I assume because they only contain comments and nothing else?)

jochil:
> This change deleted all content from our `**/cifuzz.yaml` files. (I assume because they only contain comments and nothing else?)

:facepalm: I will take a look if there is a workaround/flag/... to handle this

jochil:
I switched to prettier and applied the formatting to all types it supports (md, js, ...).
It's widely used (in the core repo, too) and it should be easy for the IDEs to pick up the config

MarkusZoppelt (approved):

PR #867, opened 2023-06-26
feat: add support for TypeScript

MarkusZoppelt:
### Motivation/Context

This PR introduces support for TypeScript projects. Users need the `CIFUZZ_PRERELEASE=1` flag set to use it.

### Description

- a TypeScript example in `examples/node-typescript`
- support for `cifuzz init` in TypeScript projects
- support for `cifuzz create` for TypeScript projects

The `init` command will show a selection TUI when run from a `node` environment and ask users wether they want to initialise a JavaScript or TypeScript project (because the instructions differ). The init command now also accepts arguments `cpp|java|kotlin|js|ts` to pass the test language, e.g. in non-interactive mode.

### How to use/reproduce

1. make sure you set `CIFUZZ_PRERELEASE=1`
2. cd to `examples/node-typescript`
3. `rm cifuzz.yaml && cifuzz init` (shows TUI) or
4. `rm cifuzz.yaml && cifuzz init ts`
5. `cifuzz create` and choose `TypeScript` to create a TypeScript fuzz test stub


internal/cmd/init/init.go
```diff
@@ -24,24 +28,37 @@ const (
...
 	BuildSystem string
+	Interactive bool   `mapstructure:"interactive"`
 	Server      string `mapstructure:"server"`
 	Project     string `mapstructure:"project"`
+	testLang    config.FuzzTestType
+}
+
+// for node, we support both js and ts so we need a map of supported test langs
+// -> label:value
+var supportedProjectLangs = map[string]string{
```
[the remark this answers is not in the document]
MarkusZoppelt: Good point. I added `func SupportedTestTypes()` to the `config` package that also checks for supported langs that are available with `CIFUZZ_PRERELEASE`

internal/cmd/init/init_test.go
```diff
@@ -29,3 +29,28 @@ func TestInitCmd(t *testing.T) {
 	assert.Error(t, err)
 	assert.ErrorIs(t, err, cmdutils.ErrSilent)
 }
+
+// TestInitCmdForNode tests the init command for Node projects (both JavaScript and TypeScript).
+func TestInitCmdForNodeWithLanguageArg(t *testing.T) {
```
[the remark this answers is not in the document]
MarkusZoppelt: Done. I would like to create tests for interactive input in a different PR where we add tests for all references of the dialog/etc package.

pkg/stubs/test.fuzz.ts.tmpl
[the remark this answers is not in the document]
MarkusZoppelt: Done

internal/cmd/create/create_test.go
```diff
@@ -81,7 +81,7 @@ func TestOkJavaScript(t *testing.T) {
 		t.Skip("Skipping test because CIFUZZ_PRERELEASE is not set.")
 	}
 
-	testDir, cleanup := testutil.BootstrapExampleProjectForTest("create-cmd-test", config.BuildSystemNode)
+	testDir, cleanup := testutil.BootstrapExampleProjectForTest("create-cmd-test", config.BuildSystemNode+"js")
```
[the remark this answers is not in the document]
MarkusZoppelt: Fixed

internal/cmd/init/init.go
```diff
@@ -62,6 +79,17 @@ func New() *cobra.Command {
 				return cmdutils.WrapSilentError(err)
 			}
 
+			if opts.Interactive {
+				opts.Interactive = term.IsTerminal(int(os.Stdin.Fd())) && term.IsTerminal(int(os.Stdout.Fd()))
```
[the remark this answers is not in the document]
MarkusZoppelt: I am not sure how. Can you explain what you mean?

`cmd.InOrStdin()` and `cmd.OutOrStdout()` are both of type `*bytes.Buffer` and do not have an `.Fd()` field.

I tried 
```go
if opts.Interactive {
  stdinFd := int(cmd.InOrStdin().(*os.File).Fd())
  stdoutFd := int(cmd.OutOrStdout().(*os.File).Fd())
  opts.Interactive = term.IsTerminal(stdinFd) && term.IsTerminal(stdoutFd)
}
```
but that [ended badly](https://github.com/CodeIntelligenceTesting/cifuzz/actions/runs/5464535945/jobs/9946702430?pr=867).

internal/cmd/create/create.go
```diff
@@ -42,7 +42,7 @@ func (opts *createOpts) Validate() error {
 	if !opts.Interactive && opts.testType == "" {
 		err := errors.New("Missing argument [cpp|java|kotlin]")
 		if os.Getenv("CIFUZZ_PRERELEASE") != "" {
-			err = errors.New("Missing argument [cpp|java|kotlin|js]")
+			err = errors.New("Missing argument [cpp|java|kotlin|js|ts]")
```
mlsandnerCI, line 45: Maybe we should make this message more dynamic with a map so we don't always have to extend it? Don't we already have a list of supported languages?

internal/config/config.go
```diff
@@ -26,7 +26,7 @@ import (
 const (
 	BuildSystemBazel  string = "bazel"
 	BuildSystemCMake  string = "cmake"
-	BuildSystemNodeJS string = "nodejs"
+	BuildSystemNodeJS string = "node"
```
mlsandnerCI, line 29: Should the const also be renamed to `BuildSystemNode`?

internal/config/types.go
```diff
@@ -21,3 +31,13 @@ type Engine string
 const (
 	Libfuzzer Engine = "libfuzzer"
 )
+
+// SupportedTestTypes returns the supported test types depending on the
+// environment variable CIFUZZ_PRERELEASE.
+func SupportedTestTypes() map[string]string {
```
mlsandnerCI, line 37: I like this function, this makes the prerelease checks definitely easier!

pkg/messaging/instructions.go
```diff
@@ -24,14 +24,19 @@ var gradleKotlinSetup string
...
+var nodetsSetup string
+
 func Instructions(buildSystem string) string {
 	switch buildSystem {
 	case config.BuildSystemBazel:
 		return bazelSetup
 	case config.BuildSystemCMake:
 		return cmakeSetup
-	case config.BuildSystemNodeJS:
+	case "nodejs":
```
mlsandnerCI, line 36: Should we have a const for this?

internal/config/config.go
MarkusZoppelt: Maybe. The whole naming and constants regarding NodeJS with JS/TS is a bit of a mess.

I had named it `BuildSystemNode` before.

See also https://github.com/CodeIntelligenceTesting/cifuzz/pull/867#discussion_r1251929308

MarkusZoppelt:
> I am a bit confused about the naming. As I understood it "Node.js" is only for javascript projects and it's "Node" when talking about typescript projects? This would mean we want to generally say we support "node projects" and not "node.js projects"? 

You are not alone

See also https://github.com/CodeIntelligenceTesting/cifuzz/pull/867#discussion_r1251929308

internal/cmd/init/init.go
```diff
@@ -62,6 +69,17 @@ func New() *cobra.Command {
...
+			}
+
+			if len(args) == 1 {
+				opts.testLang = config.FuzzTestType(args[0])
+			}
+			if !opts.Interactive && opts.BuildSystem == config.BuildSystemNodeJS && opts.testLang == "" {
+				err := errors.New("cifuzz init requires a test language for node projects [js|ts]")
```
mlsandnerCI, line 80:
```suggestion
				err := errors.New("cifuzz init requires a test language for node.js projects [js|ts]")
```

internal/config/config.go
mlsandnerCI: Ah okay. But why do we work with `node` now internally?

MarkusZoppelt: Fixed

pkg/messaging/instructions.go
MarkusZoppelt: Changed

internal/cmd/create/create.go
MarkusZoppelt: Good catch! Changed to 
```go
err := errors.New(fmt.Sprintf("Missing argument [%s]", strings.Join(maps.Values(config.SupportedTestTypes()), "|")))
```

internal/cmd/init/init.go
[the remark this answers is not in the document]
jochil: I expected `cmd.InOrStdin()` and `os.Stdin()` to be interchangeable... if this is not the case this will of course not work. Maybe we can rethink the "terminal" check... as it is also used in multiple places. But it does not have to be part of this PR

jochil (approved):

internal/cmd/init/init_test.go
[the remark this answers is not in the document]
jochil: Is there a task for it in the backlog?

PR #872, opened 2023-06-27
feat(bundle): check for multiple target methods if argument is class 

mlsandnerCI:
### Motivation & Context
Users want to build all fuzz tests in one class when the class is given as an argument to `bundle` or `remote-run`.

### Description
Calling `cifuzz bundle com.example.FuzzTestCase` now checks if the class has multiple fuzz tests inside and adds them to the bundle. 
Also: rework of how we extract the target method. I moved the code from the PreRun function of `bundle` and `remote-run` to the `jazzer_bundler`. There we handle all cases to get the fuzz tests and their target methods. A test has been added to check that.


internal/bundler/jazzer_bundler.go
```diff
@@ -327,3 +292,88 @@ func (b *jazzerBundler) createManifestJar(targetClass string, targetMethod strin
 
 	return jarPath, nil
 }
+
+// getAllFuzzTestsAndTargetMethodsForBuild extracts all fuzz tests and their target
+// methods from the fuzz tests given to the bundler.
+func (b *jazzerBundler) getAllFuzzTestsAndTargetMethodsForBuild() ([]string, []string, error) {
```
jochil, line 298: What do you thing about `FuzzTestIdentifier()` as a method name?

jochil (approved):
nice cleanup!

Do you think it makes sense to change it here as well? https://github.com/CodeIntelligenceTesting/cifuzz/pull/859/commits/3f5bdf13e80cde3bdb8ed746b1439cc486cc8f15

PR #876, opened 2023-06-28
Provide more info for the exit code 127 ldd error

jochil:
### Motivation/Context
We saw a lot of `exit status 127` errors in the past without knowing which shared object was broken/missing. This change should provide the missing information.

### Description
+ temporarily using fork of `u-root` until this PR is merged https://github.com/u-root/u-root/pull/2710
+ adding some kind of "test"

### How to use/reproduce
you can run a test triggering the error inside the `ldd` package by (this only works on linux):
```
go test ./internal/ldd/...  -v
```
To create a broken executable we used the following command:
```
patchelf --replace-needed libm.so.6 foobar.so my_fuzz_test
```


internal/ldd/ldd_unix_test.go
```diff
@@ -0,0 +1,29 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TODO: determine if this is needed
+// The reason for this test is to trigger an error in external library
+// we were facing a lot recently
+// It is more about reproducing the error and ensuring helpful a helpful
+// error message than actually testing our code
```
florianGla, line 17: Just a typo. Helpful is mentioned twice :)

florianGla (approved):

PR #883, opened 2023-06-30
Add run command for node projects

florianGla:
### Description

This PR introduces functionality for `cifuzz run` for Node projects (currently only JavaScript, TypeScript support will follow soon - maybe even in this PR)

Ideally, this requires #867 and #874 to be merged first, since some changes are required to finish / clean up this PR.

### How to use/reproduce

1. `cd examples/nodejs`
2. `npm i` installs all node dependencies
3. `cifuzz run FuzzTestCase`


integration-tests/nodejs/nodejs_test.go
```diff
@@ -0,0 +1,180 @@
...
+
+	"code-intelligence.com/cifuzz/integration-tests/shared"
+	builderPkg "code-intelligence.com/cifuzz/internal/builder"
+	"code-intelligence.com/cifuzz/internal/testutil"
+	"code-intelligence.com/cifuzz/pkg/parser/libfuzzer/stacktrace"
+	"code-intelligence.com/cifuzz/util/fileutil"
+)
+
+func TestIntegration_NodeJS_InitCreateRun(t *testing.T) {
+	if testing.Short() || os.Getenv("CIFUZZ_PRERELEASE") == "" {
```
JackuB, line 23: Why do we check for a presence of CIFUZZ_PRERELEASE envvar for a test?

examples/node-typescript/package.json
```diff
@@ -2,6 +2,7 @@
 	"name": "node-typescript",
 	"scripts": {
 		"build": "tsc",
+		"test": "jest",
 		"dryRun": "jest",
```
JackuB, line 6: Do we want to keep the `dryRun` if we already have `test`?

integration-tests/node-typescript/nodets_test.go
```diff
@@ -0,0 +1,182 @@
...
+			return strings.Join(instructions[i:len(instructions)-1], "\n")
+		}
+	}
+
+	return ""
+}
+
+func modifyFuzzTestToCallFunction(t *testing.T, fuzzTestPath string) {
+	// Modify the fuzz test stub created by `cifuzz create` to actually
+	// call a function.
```
JackuB, line 144: Should these comments be above the function name to be picked up by IDEs?
```suggestion
// Modify the fuzz test stub created by `cifuzz create` to actually call a function.
func modifyFuzzTestToCallFunction(t *testing.T, fuzzTestPath string) {
```

pkg/parser/libfuzzer/stacktrace/stacktrace.go
```diff
@@ -17,6 +17,9 @@ var framePattern = regexp.MustCompile(
 // Special pattern for Java stack traces
 var framePatternJava = regexp.MustCompile(`\sat\s(?P<source_file>\S+)[.](?P<function>\S+[^<>])[(]\S+:(?P<line>\d+)`)
 
+// Special pattern for Node stack traces
+var framePatternNode = regexp.MustCompile(`\s*at\s((?P<function>\S+)\s+(\[.*\])?\s*\()?(?P<source_file>\S+?):(?P<line>\d+):?(?P<column>\d*)\)?`)
```
JackuB, line 21: What's the reason for the lazy eval in the `source_file` part? The question mark in "\S+**?**" causes the `+` to be lazy instead of greedy by default?

```
(?P<source_file>\S+?):(?P<line>\d+)
```

integration-tests/nodejs/nodejs_test.go
MarkusZoppelt: So that it actually runs. If the env var is not set, the commands will only print

`cifuzz does not support Node.js projects yet.`

examples/node-typescript/package.json
MarkusZoppelt: Fixed

internal/completion/fuzz_test_completion.go
```diff
@@ -177,6 +179,16 @@ func validJVMFuzzTests(projectDir string, toComplete string) ([]string, cobra.Sh
 	return fuzzTests, cobra.ShellCompDirectiveNoFileComp
 }
 
+// validJVMFuzzTests returns a list of valid JVM fuzz test identifiers
```
mlsandnerCI, line 182:
```suggestion
// validNodeFuzzTests returns a list of valid JVM fuzz test identifiers
```

pkg/finding/finding.go
```diff
@@ -258,6 +261,8 @@ func (f *Finding) ShortDescriptionColumns() []string {
 		case strings.Contains(f.Details, "Security Issue:"):
 			// Jazzer findings
 			errorType = f.Details
+		case f.Details == "fuzz target exited":
+			errorType = f.Details
```
mlsandnerCI, line 265: This could be combined since they both do the same thing.

mlsandnerCI: Or did you want to keep it separate because they are for different fuzzers, Jazzer and Jazzer.js? Because then I would add a comment like `//Jazzer findings` above for the Jazzer.js one.

pkg/runner/jazzerjs/jazzerjs_runner.go
```diff
@@ -0,0 +1,97 @@
...
+	var args []string
+	args = append(args, "npm", "test")
+
+	// --------------------------
+	// --- JazzerJS options -----
+	// --------------------------
+
+	// ---------------------------
+	// --- libfuzzer options -----
+	// ---------------------------
```
mlsandnerCI, line 62: Are these necessary if they are empty blocks?

pkg/runner/jazzerjs/jazzerjs_runner.go
```diff
@@ -0,0 +1,97 @@
...
+
+	// ---------------------------
+	// --- fuzz target arguments -
+	// ---------------------------
+	args = append(args, "--")
+	args = append(args, options.JazzerJSTestPathPatternFlag(r.TestPathPattern))
+	args = append(args, options.JazzerJSTestNamePatternFlag(r.TestNamePattern))
+
+	// The environment we run the fuzzer in
```
mlsandnerCI, line 71: Not really a useful comment since the function name already says that.

pkg/runner/jazzerjs/jazzerjs_runner.go
```diff
@@ -0,0 +1,97 @@
...
+	env, err := r.FuzzerEnvironment()
```
mlsandnerCI, line 72: Not really a useful comment since the function name already says that.

internal/cmd/run/run.go
```diff
@@ -595,8 +608,12 @@ func (c *runCmd) buildFuzzTest() (*build.Result, error) {
...
 
 func (c *runCmd) runFuzzTest(buildResult *build.Result) error {
+	var err error
+
 	if c.opts.targetMethod != "" {
 		log.Infof("Running %s", pterm.Style{pterm.Reset, pterm.FgLightBlue}.Sprintf(c.opts.fuzzTest+"::"+c.opts.targetMethod))
+	} else if c.opts.testNamePattern != "" {
+		log.Infof("Running %s", pterm.Style{pterm.Reset, pterm.FgLightBlue}.Sprintf(c.opts.fuzzTest+":"+c.opts.testNamePattern))
 	} else {
 		log.Infof("Running %s", pterm.Style{pterm.Reset, pterm.FgLightBlue}.Sprintf(c.opts.fuzzTest))
```
mlsandnerCI, line 622: Maybe we could put the style into a separate variable since we always want to use the same style for this info and in case of changing it at some point, we would only have to change it in one spot.

integration-tests/node-typescript/nodets_test.go
```diff
@@ -0,0 +1,180 @@
...
+		if strings.HasPrefix(instruction, "import type") {
+			return strings.Join(instructions[i:len(instructions)-1], "\n")
+		}
+	}
+
+	return ""
+}
+
+// Modify the fuzz test stub created by `cifuzz create` to actually call a function.
+func modifyFuzzTestToCallFunction(t *testing.T, fuzzTestPath string) {
```
mlsandnerCI, line 143:
```suggestion
// modifyFuzzTestToCallFunction modifies the fuzz test stub created by `cifuzz create` to actually call a function.
func modifyFuzzTestToCallFunction(t *testing.T, fuzzTestPath string) {
```

mlsandnerCI: Same in the JS test too

internal/cmd/run/run.go
```diff
@@ -559,6 +574,8 @@ func (c *runCmd) buildFuzzTest() (*build.Result, error) {
 			return nil, err
 		}
 		return buildResult, err
+	case config.BuildSystemNodeJS:
+		return &build.Result{}, nil
```
mlsandnerCI, line 578: Why do we do this here?

mlsandnerCI: If this is correct, then a comment would be helpful because this behavior deviates so much from the other build systems.

pkg/parser/libfuzzer/libfuzzer.go
```diff
@@ -545,6 +589,7 @@ func (p *parser) finalizeAndSendPendingFinding(ctx context.Context) error {
 	parserOpts := &stacktrace.ParserOptions{
 		ProjectDir:    p.ProjectDir,
 		SupportJazzer: p.SupportJazzer,
+		SupportNode:   p.SupportJazzerJS,
```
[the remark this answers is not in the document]
MarkusZoppelt: Good catch

pkg/runner/jazzerjs/jazzerjs_runner.go
MarkusZoppelt: Removed

integration-tests/nodejs/nodejs_test.go
MarkusZoppelt: Ah, yes. The above check ensures that this test only runs on our 
`tests and Linting / test with CIFUZZ_PRERELEASE (ubuntu-latest)`
pipeline (that is dedicated to run PRERELEASE checks. We can debate if we want this to run on all pipelines. This way it is in line with our current strategy.

internal/cmd/run/run.go
MarkusZoppelt: I'd rather keep it like this.

I tried finding a good place and name for a constant but I found it too unreadable and almost as long/verbose compared to keeping it inline, because I like seeing which color is used directly where it occurs.

But I a open to clever suggestions 😉

internal/cmd/run/run.go
MarkusZoppelt: That's correct because Node.js does not require a build step.
Added a comment for explaination.

internal/cmd/run/run.go
MarkusZoppelt: Changed it before the if :)

mlsandnerCI (approved):

PR #884, opened 2023-06-30
fix(dangerfile.js): add checks for possibly undefined values

florianGla:
We should add checks for possibly undefined values. Otherwise the danger bot can crash.


jochil:
the optional chaining operator is supported by nodejs >= 12... but the dangerjs action was using v10, so I ditched the actions in favor of nodejs 18 lts and doing `yarn install` && `yarn run danger ci` manually

jochil (approved):

dangerfile.js
```diff
@@ -2,7 +2,7 @@ const { danger, warn, message, markdown } = require("danger");
 const { basename, dirname } = require("path");
 
 // Check for description
-if (danger.github.pr.body.length <= 0) {
+if (danger.github.pr.body?.length <= 0) {
```
jochil, line 5: I think this will behave as expected as `?` evaluate to `undefined` which will miss the `<=` comparison
```
if (undefined <= 0) {
  console.log("unreachable")
}
```
And I guess we want to trigger the warning when `body` is undefined

I will change it

PR #887, opened 2023-07-04
feat: pr bot improvements

mlsandnerCI:
### Motivation/Context
PR bot seemed slow when creating comments, visible in race conditions when pushing very fast in succession. Also got feedback from the team that it would be nice to acknowledge/skip files that are suggested for missing tests if they are not necessary.

### Description
I added extra checks in the `dangerfile.js` to (hopefully) improve the performance a bit and file lists are now displayed with checkboxes to tick them off.

Bildschirmfoto 2023-07-04 um 15 54 07

### How to use/reproduce
PR bot is automatic, no interaction needed.
You can test the bot locally on your PRs by following [this guide](https://danger.systems/js/guides/the_dangerfile.html). Be aware that if you want to test it with the "fake CI" process, you need to set the `DANGER_GITHUB_API_TOKEN` on your system.


dangerfile.js
```diff
@@ -17,81 +26,82 @@ if (danger.github.pr.changed_files > 15) {
 	);
 }
 
-// Checking for missing tests in created files
 const createdGoFiles = danger.git.created_files.filter(
 	(fileName) => fileName.includes(".go") && !fileName.includes("_test.go")
 );
 const createdTestFiles = danger.git.created_files.filter((fileName) =>
 	fileName.includes("_test.go")
 );
```
jochil, line 35: I know this is from the initial PR but I just noticed that the style of the arrow functions is different... one time the parameter is in he previous line, another time in the same with the statement. 

Also the `()` is not necessary when using only one parameter. 

A consistent way to write it can be

```
const createdGoFiles = danger.git.created_files.filter(
	fileName => fileName.includes(".go") && !fileName.includes("_test.go")
);
const createdTestFiles = danger.git.created_files.filter(
  fileName => fileName.includes("_test.go")
);
```

Maybe we can adjust prettier to handle this better

dangerfile.js
```diff
@@ -17,81 +26,82 @@ if (danger.github.pr.changed_files > 15) {
...
-const missingTestsForCreatedGoFiles = createdGoFiles.filter(function (x) {
-	// Create the test file names for the go file and check
-	// if it can be found in the list of created test files
-	const filePath = dirname(x);
-	const testFile = basename(x).replace(".go", "_test.go");
-	return !createdTestFiles.includes(`${filePath}/${testFile}`);
-});
+// Checking for missing tests in created files
+if (createdGoFiles?.length > 0) {
+	const missingTestsForCreatedGoFiles = createdGoFiles.filter(function (x) {
```
jochil, line 38: this can be
```
const missingTestsForCreatedGoFiles = createdGoFiles.filter(x=> {
```

dangerfile.js
```diff
@@ -17,81 +26,82 @@ if (danger.github.pr.changed_files > 15) {
...
   The following created files don't have corresponding test files:
   - ${missingTestsForCreatedGoFiles.join("\n - ")}`;
-	warn(message);
+		warn(message);
+	}
+}
+
+// Say thanks for adding new tests for already existing files
+if (createdTestFiles?.length > 0) {
+	const expandedTestCoverage = createdTestFiles.filter(function (x) {
```
jochil, line 61: same here:
```
const expandedTestCoverage = createdTestFiles.filter( x => {
```

dangerfile.js
```diff
@@ -17,81 +26,82 @@ if (danger.github.pr.changed_files > 15) {
...
+		markdown(
+			"You're a rockstar for creating tests for files that didn't have any yet ⭐"
+		);
+	}
 }
 
 // Checking for missing tests in modified files
+// Only check this if files have been modified to make the bot faster
 const modifiedGoFiles = danger.git.modified_files.filter(
 	(fileName) => fileName.includes(".go") && !fileName.includes("_test.go")
```
jochil, line 79: also the `()` are not needed here.
I think this check is used somewhere else already... to avoid duplication you can assign functions in javascript to variables and reuse them. For example
```
const testFileFilter = filename => fileName.includes(".go") && !fileName.includes("_test.go")

const createdGoFilers = danger.git.created_files.filter(testFileFilter)
...
const modifiedGoFiles = danger.git.modified_files.filter(testFileFilter)

```

dangerfile.js
```diff
@@ -17,81 +26,82 @@ if (danger.github.pr.changed_files > 15) {
...
-		!createdTestFiles.includes(`${filePath}/${testFile}`)
+if (modifiedGoFiles?.length > 0) {
+	const modifiedTestFiles = danger.git.modified_files.filter((fileName) =>
+		fileName.includes("_test.go")
 	);
-});
 
-if (missingTestsForModifiedGoFiles?.length > 0) {
-	const message = `
+	const missingTestsForModifiedGoFiles = modifiedGoFiles.filter(function (x) {
```
jochil, line 87: same with the error function

jochil (approved):
just some notes about javascript style

dangerfile.js
mlsandnerCI: Our `make fmt` actually includes the parentheses. Where could I adjust this?

mlsandnerCI: Adjusted the prettier format in a separate commit.

PR #899, opened 2023-07-11
fix(run): Fix campaign run details used for findings upload

fmeum:
The details of the campaign run created to represent a fuzz test when uploading findings are now populated correctly. As a result, the backend can show the fuzz test identifier as the display name rather than relying on heuristics to derive one.

See the individual commit messages for more details.


fmeum:
`go test ./e2e-tests/finding/` locally fails for me on `main` with errors similar to those seen in the E2E test. I think I need help to figure out what is going on with that.

This PR does not yet include tests: The integration test for the API communication doesn't seem to assert on any property of the contents of the request yet and I didn't want to build this in without having a discussion about this first. Do you have any recommendations on how to add test coverage?

fmeum:
CC @MarkusZoppelt as you are working on the related CLI-1091. Based on my testing against `https://app.code-intelligence.com` this doesn't fix the issue yet, but at least it should now be (more) clear that the problem lies in the backend.

MarkusZoppelt (approved):

PR #903, opened 2023-07-13
Add coverage support for Node.js projects

florianGla:
This PR adds coverage support for Node.js projects.

### Motivation/Context

Coverage shall also be computed for Node.js projects

### Description

We rely on the coverage computations by the jest test runner.

### How to use/reproduce

`cifuzz coverage FuzzTestCase`


integration-tests/nodejs/nodejs_test.go
```diff
@@ -176,3 +167,116 @@ func modifyFuzzTestToCallFunction(t *testing.T, fuzzTestPath string) {
...
+	err := os.WriteFile(filepath.Join(cifuzzRunner.DefaultWorkDir, "cifuzz.yaml"), []byte(configFileContent), 0644)
+	require.NoError(t, err)
+	expectedOutputExp = regexp.MustCompile(`"finding": {`)
+	cifuzzRunner.Run(t, &shared.RunOptions{
+		ExpectedOutputs: []*regexp.Regexp{expectedOutputExp},
+	})
+
+	// Check that command-line flags take precedence over config file
+	// settings (only on Linux because we only support Minijail on
+	// Linux).
```
MarkusZoppelt, line 208: same here. What has this to do with Minijail?

internal/cmd/coverage/node/node.go
```diff
@@ -0,0 +1,143 @@
...
+	// check if response is empty
+	if strings.TrimSpace(string(output)) == "" {
+		log.Error(errors.New("fuzz test not found"))
+		return cmdutils.ErrSilent
+	}
+
+	return nil
+}
+
+func (cov *CoverageGenerator) runNpxCommand(args []string, stdout, stderr io.Writer) error {
```
MarkusZoppelt, line 111:
```suggestion
func (cov *CoverageGenerator) runNPXCommand(args []string, stdout, stderr io.Writer) error {
```

MarkusZoppelt (commented):
I think we should also add a test to `internal/cmd/coverage/coverage_test.go` where we check for `TestNodeMissing`, etc.

MarkusZoppelt (approved):
LGTM

PR #904, opened 2023-07-13
test/ci: improve coverage reporting + pipeline

jochil:
### Motivation/Context
We want to get a better overview on our test coverage to better detect blind spots in our test suite.

### Description
Using the coverage instrumentation introduced in go v1.20 to gather coverage data for our integration/e2e tests

Currently the pipeline is only executed on a linux runner. If needed we can run it in a matrix and merge the results from different operating systems

### How to use/reproduce
You can run `make coverage` to execute all test suites and receive a overall report

To improve performance a bit the pipeline executes all suites in parallel and then merges the results. You can do that locally by
```
make coverage/unit
make coverage/integration
make coverage/e2e
make coverage/merge
```


internal/testutil/testutil.go
```diff
@@ -107,3 +108,23 @@ func MkdirTemp(t *testing.T, dir, pattern string) string {
...
+	_, b, _, _ := runtime.Caller(0)
+	// Note: The number of levels we go up here has to be adjusted if
+	// this source file is moved.
+	basepath := filepath.Dir(filepath.Dir(filepath.Dir(b)))
+	return basepath
+}
+
+// SetupCoverage creates a directory for coverage data and sets the
+// needed environment variable
+func SetupCoverage(t *testing.T, env []string, subdir string) []string {
```
MarkusZoppelt, line 123: Does it make sense to somehow include `t.Helper()` at the beginning of the function since `SetupCoverage` is a helper function for tests?

MarkusZoppelt (approved):

florianGla (approved):

PR #909, opened 2023-07-14
feat(coverage): support maven coverage runs with specific target methods

florianGla:
This PR adds support for maven coverage runs with specific target methods.

### Motivation/Context

We supported only fuzzing runs for maven with specific target methods. This feature was not supported for coverage runs.

### Description

Maven surfire allows to specify target methods with this syntax `-Dtest=MyTest#myMethod`
### How to use/reproduce

Execute `cifuzz coverage com.example.FuzzTestCase::myFuzzTest`


florianGla:
> Maybe we should add a check if the target method exists like we do in [jazzer_bundler.go](https://github.com/CodeIntelligenceTesting/cifuzz/blob/1be4032ca4c282645829c5ff27d24374cf6b8495/internal/bundler/jazzer_bundler.go#L330)? When calling coverage with a wrong target method (and fuzz test in that case) the error message from maven is not so clear, it doesn't mention like CMake that the test doesn't exist, just that it `failed to execute goal`.
> 
> And maybe we should have a test that this works?

Great idea! You are right. We should also add tests for target methods. I would maybe do that in a separate PR for maven, gradle and node for run and coverage.

florianGla:
I think we should check if the fuzz tests exist in a separate PR. I added a unit test .

PR #911, opened 2023-07-17
refactor: adding more debug output to bundle for 'other'

mlsandnerCI:
### Motivation/Context

Users using build system `other` often have complex setups and need more debug information when investigating an error. This should reduce the amount of time CS and the CLI team have to invest in debugging a problem.

This PR is only adding debug information, adjusting the error messages will be in a separate PR.

### Description

More debug information about
- what environment variables have been set
- what commands actually do
- what flags have been set (e.g. docker-image, output-path)

Additionally:
- Found an ignored error in bundle and included a check
- Added a check if the provided output path is missing the extension and adds it in that case
- Added `.dSYM` folders to `.gitignore`

### How to use/reproduce

Running `cifuzz bundle my_fuzz_test -v` in example/other shows the newly added debug informations.


internal/bundler/libfuzzer_bundler.go
```diff
@@ -511,6 +516,7 @@ depsLoop:
 	// Add dictionary to archive
 	var archiveDict string
 	if b.opts.Dictionary != "" {
+		log.Debug("Adding dictionary")
```
MarkusZoppelt, line 519: Does it make sense to print the value of `b.opts.Dictionary` as well?

```suggestion
        log.Debugf("Adding dictionary ", b.opts.Dictionary)
```

util/envutil/envutil.go
```diff
@@ -52,6 +53,8 @@ func Getenv(env []string, key string) string {
 // Like os.Setenv but uses the specified environment instead of the
 // current process environment.
 func Setenv(env []string, key, value string) ([]string, error) {
+	log.Debugf("Setting %s = %s", key, value)
```
MarkusZoppelt, line 56:
```suggestion
	log.Debugf("Setting ENV var: %s = %s", key, value)
```

.gitignore
```diff
@@ -22,3 +23,5 @@ coverage.out
 gotest.log
 
 /node_modules/
+
+
```
MarkusZoppelt, line 27: formatting

MarkusZoppelt (approved):

florianGla:
Maybe it is useful to print a debug output, that indicates if the current build step is a fuzzing or coverage build and prints the location of the executables

jochil:
Please be careful about printing environment variables. We don't want to leak any sensitive information by printing out some tokens/passwords/...

jochil:
> > Please be careful about printing environment variables. We don't want to leak any sensitive information by printing out some tokens/passwords/...
> 
> So not printing a debug message every time we set an environment variable but only do that explicitly?

The risk I am see here is that someone using `envutil.Setenv` might not know that they get printed... so it is easy to leak a access token into something like pipeline logs. I would prefer doing it explicitly for the environment variables that are save to print and that have valuable information like compiler flags

PR #913, opened 2023-07-17
feat(container): implement remote-run

MarkusZoppelt:
### Motivation/Context

We want users to use containers for remote run. This is the first step and our backend needs to implement support for this new behavior.
This is our first step enabling only bundling and upload of a "bundle"/image to a container registry.

### Description

Builds and uploads container bundles to a container registry.

### How to use/reproduce

Login with a OCI engine (e.g. Docker) to a container registry, e.g.:

- `docker login ghcr.io/codeintelligencetesting/cifuzz`
- `make clean && make build-container-image` (to build `dev` manifest)
- `CIFUZZ_PRERELEASE=1 cifuzz container remote-run -C examples/gradle --registry ghcr.io/CodeIntelligenceTesting/cifuzz com.example.FuzzTestCase -v`

(please remove uploaded artifacts after testing from [our github packages](https://github.com/CodeIntelligenceTesting/cifuzz/pkgs/container/cifuzz/versions))


JackuB:
Screenshot 2023-07-20 at 10 30 48
Without -verbose flag it won't print final image sha - is that intentional behavior?

jochil:
> Screenshot 2023-07-20 at 10 30 48 Without -verbose flag it won't print final image sha - is that intentional behavior?

yes. If everything goes well the user should not have to care about the image at all... at least that was our thought

internal/container/image.go
```diff
@@ -37,14 +33,52 @@ type dockerfileConfig struct {
 }
 
 // BuildImageFromBundle creates an image based on an existing bundle
```
mlsandnerCI, line 35:
```suggestion
// BuildImageFromBundle creates an image based on an existing bundle.
```

internal/cmd/container/remoterun/remoterun.go
```diff
@@ -1,24 +1,125 @@
...
+				log.Errorf(err, "Failed to parse cifuzz.yaml: %v", err.Error())
+				return cmdutils.WrapSilentError(err)
+			}
+
+			// check for registry flag
+			if opts.Registry == "" {
+				err = errors.New("no registry specified")
+				log.Error(err)
+				return cmdutils.WrapSilentError(err)
+			}
```
JackuB, line 61: Needs to follow the error code guidelines

internal/cmd/container/remoterun/remoterun.go
```diff
@@ -1,24 +1,125 @@
...
+
+			err := config.FindAndParseProjectConfig(opts)
+			if err != nil {
+				log.Errorf(err, "Failed to parse cifuzz.yaml: %v", err.Error())
+				return cmdutils.WrapSilentError(err)
+			}
+
+			// check for registry flag
+			if opts.Registry == "" {
```
JackuB, line 57: Isn't there a Viper-way for marking flag as required?

internal/cmd/container/remoterun/remoterun.go
```diff
@@ -1,24 +1,125 @@
...
+				err = errors.New("no registry specified")
+				log.Error(err)
+				return cmdutils.WrapSilentError(err)
+			}
+
+			fuzzTests, err := resolve.FuzzTestArguments(opts.ResolveSourceFilePath, args, opts.BuildSystem, opts.ProjectDir)
+			if err != nil {
+				log.Print(err.Error())
+				return cmdutils.WrapSilentError(err)
+			}
```
JackuB, line 67: Don't print error, just return it

internal/config/cifuzz.yaml.tmpl
```diff
@@ -45,5 +45,8 @@
 ## Set the project name on the CI App.
 {{if .Project}}project: {{.Project}}{{else}}#project: my-project-1a2b3c4d{{end}}
 
+## Set the container registry to upload fuzz containers to.
+# registry: ghcr.io/my-org/my-project
```
JackuB, line 49: Use a different example value:

```suggestion
# registry: registry.example.org/my-org/my-project
```

internal/container/image.go
```diff
@@ -37,14 +33,52 @@ type dockerfileConfig struct {
...
 
+// UploadImage uploads an image to a registry
+func UploadImage(imageID string, registry string) error {
+	log.Debugf("Start uploading image %s to %s", imageID, registry)
+
+	dockerClient, err := getDockerClient()
+	if err != nil {
+		return err
+	}
+	ctx := context.Background()
```
JackuB, line 52: Add a TODO/Improvement:

- make the building/pushing cancellable with SIGnals

internal/container/parse.go
```diff
@@ -0,0 +1,71 @@
...
+		if jsonMessage.Aux != nil {
+			var res types.BuildResult
+			if err := json.Unmarshal(*jsonMessage.Aux, &res); err != nil {
+				log.Errorf(err, "Failed to unmarshal Aux JSON message: %s", *jsonMessage.Aux)
+			} else {
+				if strings.HasPrefix(res.ID, "sha256:") {
+					id = res.ID[7:]
+				} else {
+					id = res.ID
+				}
```
JackuB, line 66: Do we need the else block? It's an unexpected ID format?

internal/container/parse.go
```diff
@@ -0,0 +1,71 @@
...
+			}
+		}
+	}
+	return id, nil
```
JackuB, line 70: Is empty ID an error state? Should it be checked?

JackuB (approved):
Needs E2E test (even in a follow up PR), since unit testing Docker SDK calls is too painful

PR #914, opened 2023-07-17
fix(maven): trim maven xml output

florianGla:
This fixes an issue, where the xml output was polluted by leading and trailing bytes.

### Motivation/Context

This issue occurs on Ubuntu  23.04 with Maven 3.8.7.

### Description

Only the xml output between `<project></project>` is parsed.

### How to use/reproduce

Start a maven fuzzing run on  Ubuntu  23.04 with Maven 3.8.7


internal/build/maven/xml.go
```diff
@@ -49,13 +50,19 @@ type Project struct {
...
 
+	// only parse output contained in <project></project> tags
+	startTag := []byte("<project")
+	endTag := []byte("</project>")
+	startIdx := bytes.Index(output, startTag)
+	endIdx := bytes.Index(output, endTag) + len(endTag)
+
 	var project Project
-	err = xml.Unmarshal(bytes, &project)
+	err = xml.Unmarshal(output[startIdx:endIdx], &project)
```
jochil, line 65: This asks for a unit test making sure it works as expected. Also I think a comment explaining the reasoning for doing it and/or linking to the bug ticket would be helpful

florianGla: Ok, we already have a unit test that assumes that the output is stripped. I can pollute it a little bit :). I will write a longer comment.

mlsandnerCI (approved):

PR #919, opened 2023-07-18
feat: change license

MarkusZoppelt:
### Motivation/Context

CI Fuzz is changing its license to CC BY-NC-ND 4.0.

### Description

- Switched license and added a short preface.
- Removed all server connection dialogs that asked users to authenticate before `run` & `finding`
- Removed different messaging options for the connection dialogs (cleanup)
- `finding` cmd shows all the available information regardless of authentication status (`error-details` still require a valid authentication)

### How to use/reproduce

* try running the `run`, `remote-run` or `finding` cmd without authentication to check for usage warning.


internal/cmd/container/run/run.go
```diff
@@ -110,7 +113,13 @@ container is built and run locally instead of being pushed to a CI Sense server.
 }
 
 func (c *containerRunCmd) run() error {
-	var err error
+	authenticated, err := auth.GetAuthStatus(c.opts.Server)
+	if err != nil {
+		return err
+	}
+	if !authenticated {
```
jochil, line 120: It should be easy to test for the display of the usage warning for all affected commands

jochil: ok, thanks :+1:

jochil (approved):

PR #920, opened 2023-07-18
refactor: only show stacktrace in verbose mode

mlsandnerCI:
... and adjusted the error output to handle wrapping

### Motivation/Context

Feedback from users to CS showed that always showing the stack trace when an "unexpected" error occurs confused them and signaled that there was a problem with our CLI even if the error was a user error. 

### Description

Error messages in root have been adjusted to only show stack traces in verbose mode. 
Wrapped errors have been taken into consideration so that only the stack trace of the root error will be printed but with all accompanying messages. If an error has no stack trace (either because we didn't add it or because the error doesn't implement the interface), it will just be printed as is. 
In case of an error in non-verbose mode a support message points to verbose mode for more information.
The output now also follows our usual format, prefixing errors with "❌" except in plain mode.


jochil (approved):

adombeck:
> Feedback from users to CS showed that always showing the stack trace when an "unexpected" error occurs confused them and signaled that there was a problem with our CLI even if the error was a user error.

User errors should always be expected, so if a stack trace is shown for a user error, we should change the code to catch that error and show a helpful error message without a stack trace.

adombeck:
I'm worried that this change will lead us to receive less helpful bug reports because the stack trace is omitted (unless the user re-runs the command with `--verbose`).

adombeck:
Also, bugs which can't easily be reproduced will be harder to debug because we don't print the stack trace.

jochil:
Giving more precise and helpful error messages is definitely something we should do, but this does not necessarily means we have to print a stack trace. One can argue that a stack trace is a verbose form of displaying an error. So it is fine moving it behind the `--verbose` flag

We don't have a problem with bug reports missing the `--verbose` output... and also we have this guideline: https://code-intelligence.atlassian.net/wiki/spaces/CI/pages/325877781/How+to+quickly+solve+customer+issues+Bug+guideline as well

adombeck:
> Giving more precise and helpful error messages is definitely something we should do, but this does not necessarily means we have to print a stack trace.

That's what I'm saying. Whenever we know why an error occurs, we print a helpful error message *without* a stack trace. But for unexpected errors (for which we can't show a helpful error message) a stack trace is useful.

PR #923, opened 2023-07-19
fix(cmdutils): use fully qualified name for java fuzz test classes containing one fuzz test

florianGla:
We want to use the fully qualified name for java fuzz test classes containing only one fuzz test. This is necessary in order to generate consistent names in bundles.

### Description

We used to handle this case in a different way. This PR just removes this extra case.

### How to use/reproduce

Run `cifuzz bundle` in our example maven or gradle projects.


jochil (approved):
great PR... more tests and removed code :heart_eyes:

PR #927, opened 2023-07-20
Fixes and improvements of the `container run` command

adombeck:
### Description

Multiple minor improvements and fixes. See the commit messages and comments in the code for details.


internal/container/container.go
```diff
@@ -30,6 +31,11 @@ func Create(fuzzTest string) (string, error) {
 		containerConfig.Cmd = append(containerConfig.Cmd, "-v")
 	}
 
+	if os.Getenv("CIFUZZ_BREAK_CONTAINER") == "1" {
```
jochil, line 34: What is the thought behind this? Couldn't we always run the container with `-it <tag> /bin/bash` to inspect it? Or just start it once by hand with `cifuzz && sleep` and then attach to it? Why do we need code to handle this?

btw. this might be something that could have been explained in the PR description as well

internal/cmd/container/run/run.go
```diff
@@ -161,8 +161,8 @@ func (c *containerRunCmd) run() error {
 
 	// TODO: make output pretty
 	//  Remove 'cifuzz version' from output
-	fmt.Println(containerStdOut.String())
-	fmt.Println(containerStdErr.String())
+	_, _ = fmt.Fprintln(os.Stdout, containerStdOut.String())
+	_, _ = fmt.Fprintln(os.Stderr, containerStdErr.String())
```
jochil: It might me temporary code but wouldn't `cmd.OutOrStdOut()` && `cmd.ErrOrStdErr()` more cobra like

internal/container/container.go
adombeck: > What is the thought behind this? 

It allows us to get a shell in the container to debug things. In my experience that's a huge benefit over the "patch something -> rebuild -> rerun the command" cycle.

> Couldn't we always run the container with -it <tag> /bin/bash to inspect it?

I don't think so, because you can only attach to running containers.

>  Or just start it once by hand with cifuzz && sleep and then attach to it?

How do you do that?

>  Why do we need code to handle this?

Because there is no other convenient way as far as I know.

adombeck: > btw. this might be something that could have been explained in the PR description as well

I'll add a comment in the code if we agree on the feature.

internal/cmd/container/run/run.go
adombeck: We're using `os.Stdout` and `os.Stderr` in many places in our code by now - IMO for good reasons, because passing on `cmd.OutOrStdOut()` and `cmd.ErrOrStdErr()` to all functions which print something to stdout/stderr is overhead and clutters the code and it's not useful IMO, unless we actually use cobra's `Out` and `Err` streams which I think we don't.

internal/container/container.go
jochil: I got the intention, but you can just run `docker run -it cifuzz sleep infinity`  and then `docker exec ...` to further inspect the container?
What is reasoning behind doing it via cifuzz?

adombeck: > I got the intention, but you can just run docker run -it cifuzz sleep infinity

That's a different container, it does not use the config we use in cifuzz to create containers (https://github.com/CodeIntelligenceTesting/cifuzz/blob/50282a40a490c040ccdfbbdbf0da24cca660df7e/internal/container/container.go#L24-L50). Admittedly, so far we don't configure much there, so the container started via `docker run cifuzz` is similar to the one that's started by cifuzz. But one thing that's different is that `CIFUZZ_BREAK_CONTAINER` allows you to easily run the original command via `eval $CMD`, otherwise you have to figure out that command via `docker inspect $CONTAINER_STARTED_VIA_CIFUZZ` first.

We had this feature in ci-daemon and I always found it very useful. It was more important then because we configured a lot of things in the container (e.g. bind mounts) which were tedious to recreate manually. I'm still in favor of supporting the feature because:
* It makes it easy to re-run the original command in the container via `eval $CMD`
* It allows other devs to discover the feature
* It's a very small feature (3 lines + usage message)

internal/cmd/container/run/run.go
jochil: In the GitHub CLI they are also using an abstraction for the `Std*` stuff... see: https://github.com/cli/cli/blob/ad283cff72863959d71f52d295f3969006bb5e73/pkg/cmd/auth/status/status.go#L68
and https://github.com/cli/cli/blob/8079d18efd2e195456280d9b9742c13d33702e6a/pkg/iostreams/iostreams.go

adombeck: Yes, they are using an opts struct which they pass around to each function. We don't do that currently and I don't see how the benefit of doing that outweighs the cost for us (refactoring a lot of our code + the added complexity). 

Which part of our code could we simplify by using `cmd.OutOrStdOut()` and `cmd.ErrOrStdErr()`?

jochil: > Yes, they are using an opts struct which they pass around to each function. We don't do that currently and I don't see how the benefit of doing that outweighs the cost for us (refactoring a lot of our code + the added complexity).
> 
> Which part of our code could we simplify by using `cmd.OutOrStdOut()` and `cmd.ErrOrStdErr()`?

Like I said, it would improve the testability of our application and simplify the test code... for example: https://github.com/CodeIntelligenceTesting/cifuzz/blob/main/internal/cmdutils/tests.go#L14
Which was already worked on here: https://github.com/CodeIntelligenceTesting/cifuzz/pull/921
The github CLI is just an example on how to do it. 

IMO using `cmd.OutOrStdOut()` is a move in the right direction... also there is no disadvantage over using it compared to `os.Stdout`. Or asked in a different way: What do we gain when using `os.Stdout`?

internal/container/container.go
adombeck: In ci-daemon we used `*_SLEEP_FOREVER`. I always found that a bit clumsy so I opted for `*_BREAK_CONTAINER` here, which was inspired by the `break` kernel command-line option which spawns a shell in the initramfs boot stage (search for "break" in [the initramfs man page](https://manpages.ubuntu.com/manpages/xenial/man8/initramfs-tools.8.html)). 

To me `*_KEEP_ALIVE` sounds like it would run the container as usual but keep it running after it finished what it should do. So from the current options, my preference is `*_BREAK_CONTAINER` > `*_SLEEP_FOREVER` > `*_KEEP_ALIVE`. That being said, it's not important enough to me to continue arguing over it, so please tell me your preference and I'll change it.

internal/cmd/container/run/run.go
adombeck: > also there is no disadvantage over using it compared to os.Stdout. Or asked in a different way: What do we gain when using os.Stdout?

We gain simplicity, by not having to drag two parameters through all the functions called from the `cobra.Command.Run` up until the function which prints something to stdout or stderr. In this case, it turned out to be a short chain, so I did as you requested. In general, I still don't think we should add those parameters to all our functions.

internal/container/container.go
adombeck: > I'll add a comment in the code if we agree on the feature.

done

jochil: ok, I would like `CIFUZZ_CONTAINER_SLEEP` > `CIFUZZ_CONTAINER_KEEP_ALIVE` > `CIFUZZ_BREAK_CONTAINER` > `CIFUZZ_CONTAINER_SLEEP_FOREVER`

jochil (approved):

internal/container/container.go
adombeck: done

adombeck:
While fixing the failing test, I found it hard to parse the GitHub action logs, so I changed some things to improve them. Please take a look at the last two commits.

e2e-tests/e2e-test-runner-assertions.go
```diff
@@ -1,38 +1,49 @@
...
 func (co *CommandOutput) Failed() *CommandOutput {
-	assert.NotEqualValues(co.t, 0, co.ExitCode)
+	require.NotEqualValues(co.t, 0, co.ExitCode)
 	return co
 }
 
 func (co *CommandOutput) OutputContains(expected string) *CommandOutput {
-	assert.Contains(co.t, co.Stdout, expected)
+	if !strings.Contains(co.Stdout, expected) {
+		require.FailNow(co.t, fmt.Sprintf("stdout does not contain %q", expected))
```
jochil, line 22: why not using `require.Contains`?

e2e-tests/e2e-test-runner-assertions.go
```diff
@@ -1,38 +1,49 @@
 package e2e
 
-import "github.com/stretchr/testify/assert"
+import (
+	"fmt"
+	"strings"
+
+	"github.com/stretchr/testify/require"
```
jochil, line 7: I am not sure if exiting the test on the first wrong assertion is a desired thing here...especially because of the function chaining.
Imagine having:
```
co.Success().OutputContains("foo")
```
If `Success` fails I would still like to see if the output was correct

adombeck: In my experience it's more useful to have the error which caused the test to fail more discoverable (than having it at the bottom of the log) than to have the output of multiple different failing tests. If `Success` fails, then the command that failed should have printed an error message which should be good enough to debug the issue, I don't want to see a different, distracting error about the output not being correct because that's most probably a side effect of the command having failed.

e2e-tests/e2e-test-runner-assertions.go
jochil: Ok, I see... Is the output always printed or only with `go test -v` ? If the first is the case I am fine

e2e-tests/e2e-test-runner-assertions.go
jochil: Ok, makes sense. I use `assert` and `require` very intentional... so maybe @JackuB did the same here.

e2e-tests/e2e-test-runner-assertions.go
adombeck: It's always printed

e2e-tests/e2e-test-runner-assertions.go
JackuB: > I don't want to see a different, distracting error about the output not being correct because that's most probably a side effect of the command having failed.

👍 failing test quickly and avoiding figuring out what assertion failed and what is a red herring is 🤐 

When debugging the test locally you can always add more debugger points or print statements as needed.

PR #931, opened 2023-07-20
refactor: remove unneccesary usage of SilentErrors

mlsandnerCI:
### Description

- Refactored occurrences where we used to log and then silence an error to only hide the stack trace in non-verbose mode. Now they just return the error and let the root command handle it
- Adjusted the order of error checks in root because otherwise specific errors like `IncorrectUsageError` (where we never want or need the stacktrace) would not be recognized because we always would run into the case "not a SilentError" first
- Whenever it was necessary to write an error message multiple times to give context e.g. "Failed to parse cifuzz.yaml" it got moved into the respective function 

Notes:
I tried to not change any functionality of when we print a stack trace. We only print the stack trace of the root error so if that one doesn't have one attached, no stack trace will be printed in verbose mode. For example see this timeout parsing issue:
Bildschirmfoto 2023-07-20 um 15 27 11
Also whenever we used log.Error() before, we would print the stack trace in verbose mode so this should be handled the same now.
**If you find any occurrences where the initial idea of whether or not to print a stack trace got changed accidentally, please leave a comment!**


adombeck:
It's bothering me that this PR removes the "this is an expected error" bit of information from our errors. It makes sense since !920 treats expected and unexpected errors the same, which I think is a mistake.

If I understand correctly, !920 was implemented because users were confused that their (non-verbose) command output showed a stack trace, which led them to believe that cifuzz was broken while actually they were just using it wrong. IMO the right fix is to take a look at the error that caused this stack trace to be printed and handle it as an expected error, i.e. print a helpful error message for it and hide the stack trace by returning a silent error. Arguably, the fact that we printed a stack trace for an expected error does mean that cifuzz was broken, because it should have handled that error differently.

And for cases where cifuzz is actually broken, it is useful to print a stack trace in non-verbose mode to make it easy for the user to distinguish these cases from cases where they are just using it wrong (and to make it easier for us to debug it with that output).

adombeck:
> Shouldn't the error message itself give us enough information to distinguish if it is an expected or unexpected error? 

Currently it doesn't, because for unexpected errors we just print the error message that we got from the function that returned the error.

> Plus we still don't print stacktraces for expected errors, e.g. the ones we wrap as an IncorrectUsageError. 

IncorrectUsageErrors are just a small subset of expected errors, i.e. the ones which we can directly link to "you specified an invalid command-line argument" and we handle it by printing the usage message. Other expected errors are handled by printing a useful error message and returning a SilentError.

> I would propose using those instead of logging + wrapping as an SilentError in the code to not log in two different places just to hide a stack trace.

I don't understand what you mean with "to not log in two different places". We don't log expected errors in two different places, we only log them once and then return a SilentError which is not logged.

> The root command takes care of whether or not to print a stacktrace, we just have to return the error correctly and with an appropriate message.

My issue with that is that, with this PR, the root command doesn't know if the error is an expected error or not, so it handles expected and unexpected errors the same: It prints a stacktrace if and only if the `--verbose` flag was used. My point is that it's actually useful to handle expected and unexpected errors differently, because it makes it clear to the user if the tool broke because they were using it wrong or if it broke because there was an internal error which they should report to us.

> Could you give me some examples in this PR where you think we should use an error that doesn't show the stacktrace?

Almost all of the ones which were SilentErrors before. Some examples: 
* "No fuzz test(s) could be found in the project directory"
* "No valid targets found for patterns: %s"
* "Failed to parse cifuzz.yaml: %v"
 
> Maybe we also have different definitions of expected/unexpected error? 

I'm using "expected error" as "an error which we're aware of that it might happen (and which we therefore already handle in a special way, usually by printing a helpful error message)". Unexpected errors are all other errors (which we don't have special handling for).

adombeck:
FWIW the concept of handling errors in the place where we know how to handle them and then returning a `SilentError` comes from the (GitHub CLI tool)[https://github.com/cli/cli] (e.g. https://github.com/cli/cli/blob/ad283cff72863959d71f52d295f3969006bb5e73/pkg/cmd/auth/status/status.go#L76-L78), a well written command-line tool (in my opinion) which we took inspiration from when we first designed and implemented cifuzz.

adombeck:
My proposal is that we
* update our Go Best Practices page to explain that we handle errors where we know how it should be handled, and then return a `SilentError`
* revert #920
* close this PR

I would be grateful for suggestions on how to make it easier to understand why we're using this `SilentError` concept when looking at our code (maybe a better name for the error?). I myself have not come up with anything.

jochil:
Looking at the changed code in this PR, I would prefer the "wrap & let root handle it" approach compared to "log & return silent". Most of the time we were logging the same message anyways

I understand the reasoning behind the SilentError and I think there are cases where it makes sense to still use it, as @mlsandnerCI  outlined in the best practices page. 
But I am not sure if the expected vs unexpected error categorization is helpful for us. If an error is expected but we still have to stop the program execution, what is the difference from an unexpected error. At the end we have to stop and print a message (and this looks easier with this change). Maybe we can look at it as "terminating vs. handleable" error :thinking:

internal/cmd/root/root.go
```diff
@@ -214,6 +177,37 @@ For more information on cifuzz sandboxing, see:
...
+				// but only print the stacktrace of the root error cause
+				if errors.As(errors.Cause(err), &st) {
+					_, _ = fmt.Fprint(cmd.ErrOrStderr(), style.Sprintf("\n%s%v%+v\n", icon, err, st.StackTrace()))
+				} else {
+					// Catch cases where we either did not add any stacktrace/wrapped the error
+					// or the error does not implement the interface for the stacktracer e.g. os.ErrExist
+					_, _ = fmt.Fprint(cmd.ErrOrStderr(), style.Sprintf("\n%s%v\n", icon, err))
+				}
+			} else {
+				supportMsg := "More information can be acquired running the command in verbose mode (-v).\n"
```
JackuB, line 206: I'd have a preference for directing the user to use a spelled-out `--verbose` flag, rather than a shorthand `-v` in messages and docs.

Maybe updating the tone and expanding the message could be helpful, something like:

> Error: Something went wrong while processing the command. 
>
> $THE_ERROR_MESSAGE
>
> For more information about the issue, try running cifuzz in verbose mode using the '--verbose' option.
> 
> Example:
> cifuzz my_command --verbose (idk, can we print exactly what user ran, so they can just copypasta it? If not, then let's skip the example part!)
> 
> If you continue to encounter issues, please refer to the documentation https://docs.code-intelligence.com or run 'cifuzz --help' for additional options.

JackuB (approved):

internal/cmd/root/root.go
mlsandnerCI: I'm generally unsure if I want to keep this message because we do have some errors that don't have a stack trace (because we expect them) and it is weird to then run the command in verbose mode and not have any more information than before. Any ideas?

JackuB: Hmm, we could also detect if there is a stacktrace in error? Or change the message to something like "there may be more information when running the command in a verbose mode".

mlsandnerCI: Hm, I could probably to the same check like I do in l.198 to see if there is a stack trace 🤔 I will check if this works.

mlsandnerCI: That worked, thanks! I do not know why I didn't see this solution earlier 😅

PR #933, opened 2023-07-20
Init command: arg should override project type detection

JackuB:
It was reported that arguments aren't taking priority, leading to awkward interactions like cifuzz refusing to init in a Node.js folder without the `prerelease` switch:

```terminal
cifuzz on  main [?] via  v1.20.6 via  v18.16.0
❯ cifuzz init java
cifuzz version 1.0.0
cifuzz does not support Node.js projects yet.
```

### Description

CLI-1162

### How to use/reproduce

E2E test case shows the behavior


adombeck (changes requested):
I expect that the command still fails if the build system which was determined automatically is not supported on the current platform, because we still call `config.ValidateBuildSystem(opts.BuildSystem)` with the `opts.BuildSystem` which was determined automatically. I think we should not call `config.DetermineBuildSystem(opts.Dir)` at all if `opts.testLang` was provided.

adombeck:
I appreciate the test driven approach but I would prefer it if you would reverse the order of the commits (i.e. having the commit which adds the test come after the commit which fixes the issue), so that the tests pass for each separate commit which is useful for git bisecting.

JackuB:
> I think we should not call config.DetermineBuildSystem(opts.Dir) at all if opts.testLang was provided.

Wouldn't that complicate the decision process between the maven or gradle init, when user specified `java` as the `testLang`?
This mapping is awkward on its own already. Anyone else wants to chime in?

---

> the tests pass for each separate commit which is useful for git bisecting

Sure, I can swap the commits, but I don't see "tests passing for commits" as something we could enforce going forward 🤔   we do rely on way too many external factors, especially in integration tests.

internal/config/types.go
```diff
@@ -12,6 +12,27 @@ const (
...
+	case CPP:
+		return BuildSystemOther
+	case Java:
+		// Respect the detected build system if it is Gradle or Maven. Otherwise default to Maven.
+		if detectedBuildSystem != BuildSystemGradle && detectedBuildSystem != BuildSystemMaven {
+			return BuildSystemMaven
+		}
+		return detectedBuildSystem
+	case Kotlin:
+		return BuildSystemMaven
```
jochil, line 28: I think Kotlin can be used with gradle as well... or?

internal/config/types.go
```diff
@@ -12,6 +12,27 @@ const (
 	TypeScript FuzzTestType = "ts"
 )
 
+// GetBuildSystemFromTestType creates a mapping between test type and build system.
+// detectedBuildSystem is optional
+func GetBuildSystemFromTestType(testType FuzzTestType, detectedBuildSystem string) string {
+	switch testType {
+	case CPP:
+		return BuildSystemOther
```
jochil, line 20: IMO CPP can be used with `cmake` too

adombeck:
>> I think we should not call config.DetermineBuildSystem(opts.Dir) at all if opts.testLang was provided.

> I moved the validate check to an else branch if testLang is not specified.

Still not really happy with how we handle things in case that that `testLang` is specified. Let's assume a project which has a `node_modules` directory but the user wants to set it up for Java fuzzing with Gradle. They call `cifuzz init java` to avoid that cifuzz handles it as a JavaScript project. Currently, cifuzz still calls `config.DetermineBuildSystem` which returns `BuildSystemNodeJS`. Because the user specified the language, this decision is overwritten in `GetBuildSystemFromTestType` to `BuildSystemMaven` even if the directory contains Gradle files and no `pom.xml`.

When `testLang` is specified, I think we should only try to determine the build system if we support multiple build systems for the  specified language (currently only the case for Java) and then only check for those.

JackuB:
> Could we just let the user specify the build system directly instead of the language?

I think this is the only solution that will not make us crazy. I will update the PR

internal/cmd/init/init.go
```diff
@@ -57,23 +57,31 @@ func New() *cobra.Command {
...
+			if len(args) == 1 {
+				opts.testLang = config.FuzzTestType(args[0])
 			}
 
-			err = config.ValidateBuildSystem(opts.BuildSystem)
-			if err != nil {
-				return err
+			// Override detected build system if test language is specified.
+			if opts.testLang != "" {
+				// We validate arg is allowed, thus we can pick from the map
```
MarkusZoppelt, line 66: I don't understand this comment

internal/cmd/init/init.go
```diff
@@ -205,3 +213,12 @@ func getNodeProjectLang() (config.FuzzTestType, error) {
...
 	return config.FuzzTestType(userSelectedLang), nil
 }
+
+// map of supported test types/build systems for init command
+var supportedInitTestTypes = map[string]string{
+	"cmake":  string(config.BuildSystemCMake),
+	"maven":  string(config.BuildSystemMaven),
+	"gradle": string(config.BuildSystemGradle),
+	"js":     string(config.BuildSystemNodeJS),
+	"ts":     string(config.BuildSystemNodeJS),
```
adombeck, line 223: My IDE complains that this is a redundant type conversion

```suggestion
	"cmake":  config.BuildSystemCMake,
	"maven":  config.BuildSystemMaven,
	"gradle": config.BuildSystemGradle,
	"js":     config.BuildSystemNodeJS,
	"ts":     config.BuildSystemNodeJS,
```

internal/cmd/init/init.go
```diff
@@ -38,7 +38,7 @@ func New() *cobra.Command {
 	var bindFlags func()
 	opts := &options{}
 	cmd := &cobra.Command{
-		Use:   fmt.Sprintf("init [%s]", strings.Join(maps.Values(config.SupportedTestTypes()), "|")),
+		Use:   fmt.Sprintf("init [%s]", strings.Join(maps.Keys(supportedInitTestTypes), "|")),
```
adombeck, line 41: Not introduced by this PR but we should sort this slice to have a deterministic usage message. Currently, `cifuzz init --help` prints the possible values in a different order each time it's called.

adombeck:
I like the solution but we should be aware that it's a breaking change. So if for example users call `cifuzz init cpp` or `cifuzz init java` in a script, that script will fail once they update to the next cifuzz release. If we're fine with that, we should mention it in the release notes.

internal/cmd/init/init.go
adombeck: I understand it as: It's fine to use `opts.testLang` as a key of the `supportedInitTestTypes` map here because we validated above that `opts.testLang` is of type `config.FuzzTestType` which means that it's also a valid key of the `supportedInitTestTypes` map. I also think that the comment could be improved to make it more clear.

MarkusZoppelt (approved):

internal/cmd/init/init.go
adombeck: @JackuB Could try to make the comment more clear?

internal/cmd/init/init.go
JackuB: Added a sorting function to ensure ts and js stay next to each other.

PR #936, opened 2023-07-21
fix: add NodeJS API fuzz target

MarkusZoppelt:
### Motivation/Context

`cifuzz run` failed with an error during upload because the NodeJS build system was not handled as a known API fuzz test config.

### Description

Adds the missing case for NodeJS and the API fuzz test config.

### How to use/reproduce

`cifuzz run` with upload to CI Sense. Compare to behavior before and after this PR.

**💝 Bonus feature:**
* I added a new function to the mock server that allows to assert on contents of the API requests. You can see it in action in `integration-tests/shared/run_with_upload.go` and I added `testRunWithUpload()` to the `Maven` and `NodeJS` integration tests (CMake already had this) to assert for the correct `engine` value. Special thanks to @jochil [who had the idea](https://github.com/CodeIntelligenceTesting/cifuzz/pull/941#issuecomment-1650032587).


jochil:
Isn't it always the fuzz test name we have to add to the request, why having multiple structures to handle it? Can you maybe think of a way that is not depended on the build system, making this change unnecessary when supporting the next language/build system?

internal/api/campaign_run.go
```diff
@@ -103,6 +103,8 @@ func (client *APIClient) CreateCampaignRun(project string, token string, fuzzTar
 		fuzzTargetConfig.CAPIFuzzTarget = &CAPIFuzzTarget{APIFuzzTarget: apiFuzzTarget}
 	case config.BuildSystemMaven, config.BuildSystemGradle:
 		fuzzTargetConfig.JavaAPIFuzzTarget = &JavaAPIFuzzTarget{APIFuzzTarget: apiFuzzTarget}
+	case config.BuildSystemNodeJS:
```
JackuB, line 106: Since this is a regression fix, this case is currently not covered by tests, right?

MarkusZoppelt: Do you know a good place to test this?
We currently don't have a clear place where to check property of the contents of API requests, yet.
https://github.com/CodeIntelligenceTesting/cifuzz/pull/899 faced a similar issue.

MarkusZoppelt: I mean we _could_ print the info in `verbose` mode and check for the output... but that doesn't seem very elegant.

MarkusZoppelt: Never mind. I changed the mock server to assert for contents of the API call body.

integration-tests/shared/run_with_upload.go
```diff
@@ -28,6 +28,20 @@ func TestRunWithUpload(t *testing.T, dir string, cifuzz string, args ...string)
...
+		server.AssertRequestBodyContains(t, fmt.Sprintf("/v1/projects/%s/campaign_runs", projectName), "c_api")
+		server.AssertRequestBodyContains(t, fmt.Sprintf("/v1/projects/%s/campaign_runs", projectName), "LIBFUZZER")
+	case "com.example.FuzzTestCase":
+		server.AssertRequestBodyContains(t, fmt.Sprintf("/v1/projects/%s/campaign_runs", projectName), "java_api")
+		server.AssertRequestBodyContains(t, fmt.Sprintf("/v1/projects/%s/campaign_runs", projectName), "JAVA_LIBFUZZER")
+	case "FuzzTestCase":
+		server.AssertRequestBodyContains(t, fmt.Sprintf("/v1/projects/%s/campaign_runs", projectName), "nodejs_api")
+		server.AssertRequestBodyContains(t, fmt.Sprintf("/v1/projects/%s/campaign_runs", projectName), "JAZZER_JS")
+	}
+
```
jochil, line 44: Nice having that test kind of test now :+1: 

I would suggest one change: Currently `AssertRequestBodyContains()` is used differently than the other response handler. So wdyt about using it something like this:
```
server.Handlers[path] = mockserver.BodyContains(t, "foo")
```

This way it would be consistent with the use of `ReturnResponse`
We can also think about combining functions or chaining them, for example:
```
server.Handlers[path] = mockserver.CheckBodyAndReturn(t, "foo", "{}")
server.Handlers[path] = mockserver.BodyContains(t, "foo").ReturnResponse(t, "{}")
```
But no need to do it know

jochil (approved):

integration-tests/shared/run_with_upload.go
MarkusZoppelt: Nice idea!

>So wdyt about using it something like this:
>
>    `server.Handlers[path] = mockserver.BodyContains(t, "foo")`

I don't think the first option wouldn't work in the case we would like to call `BodyContains` multiple times, e.g.:

```go
server.Handlers[path] = mockserver.BodyContains(t, "foo")
server.Handlers[path] = mockserver.BodyContains(t, "bar") // overwrites previous handler
```

I like the idea of chaining calls, tho.
Will think of a good solution in another PR.

MarkusZoppelt: CLI-1172

PR #838, opened 2023-06-15
Run E2E tests in containers

JackuB:
mfw

- We want to E2E test our new cifuzz container commands and those rely on Docker socket being available (*for now! :partyparrot:). To make this work, we need a working Docker Engine/Socket. This is possible to achieve on GitHub Actions runners.
- With this PR, every single E2E test case runs in its own container - preventing test cross-pollination and hopefully enabling more parallelisation options. Seeing how genhtml (and similar issues) threw off our tests and a release pipelines, I’d say we should strive for better test environment isolation and reproducibility. Working with [cached] and predictable containers could be a way for us to achieve that.
- We also want to test more and more complex setups and environments and being able to containerize sample projects, their dependencies and configuration should allow us to greatly extend our E2E coverage.

This PR only updated the E2E Runner, no changes to the tests were needed. Code changes are not that major, as the current E2E structure was kept, but I moved some of the logic into separate functions and files for brevity.

But this PR is also dropping testing pipeline for macOS (we can introduce a limited (--short?) version of E2E verification just for macOS in a following PRs).

This PR is also removing release pipeline testing, as we should consider a slightly different approach. And the current one is not being used beyond a PoC.

PR #937, opened 2023-07-24
Test cifuzz container run command

JackuB:
Adds E2E tests for `cifuzz container run` command.
To make this work, E2E tests are now running in isolated containers https://github.com/CodeIntelligenceTesting/cifuzz/pull/838
And this PR introduces capabilities to add required tools and configs to the E2E containers - for example tooling like Java etc.

Neat thing about this setup is that it also allows you to test Fuzz Containers _inside a container_.
`cifuzz container run` tests skip running on Windows because it adds complexity and CI runtime that we don't currently need.

### How to use/reproduce

Running `make test/e2e` should work locally and you can inspect `cifuzz-e2e-test` containers for full output and details.


Makefile
```diff
@@ -286,7 +288,8 @@ push-container-image: build-container-image
 	docker push "$(image_tag)"
 
 .PHONY: installer-via-docker
+installer-via-docker: export DOCKER_BUILDKIT = 1
 installer-via-docker:
 	@echo "Building a cifuzz Linux installer"
 	mkdir -p build/bin
-	DOCKER_BUILDKIT=1 docker build --platform linux/amd64 -f docker/cifuzz-builder/Dockerfile . --target bin --output build/bin
+	docker build --platform linux/amd64 -f docker/cifuzz-builder/Dockerfile . --target bin --output build/bin
```
adombeck, line 298: Why do you remove `DOCKER_BUILDKIT=1`? Was buildkit causing problems?

JackuB: Yup, docker for windows in GH Actions is not supporting it out of the box.

But I'm keeping it in an if/else block for now

MarkusZoppelt (approved):
LGTM!
Hera are the test runs succeeding: https://github.com/CodeIntelligenceTesting/cifuzz/actions/runs/5654616997

PR #939, opened 2023-07-25
Add source map to java bundles

florianGla:
This PR add a source_map.json file to java bundles. This file maps all java and kotlin source files to
the corresponding package name. This is needed for CI-Sense to reconstruct the source file
location of a finding. The stack trace only provides the source file name and the package name.
Note that this is only a heuristic. There can be cases where the source file location cannot
be reconstructed or where ambiguities can arise. This is a best effort approach, which should work more the majority of cases.

### How to use/reproduce

Create a java bundle and upload it to CI-Sense. If the project is connected with github, the source file location of the finding can be viewed in the linked github repo.


internal/bundler/jazzer_bundler.go
```diff
@@ -377,3 +407,44 @@ func (b *jazzerBundler) getTestDirs() ([]string, error) {
...
 	return testDirs, nil
 }
+
+func (b *jazzerBundler) createSourceMap() (*SourceMap, error) {
+	sourceMap := SourceMap{
+		JavaPackages: make(map[string][]string),
+	}
+
+	// Use zglob to find all java or kotlin files in the project dir.
+	// Searching the whole project dir is a temporary solution
```
mlsandnerCI, line 417: Maybe add a `TODO:` here so it can be easily found when we can remove the temporary solution.

mlsandnerCI (approved):

florianGla:
I restructured the java integration tests. This was needed in order to resolve a package conflict. The test buckets list specific go packages, which were changed by my PR. So the previous test buckets contained invalid packages and did not contain the new packages.

PR #940, opened 2023-07-25
Test container run with CMake

adombeck:
### Motivation/Context

We support the container run command with CMake, so we should have a test for that feature.

### Description

* Add a Dockerfile to the cmake example which contains the dependencies required to use the container run command in that project.

* Add a step in our "Run cifuzz examples" GitHub action which tests that the container run command works in the cmake example project.

* Add an integration test which tests that the container run command works with a CMake project.

### How to use/reproduce

To run the integration test:
```
go test -v ./integration-tests/cmake/... -run TestIntegration_CMake/containerRun
```

To run the GitHub action, trigger one of our E2E pipelines.


jochil:
@JackuB isn't this something that can be done in the `e2e-tests` as well? Like here: https://github.com/CodeIntelligenceTesting/cifuzz/blob/main/e2e-tests/container_run/container_run_test.go

adombeck:
> isn't this something that can be done in the e2e-tests as well? Like here: https://github.com/CodeIntelligenceTesting/cifuzz/blob/main/e2e-tests/container_run/container_run_test.go

I tried that first and spent most of the time I worked on this branch on that until I noticed that `runTestCaseInContainer()` doesn't install cifuzz in the container but just bind-mounts the executable which means that it's missing the CMake files which are needed to use cifuzz with CMake. I found extending the "integration tests" (which are actually e2e tests as well AFAICT) a lot easier because they don't try to use the same function for all build systems but only share code which works for all supported build systems.

integration-tests/cmake/cmake_test.go
```diff
@@ -167,6 +170,13 @@ func TestIntegration_CMake(t *testing.T) {
 	t.Run("runNotAuthenticated", func(t *testing.T) {
 		testRunNotAuthenticated(t, cifuzzRunner)
 	})
+
+	t.Run("containerRun", func(t *testing.T) {
+		if runtime.GOOS != "linux" && !config.AllowUnsupportedPlatforms() {
+			t.Skip("The container run command is currently only supported on Linux")
```
MarkusZoppelt, line 176: Is it? 🤔
It should also run on macOS. Not sure about Windows.

adombeck: We only support building bundles with CMake on Linux. I'll update the string to make that clear.

adombeck: done

MarkusZoppelt (approved):
LGTM, although we could also use `e2e-tests` for this, similar to #937

adombeck:
> LGTM, although we could also use e2e-tests for this, similar to https://github.com/CodeIntelligenceTesting/cifuzz/pull/937

I explained [in the comment above](https://github.com/CodeIntelligenceTesting/cifuzz/pull/940#issuecomment-1651300859) why I ended up adding it to the integration tests instead of the e2e tests.

PR #941, opened 2023-07-25
fix: set correct Engine value in campaign run (for finding upload)

MarkusZoppelt:
### Motivation/Context

The `engine` field in the campaign run object was always populated with `LIBFUZZER`, regardless of the language and build system of the project.

### Description

This PR now sets the correct `engine` value depending on the build system.

Note: The value `JAVA_LIBFUZZER` is temporary and will replaced with `JAZZER` in the future. (The backend expects `JAVA_LIBFUZZER` for now.)

### How to use/reproduce

 * run `cifuzz run` (with upload) to see that the engine field is set correctly.

### Testing issues:

Unfortunately, we cannot easily test/assert the correct value of our API requests.
This has already causes problems in these PRs:
https://github.com/CodeIntelligenceTesting/cifuzz/pull/936#discussion_r1273554118
https://github.com/CodeIntelligenceTesting/cifuzz/pull/899

 We should think of a way to easily test content of our API calls...


jochil (approved):

MarkusZoppelt:
> defining a handler for the `mock.server` that uses `Assert` for example?

thx, this was helpful. I implemented this as part of #936.

PR #944, opened 2023-07-27
Use maven test source dir from pom.xml

florianGla:
For maven projects we should check if the test source directory is configured in the pom.xml file and use it in this case.

### Motivation/Context

Without this change no fuzz tests are detected for projects with custom test source directories.

### Description

We parse the pom.xml file and check if a test source directory is defined.

### How to use/reproduce

Check if `cifuzz bundle` bundles fuzz test for this example project: https://github.com/CodeIntelligenceOSSOnboardings/google-auth-library-java/tree/1.19.0-fuzzing/oauth2_http


internal/cmdutils/resolve/testdata/maven/src/test/kotlin/com/example/fuzz_test_2/FuzzTestCase.kt
adombeck, line 1: This file seems to be unused

internal/cmdutils/resolve/testdata/maven/src/test/java/com/example/fuzz_test_2/FuzzTestCase.java
adombeck, line 1: This file seems to be unused

adombeck (approved):

internal/cmdutils/resolve/testdata/maven/src/test/java/com/example/fuzz_test_2/FuzzTestCase.java
adombeck: @florianGla  I see that the file is still there - was it actually required?

florianGla: Yes, we add extra unused source files for every resolve test in order to check that the correct one gets resolved.

PR #945, opened 2023-07-27
feat(mockserver): improve API body content checking

MarkusZoppelt:
### Motivation/Context

To ensure that our tests catch errors in the API requests we are sending to a backend, we need to check the contents in the actual request we are sending.
A simple approach is to check if the body of the request `Contains` or `NotContains` certain data.

### Description

This change improves the `ReturnResponse` feature of our mockserver to optionally check for `expected` or `unexpected` strings in the request body.

### How to use/reproduce

Run one of these integration tests:

```bash
go test ./integration-tests/cmake/... -run TestIntegration_CMake/runWithUpload
go test ./integration-tests/java-maven/... -run TestIntegration_Maven/runWithUpload
go test ./integration-tests/java-gradle/... -run TestIntegration_Gradle/runWithUpload
go test ./integration-tests/nodejs/... -run TestIntegration_NodeJS/runWithUpload
go test ./integration-tests/node-typescript/... -run TestIntegration_NodeTS/runWithUpload
```

image


adombeck:
When looking at the commit changes + commit message, I couldn't figure out the reason for this change. Only in combination with the PR description it made sense. If the new focus on the PR description (which is now enforced by the bot) leads to less descriptive commit messages then I think that's a step in the wrong direction, because commit messages are a lot easier to access later on when reading the code and wondering why something was implemented this way.

integration-tests/shared/mockserver/server.go
```diff
@@ -58,9 +50,23 @@ func (server *MockServer) Start(t *testing.T) {
 }
 
 func ReturnResponse(t *testing.T, responseString string) http.HandlerFunc {
+	return CheckBodyAndReturnResponse(t, responseString, nil, nil)
+}
+
+func CheckBodyAndReturnResponse(t *testing.T, responseString string, expected *[]string, unexpected *[]string) http.HandlerFunc {
```
adombeck, line 56: Why do use `*[]string` instead of `[]string`? That's unconventional. Quoting `Go Code Review Comments`[https://github.com/golang/go/wiki/CodeReviewComments#receiver-type]:
> If the receiver is a slice and the method doesn't reslice or reallocate the slice, don't use a pointer to it.

integration-tests/shared/mockserver/server.go
```diff
@@ -58,9 +50,23 @@ func (server *MockServer) Start(t *testing.T) {
...
+		if expected != nil {
+			for _, expectedString := range *expected {
+				require.Contains(t, string(body), expectedString)
+			}
+		}
+		if unexpected != nil {
+			for _, unexpectedString := range *unexpected {
+				require.NotContains(t, string(body), unexpectedString)
+			}
+		}
```
adombeck, line 69: If you don't pass `expected` and `unexpected` as pointers, you can simplify this to

```suggestion
		for _, expectedString := range expected {
			require.Contains(t, string(body), expectedString)
		}
		for _, unexpectedString := range unexpected {
			require.NotContains(t, string(body), unexpectedString)
		}
```

integration-tests/shared/mockserver/server.go
MarkusZoppelt: You're totally right, thanks! Changed.

MarkusZoppelt:
> When looking at the commit changes + commit message, I couldn't figure out the reason for this change. Only in combination with the PR description it made sense. 

I added more info to the first commit message

adombeck (approved):

PR #946, opened 2023-07-27
CLI-1168 Set Docker entrypoint

adombeck:
### Motivation/Context

By setting the entrypoint to "cifuzz execute", we allow the user to run the fuzz
test via `docker run cifuzz my_fuzz_test` instead of `docker run cifuzz
cifuzz execute my_fuzz_test`.

### Description

Beside adding a Docker entrypoint, the PR contains multiple fixes for the container run command which are explained in the commit messages. 

### How to use/reproduce

To test the entrypoint, first create a container by executing `cifuzz container run` once:

```
CIFUZZ_PRERELEASE=1 cifuzz -C examples/gradle container run com.example.FuzzTestCase::myFuzzTest
```

Then execute the docker image directly and specify the fuzz test name as an argument to the entrypoint:
```
docker run cifuzz com.example.FuzzTestCase::myFuzzTest
```


internal/container/signals_unix.go
MarkusZoppelt, line 1: Any particular reason why this and signals_windows.go are two separate files?
Why don't we check for `runtime.GOOS` during runtime? Is the check for OS during compile time that important?

Just curious...

adombeck: I copied it from https://github.com/docker/cli/blob/e0e27724390cbce21cf6d67568972a4227b07382/cli/command/container/signals_unix.go. I also thought about using `runtime.GOOS` instead but I think Docker's approach has a performance benefit because it doesn't require a check during runtime each time the method is called.

MarkusZoppelt: Maybe we can use this style instead of `runtime.GOOS` in more places then, if it's a noticeable performance boost.

Something we can investigate/revisit in the future.

MarkusZoppelt (approved):

PR #947, opened 2023-07-28
Use execute command as entrypoint for fuzz containers

JackuB:
Using `ENTRYPOINT` for fuzz containers means you can start a fuzz test only by providing a fuzz target - `docker run <image> <fuzz test id>` instead of providing `cifuzz[.exe] execute`

### Motivation/Context

This will make it easier for CI Sense to start containers.

### Description

Change is tested by E2E.

### How to use/reproduce

If you use `CIFUZZ_PRERELEASE=1 cifuzz container run -C examples/gradle com.example.FuzzTestCase::myFuzzTest` and then inspect and re-run the fuzz container produced by this command, you can specify just the fuzz target, like in the example above.


JackuB:
Closing in favor of #946

PR #950, opened 2023-07-31
feat: wrapcheck linter

mlsandnerCI:
### Motivation/Context

We decided on new guidelines for error handling and one point is to make sure that we add a stack trace to errors returned from external functions or wrap them (which also includes a stack trace) in a helpful error message.

### Description

Added the [wrapcheck](https://github.com/tomarrell/wrapcheck) linter to check that we add stack traces to errors from functions from external packages (e.g. `filepath`, `url`, `os`) and resolved the linter warnings.

I explicitly ignore our own code (see `ignoreSigs` configuration in `.golangci.yaml`) because I think we should decide  on our own if we need to add context.


internal/api/api.go
```diff
@@ -203,7 +203,7 @@ func (client *APIClient) UploadBundle(path string, projectName string, token str
 
 	err := routines.Wait()
 	if err != nil {
-		return nil, err
+		return nil, errors.WithStack(err)
```
adombeck, line 206: `routines.Wait` returns the first non-nil error returned by any of the functions executed via `routines.Go`, so it returns an error defined by us which should already have a stack trace -> let's not add a duplicate stack trace here.

internal/api/api.go
```diff
@@ -340,7 +340,11 @@ func getCustomTransport() *http.Transport {
...
 	dialer := proxy.FromEnvironment()
 	dialContext := func(ctx context.Context, network, address string) (net.Conn, error) {
-		return dialer.Dial(network, address)
+		conn, err := dialer.Dial(network, address)
+		if err != nil {
+			return nil, errors.WithStack(err)
+		}
+		return conn, nil
 	}
 	return &http.Transport{DialContext: dialContext}
```
adombeck, line 349: Adding a stack trace here **could** have unexpected side effects because this error is being returned to a function in an external package (the `http` package). I don't expect that the `http` package has special handling for errors returned from the `dialer` package though (and if so, it should use `errors.As` or `errors.Is` which would automatically unwrap the error), so it should be fine.

adombeck (changes requested):
This linter seems useful!

internal/api/api.go
mlsandnerCI: Cool, thanks for the explanation!

adombeck (approved):

PR #951, opened 2023-08-01
feat(maven): add function to get source dir from pom.xml

florianGla:
This PR adds a function to get the source dir from the pom.xml. This function is needed in order to speed up the computation of the source map used in bundles.


internal/build/maven/maven.go
```diff
@@ -191,3 +191,17 @@ func GetTestDir(projectDir string) (string, error) {
 	log.Debugf("Found maven test source at: %s", project.Build.TestSourceDirectory)
 	return strings.TrimSpace(project.Build.TestSourceDirectory), nil
 }
+
+// GetSourceDir returns the value of <sourceDirectory> from the projects
+// pom.xml as an absolute path.
+// Note: If no tag is specified, the parser will return the
+// default value "projectDir/src/main/java".
+func GetSourceDir(projectDir string) (string, error) {
```
MarkusZoppelt, line 199: Can you explain why this is necessary? As far as I can see, you only use this function in tests... shouldn't it better be a test helper? What exactly does it speed up?

MarkusZoppelt: Alright, thanks. Maybe add this to the commit description as well 👍🏼

MarkusZoppelt (approved):

PR #952, opened 2023-08-01
fix: print and upload correct exec/s number

MarkusZoppelt:
CLI-1145


adombeck:
>  fix(reporthandler): print n/a if exec/s is 0 

Why is printing "n/a" better than "0" if the stats which we get from the fuzzer indicate that there are 0 executions per second? What about the case that the fuzz test is stuck or just super slow - wouldn't that cause the exec/s indicated by the fuzzer to be 0? And wouldn't that be a useful information to the user?

internal/api/campaign_run.go
```diff
@@ -166,3 +134,46 @@ func (client *APIClient) CreateCampaignRun(project string, token string, fuzzTar
...
+				Features:                 firstMetrics.Features,
+				CorpusSize:               firstMetrics.CorpusSize,
+				SecondsSinceLastCoverage: fmt.Sprintf("%d", firstMetrics.SecondsSinceLastFeature),
+				TotalExecutions:          fmt.Sprintf("%d", firstMetrics.TotalExecutions),
+				Edges:                    firstMetrics.Edges,
+				SecondsSinceLastEdge:     fmt.Sprintf("%d", firstMetrics.SecondsSinceLastEdge),
+			},
+		}
+	}
+	return metricsList
```
adombeck, line 178: Having the refactoring and the other changes in separate commits would make this easier to review.

JackuB: Rebased and split the commit

JackuB:
> Why is printing "n/a" better than "0" if the stats which we get from the fuzzer indicate that there are 0 executions per second? What about the case that the fuzz test is stuck or just super slow - wouldn't that cause the exec/s indicated by the fuzzer to be 0? And wouldn't that be a useful information to the user?

If we are talking about this case https://github.com/CodeIntelligenceTesting/cifuzz/pull/952/files#diff-9a64a71516d4bf83b9bacb08ef7135d960cabca683df16a024bb0434d1162f00R319-R323

The case we observed was sub-one-second fuzzing runs, where libfuzzer reports a zero for a first second.
I'm not married to that choice, but =<0 execs/s indicates to me that something is off and we were not able to calculate the value properly.

jochil (approved):

PR #955, opened 2023-08-03
docs: cifuzz container

JackuB:
Adding public documentation for the `cifuzz container` functionality.


docs/Fuzz-Containers.md
```diff
@@ -0,0 +1,91 @@
...
+> This page is documenting a limited preview of the Fuzz Containers. Functionality is experimental and we are open to feedback.
+
+We aim to improve the experience of running fuzz tests in [CI Sense](https://www.code-intelligence.com). Container-based workflow proposed by the `cifuzz container` commands should replace our current bundle mechanism and become a primary way of running fuzz tests in CI Sense.
+
+## Motivations
+
+We want to solve multiple user problems with the current `cifuzz bundle` and `cifuzz remote-run` commands. To name a few:
+
+- **Reproducibility and debugging**: container images are providing a good base for replaying and debugging issues. Either locally or in CI Sense. Current bundles are not immutable and are not providing local reproducibility.
+- **Removing runtime limitations**: allowing users to rely on repeatable container runtimes we aim to allow better precision and reusability of fuzzing workloads.
```
adombeck, line 14: I don't understand what you mean with "allowing users to rely on repeatable container runtimes".

docs/Fuzz-Containers.md
```diff
@@ -0,0 +1,91 @@
...
+`cifuzz container` commands provide a way to build a Fuzz Container image from a project directory. The `run` command is a way to run a Fuzz Container locally:
+
+```terminal
+cifuzz container run -C examples/gradle com.example.FuzzTestCase::myFuzzTest
+```
+
+This command will:
+
+1. Build and **bundle your Fuzz Tests** using the functionality of the `cifuzz bundle` command.
+1. **Uses the bundle to build a Fuzz Container image** including a copy of `cifuzz` and contents of the bundle. Base Docker image can be set by the user with a command-line flag.
```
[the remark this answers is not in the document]
JackuB: > internal details which are not really interesting for the user

Yea, I was going back and forth there about this part too, but what let me keep it is the fact that we expose the "bundle" concept as part of the current `cifuzz bundle` command.

I wanted to communicate that it's not a _magic_ container, but it's an existing concept.

docs/Fuzz-Containers.md
JackuB: Expanded this section "with dependencies and settings fixed in time as part of the container image" - is that better?

adombeck: Yes, thanks!

docs/Fuzz-Containers.md
```diff
@@ -0,0 +1,91 @@
...
+    cifuzz-."$ cifuzz container run".->project-->container
+
+    subgraph project[Local Project files]
+        bundle[Bundled fuzz tests]
+        tests[Fuzz Tests]-->bundle
+        binaries[Instrumented application binaries]-->bundle
+    end
+```
+
+You can define the Docker image used for creating the Fuzz Container with a `--docker-image` flag. Or if you already have a Fuzz Container, you can use the `--container` flag to run it.
```
adombeck, line 79: I didn't know about the `--container` flag and after taking a look at the code, it's functionality doesn't seem to be implemented yet.

docs/Fuzz-Containers.md
```diff
@@ -0,0 +1,91 @@
...
+
+See `cifuzz container run --help` for a full list of options.
+
+Currently, the building and bundling of a Fuzz Tests happens on a local machine. But even this step could be scripted and moved into a container in the future, allowing repeatable and reproducible builds everywhere.
```
adombeck, line 83: The building and bundling happens wherever you execute the `cifuzz container run/remote-run` command. I think that could already be inside a Docker container (which has access to a Docker daemon). I'm skeptical if it is a good idea to have it as a built-in feature to execute the build in a specified Docker image but I'm open to discuss it.

docs/Fuzz-Containers.md
```diff
@@ -0,0 +1,91 @@
...
+```
+
+Because Fuzz Container is a Container Image, it can be freely shared, stored and inspected.
+
+## Building and running a Fuzz Container locally
+
+`cifuzz container` commands provide a way to build a Fuzz Container image from a project directory. The `run` command is a way to run a Fuzz Container locally:
+
+```terminal
+cifuzz container run -C examples/gradle com.example.FuzzTestCase::myFuzzTest
```
adombeck, line 56: I think we should also have an example for a C/C++ fuzz test and explain that those require specifying a Docker image which has LLVM installed, because figuring out how to get that to work is a lot more complex than in the Java case.

docs/Fuzz-Containers.md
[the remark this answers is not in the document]
adombeck: Ok, I'm fine with that.

docs/Fuzz-Containers.md
JackuB: Merde, I thought I removed all references to that already.

Removed

docs/Fuzz-Containers.md
JackuB: Added a note to the doc

adombeck (approved):

PR #961, opened 2023-08-04
CLI-1176 Test `cifuzz container run` with other build systems

adombeck:
### Motivation/Context

We support the container run command with other build systems, so we should have a test for that feature.

### Description

Beside the new integration test of the `container run` command for other build systems, the PR contains some refactorings and fixes which are described in the respective commit messages.

### How to use/reproduce

To run the new test:

```
go test -v ./integration-tests/other/... -run TestIntegration_Other/containerRun
```


internal/container/container.go
```diff
@@ -39,7 +39,8 @@ func Create(fuzzTest string) (string, error) {
 	// This is useful for debugging, as it allows to exec into the container,
 	// run the command manually and debug things in the container.
 	if os.Getenv("CIFUZZ_CONTAINER_SLEEP") != "" {
-		containerConfig.Env = append(containerConfig.Env, "CMD="+strings.Join(containerConfig.Cmd, " "))
+		cmd := "cifuzz execute " + strings.Join(containerConfig.Cmd, " ")
+		containerConfig.Env = append(containerConfig.Env, "CMD="+cmd)
```
mlsandnerCI, line 43: How did you find out that this didn't work? Could you write a test reproducing it?

mlsandnerCI (approved):

internal/container/container.go
adombeck: I think it would be pretty hard to write a test for the `CIFUZZ_CONTAINER_SLEEP` feature. IMO it's fine that we don't have a test for it because it's only meant for development and debugging so it's not critical that it works as expected.

PR #975, opened 2023-08-09
feat(jazzer_bundler): make sure that dependency names are unique

florianGla:
There are some projects for which the base names of all jar files are not unique. This PR appends an incrementing counter to those files, when creating a bundle.

### Motivation/Context

Some java projects cannot be bundled.

### How to use/reproduce

Example project: https://github.com/hsk-ci/WebGoat-1/tree/fuzz


internal/bundler/jazzer_bundler.go
```diff
@@ -152,9 +156,14 @@ func (b *jazzerBundler) assembleArtifacts(buildResults []*build.Result) ([]*arch
 					return nil, err
 				}
 			} else {
-				// if the current runtime dependency is a file we add it to the
-				// archive and add the runtime paths of the metadata
-				archivePath := filepath.Join(runtimeDepsPath, filepath.Base(runtimeDep))
+				// If the current runtime dependency is a file, we check if
+				// it's name is unique and append a hash of the file contents
+				// otherwise.
+				artifactName, err := getUniqueArtifactName(&buildResult.RuntimeDeps, idx)
```
fmeum, line 162: If we change the artifact name in the archive, do we also have to update the classpath?

internal/bundler/jazzer_bundler.go
```diff
@@ -450,3 +459,28 @@ func (b *jazzerBundler) createSourceMap() (*SourceMap, error) {
 
 	return &sourceMap, nil
 }
+
+func getUniqueArtifactName(runtimeDeps *[]string, index int) (string, error) {
+	dep := (*runtimeDeps)[index]
+	for _, otherDep := range *runtimeDeps {
```
fmeum, line 465: Since we are calling this function in a loop over `runtimeDeps`, so this loop will cause the whole bundling process to have runtime quadratic rather than linear in the number of deps.

Instead, how about moving this into a preprocessing step that collects all base names in a `map[string]bool` and then using that map in the real bundling step to decide whether to add a hash?

florianGla: Great idea

internal/bundler/jazzer_bundler.go
florianGla: No, I think the bundles do not contain an explicit class path.

internal/bundler/jazzer_bundler.go
```diff
@@ -36,10 +40,13 @@ const runtimeDepsPath = "runtime_deps"
 type jazzerBundler struct {
 	opts          *Opts
 	archiveWriter archive.ArchiveWriter
+
+	// This map is used to check if the base name of a runtime dependency is unique
+	artifactsMap map[string]bool
```
fmeum, line 45: It doesn't look like this really has to be a field on the `jazzerBundler` struct. Could you make it a local variable in `assembleArtifacts` instead?

internal/bundler/jazzer_bundler.go
```diff
@@ -450,3 +462,29 @@ func (b *jazzerBundler) createSourceMap() (*SourceMap, error) {
 
 	return &sourceMap, nil
 }
+
+func (b *jazzerBundler) getUniqueArtifactName(dependency string) (string, error) {
+	baseName := filepath.Base(dependency)
+	_, found := b.artifactsMap[baseName]
+	if !found {
+		// If the base name of the dependency is unique, we add it to the map
```
fmeum, line 470:
```suggestion
		// If the base name of the dependency hasn't been seen before, we add it to the map
```

internal/bundler/jazzer_bundler.go
```diff
@@ -450,3 +462,29 @@ func (b *jazzerBundler) createSourceMap() (*SourceMap, error) {
...
+		// and return it.
+		b.artifactsMap[baseName] = true
+		return baseName, nil
+	}
+
+	// If the base name of the dependency is not unique, we append a hash
```
fmeum, line 476:
```suggestion
	// If the base name of the dependency has been seen before, we append a hash
```

internal/bundler/jazzer_bundler.go
```diff
@@ -450,3 +462,29 @@ func (b *jazzerBundler) createSourceMap() (*SourceMap, error) {
...
+		return "", errors.WithStack(err)
+	}
+	hash := sha256.New()
+	err = binary.Write(hash, binary.BigEndian, bytes)
+	if err != nil {
+		return "", errors.WithStack(err)
+	}
+	hashString := base32.StdEncoding.EncodeToString(hash.Sum(nil))[:8]
+	artifactBaseName := strings.TrimSuffix(filepath.Base(dependency), filepath.Ext(dependency))
+	return fmt.Sprintf("%s-%s.jar", artifactBaseName, hashString), nil
```
fmeum, line 489: What I only noticed now: If there are three files with identical basenames, two of which also happen to have the same content, then the last two will have the same basename even after appending the hash.

Instead of appending a hash, we could make `artifactsMap` a `map[string]uint` and just append an incrementing counter in case of collisions.

internal/bundler/jazzer_bundler.go
```diff
@@ -450,3 +455,21 @@ func (b *jazzerBundler) createSourceMap() (*SourceMap, error) {
...
+		artifactsMap[baseName] = 1
+		return baseName
+	}
+
+	// If the base name of the dependency has been seen before, we increment its
+	// count in the map and append the count to the artifact base name.
+	artifactsMap[baseName]++
+
+	artifactBaseName := strings.TrimSuffix(filepath.Base(dependency), filepath.Ext(dependency))
+	return fmt.Sprintf("%s-%d.jar", artifactBaseName, count)
```
fmeum, line 474: You also have to add the new basename to the map in case our suffix scheme ends up creating a collision (should be very rare though).

fmeum (approved):

PR #977, opened 2023-08-09
chore(deps): update jazzer.js to version 1.6.1

florianGla:
Update jazzer.js dependency of our javascript/typescript example projects to the latest version.


jochil (changes requested):
I think it should be version 1.6.1 instead of 1.6.0
And usually we use something like `chore(deps): ` as a conventional commit message for this kind of version updates

florianGla:
> I think it should be version 1.6.1 instead of 1.6.0 And usually we use something like `chore(deps): ` as a conventional commit message for this kind of version updates

Sorry, I did this PR too fast without thinking too much :)

jochil (approved):

PR #978, opened 2023-08-10
fix: use stable input for finding name generation

jochil:
**BREAKING CHANGE:** As we had to change the seed for our deterministic name generation, existing findings will be duplicated with a new name.

### Motivation/Context
Keeping your tool chain up to date is of course very desirable 

### Description
- update all usages of go from 1.20 to 1.21
- using a custom string representation of the stacktrace as part of the input for the name generation, due to changes in `encoding/gob` (more explanation in the commit message/pr comments)

### How to use/reproduce
To run the test that was failing after the upgrade you can use `go test ./internal/cmd/run/reporthandler/... -short -v`

To reproduce the duplication of findings with new names you can do the following:
```
sh -c "$(curl -fsSL https://raw.githubusercontent.com/CodeIntelligenceTesting/cifuzz/main/install.sh)"
cifuzz -C examples/cmake run my_fuzz_test --interactive=false
cifuzz -C examples/cmake  findings
make install
cifuzz -C examples/cmake run my_fuzz_test --interactive=false
cifuzz -C examples/cmake  findings
```


florianGla (approved):

florianGla:
It looks like the go version update changed our generated names for findings

jochil:
> It looks like the go version update changed our generated names for findings

I noticed this too, but still investigating... :thinking:

MarkusZoppelt:
> > It looks like the go version update changed our generated names for findings
> 
> 
> 
> I noticed this too, but still investigating... :thinking: 

see https://go.dev/blog/compat

jochil:
> > > It looks like the go version update changed our generated names for findings
> > 
> > 
> > I noticed this too, but still investigating... thinking
> 
> see https://go.dev/blog/compat

Is there something specific you think that might responsible? I ask because from quickly scanning it I see nothing related

adombeck:
> 
> 
>     This results in different names being generated for the same finding,
>     but this is not a breaking change as existing findings will not be
>     renamed or created again with the new name.

I tried to test that and it seems to be the case that at least for our example projects, no new findings are produced. AFAICT, the reason for that is that for all the findings of our example project, the stack trace encoding is the same with Go 1.20 and Go 1.21. Maybe the only thing that changed is the encoding of `nil`? That would mean that we only generate a different display name for findings which don't have a stack trace.

jochil:
> > ```
> > This results in different names being generated for the same finding,
> > but this is not a breaking change as existing findings will not be
> > renamed or created again with the new name.
> > ```
> 
> I tried to test that and it seems to be the case that at least for our example projects, no new findings are produced. AFAICT, the reason for that is that for all the findings of our example project, the stack trace encoding is the same with Go 1.20 and Go 1.21. 



first 1.20 with gob encoder
go120-gob

then 1.21 with json encoder
go121-json

I tried it again and now I got another finding... I am not sure if I made a mistake when testing it in the first place or if there was an actual reason, especially as you observed the same behavior :thinking: 
I checked related code and this is actual how it should be... so it is a breaking change (if we don't add code to handle it)

> Maybe the only thing that changed is the encoding of `nil`? That would mean that we only generate a different display name for findings which don't have a stack trace.

I explicitly tested if it only broke for empty stack traces, this is why I added an extra test to cover the case with stack traces as well. Unfortunately the change affects also findings with a stack trace

adombeck:
When using `gob.Encoder`, I get the same finding name in our example projects with Go 1.20 and 1.21. This is output with Go 1.20:

```
f.StackTrace: [0xc0001dc4e0]
f.InputData: [64 106 97 122 46 90 101 114 64 64 64 64 64 64 127 117]
f.Bytes: [13 255 131 2 1 2 255 132 0 1 255 130 0 0 76 255 129 3 1 2 255 130 0 1 5 1 10 83 111 117 114 99 101 70 105 108 101 1 12 0 1 4 76 105 110 101 1 6 0 1 6 67 111 108 117 109 110 1 6 0 1 11 70 114 97 109 101 78 117 109 98 101 114 1 6 0 1 8 70 117 110 99 116 105 111 110 1 12 0 0 0 41 255 132 0 1 1 21 99 111 109 46 101 120 97 109 112 108 101 46 69 120 112 108 111 114 101 77 101 1 19 3 9 101 120 112 108 111 114 101 77 101 0]
f.nameSeed: [13 255 131 2 1 2 255 132 0 1 255 130 0 0 76 255 129 3 1 2 255 130 0 1 5 1 10 83 111 117 114 99 101 70 105 108 101 1 12 0 1 4 76 105 110 101 1 6 0 1 6 67 111 108 117 109 110 1 6 0 1 11 70 114 97 109 101 78 117 109 98 101 114 1 6 0 1 8 70 117 110 99 116 105 111 110 1 12 0 0 0 41 255 132 0 1 1 21 99 111 109 46 101 120 97 109 112 108 101 46 69 120 112 108 111 114 101 77 101 1 19 3 9 101 120 112 108 111 114 101 77 101 0 64 106 97 122 46 90 101 114 64 64 64 64 64 64 127 117]
💥 [wizardly_piranha] Security Issue: Remote Code Execution in exploreMe (com.example.ExploreMe:19)   
```

And this is output with Go 1.21:

```
f.StackTrace: [0xc000383470]
f.InputData: [64 106 97 122 46 90 101 114 64 64 64 64 64 64 127 117]
f.Bytes: [13 255 131 2 1 2 255 132 0 1 255 130 0 0 76 255 129 3 1 2 255 130 0 1 5 1 10 83 111 117 114 99 101 70 105 108 101 1 12 0 1 4 76 105 110 101 1 6 0 1 6 67 111 108 117 109 110 1 6 0 1 11 70 114 97 109 101 78 117 109 98 101 114 1 6 0 1 8 70 117 110 99 116 105 111 110 1 12 0 0 0 41 255 132 0 1 1 21 99 111 109 46 101 120 97 109 112 108 101 46 69 120 112 108 111 114 101 77 101 1 19 3 9 101 120 112 108 111 114 101 77 101 0]
f.nameSeed: [13 255 131 2 1 2 255 132 0 1 255 130 0 0 76 255 129 3 1 2 255 130 0 1 5 1 10 83 111 117 114 99 101 70 105 108 101 1 12 0 1 4 76 105 110 101 1 6 0 1 6 67 111 108 117 109 110 1 6 0 1 11 70 114 97 109 101 78 117 109 98 101 114 1 6 0 1 8 70 117 110 99 116 105 111 110 1 12 0 0 0 41 255 132 0 1 1 21 99 111 109 46 101 120 97 109 112 108 101 46 69 120 112 108 111 114 101 77 101 1 19 3 9 101 120 112 108 111 114 101 77 101 0 64 106 97 122 46 90 101 114 64 64 64 64 64 64 127 117]
💥 [wizardly_piranha] Security Issue: Remote Code Execution in exploreMe (com.example.ExploreMe:19)  
```

But the `TestReportHandler_GenerateName` test produces different bytes and display names.

MarkusZoppelt:
What's the decision on this? Should we accept the change and use `encoding/json` going forward?

jochil:
> What's the decision on this? Should we accept the change and use `encoding/json` going forward?

The current change will be breaking, but it looks like it is only showing up during the tests... so we have to dig deeper and maybe we are able to avoid the breaking change.
As the update is not urgent we postponed it a bit

adombeck:
This is blocking updates of pterm now (see https://github.com/CodeIntelligenceTesting/cifuzz/pull/1071#issuecomment-1722271459). We should make a decision soon.

jochil:
Ok, what do you think about adjusting just the tests? As it seems it is only affecting the test runtime we can keep the actual code as it is without a breaking change

jochil:
I tested it again and also create a little script (see test121.sh in the "debug" commit) to avoid manual errors: The gob encoder definitely produces different results between 1.20 and 1.21, but AFAICT this is not a bug in the package itself... there was never a guarantee that the byte representation is stable, as long as encode/decode works.
I don't see a way that we can produce the same results with the same encoder when updating go to 1.21. If we just change it to something stable (eg. json encoder or a stable string representation of the stack frame) we will still have a breaking change (all findings will show up again with a different name)... I think we need some kind of deduplication (checking stack frame + input) to make sure we reuse existing names. I will start a discussion with everyone affected in slack

Would be great if someone can check this out too... just to be really sure

jochil:
I implemented a custom string representation of the stack trace to use it as the input of the name generator. This way we are in control what goes into the seed. I decided against using `encoding/json` as this will break again if we delete|rename|add fields in the stack frame struct.

MarkusZoppelt (approved):
LGTM

PR #979, opened 2023-08-11
debug: use req library for tracing failed upload

JackuB:
WIP

Debug build to trace failed upload


MarkusZoppelt:
@JackuB Should we continue with this or is this stale? AFAIK it is not in our sprint and is not blocking anything.

PR #981, opened 2023-08-14
feat(jazzer-bundler): create source map from source files obtained from maven/gradle

florianGla:
Now that we can obtain all source files from maven and gradle, we only scan through those directories when creating the source map for java bundles.

### Motivation/Context

This is a performance optimization.

### How to use/reproduce

Create a bundle for maven or gradle.


internal/build/gradle/gradle.go
```diff
@@ -202,10 +203,55 @@ func GetTestSourceSets(projectDir string) ([]string, error) {
...
 	paths := strings.Split(strings.TrimSpace(result[1]), string(os.PathListSeparator))
 
-	log.Debugf("Found gradle test sources at: %s", paths)
-	return paths, nil
+	// only return valid paths
+	var sourceSets []string
+	for _, path := range paths {
+		exists, err := fileutil.Exists(path)
+		if err != nil {
+			return nil, err
```
mlsandnerCI, line 215:
```suggestion
			return nil, errors.Wrapf(err, "Gradle test source path %s does not exist", path)
```

internal/build/gradle/gradle.go
```diff
@@ -202,10 +203,55 @@ func GetTestSourceSets(projectDir string) ([]string, error) {
...
+		return nil, errors.New("Unable to parse gradle main sources.")
+	}
+	paths := strings.Split(strings.TrimSpace(result[1]), string(os.PathListSeparator))
+
+	// only return valid paths
+	var sourceSets []string
+	for _, path := range paths {
+		exists, err := fileutil.Exists(path)
+		if err != nil {
+			return nil, err
```
mlsandnerCI, line 248:
```suggestion
			return nil, errors.Wrapf(err, "Gradle main source path %s does not exist", path)
```

mlsandnerCI (approved):

internal/build/gradle/gradle.go
florianGla: I don't think that this error message makes sense. The `exists` return value indicates if the file can be found. An error is only returned if something besides `os.ErrExist` goes wrong. We just return err in all our occurences of `fileutil.Exists` in our code base.

florianGla: We could rephrase the error message, if you prefer a better error description.

mlsandnerCI: Yeah, I wasn't sure either how to phrase it but I think it makes sense to add the context what we are checking so the user not only gets a random "this does not work, do something about it".

PR #982, opened 2023-08-14
feat: remove `PRERELEASE` flag for NodeJS

MarkusZoppelt:
### Motivation/Context

JavaScript and TypeScript support is ready for release.

### Description

Removed all checks for `CIFUZZ_PRERELEASE` for NodeJS.

### How to use/reproduce

.


e2e-tests/init/init_nodejs_test.go
```diff
@@ -25,7 +25,6 @@ var nodeInitTests = &[]e2e.TestCase{
 		Description:  "init command in Node.js (JS) project with prerelease flag succeeds and creates a config file",
```
mlsandnerCI, line 25: Can you adjust the descriptions to not reference the pre-release flag? Are all of these tests even still necessary?

integration-tests/node-typescript/nodets_test.go
```diff
@@ -44,7 +44,6 @@ func TestIntegration_NodeTS_InitCreateRunCoverage(t *testing.T) {
 
 	// Execute the init command
 	instructions := cifuzzRunner.CommandWithFilterForInstructions(t, "init", &shared.CommandOptions{
-		Env:  append(os.Environ(), "CIFUZZ_PRERELEASE=1"),
```
mlsandnerCI, line 47: Don't we still need `Env: os.Environ()`? Or is this default?

mlsandnerCI (changes requested):

internal/config/types.go
```diff
@@ -36,8 +38,7 @@ const (
 // environment variable CIFUZZ_PRERELEASE.
 func SupportedTestTypes() map[string]string {
 	if os.Getenv("CIFUZZ_PRERELEASE") != "" {
-		supportedTestTypes["JavaScript"] = string(JavaScript)
-		supportedTestTypes["TypeScript"] = string(TypeScript)
+		// e.g. supportedTestTypes["Rust"] = string(Rust)
 	}
```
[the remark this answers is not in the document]
MarkusZoppelt: I would like to keep it for future language support. It's a nice way to ensure that we always use the correct set of languages. We also call this wrapper function and removing that block looks confusing, imho..

integration-tests/node-typescript/nodets_test.go
MarkusZoppelt: It is the default, so removing this line is fine

e2e-tests/init/init_nodejs_test.go
MarkusZoppelt: Good catch, thanks! One of these tests was also expecting the init command to fail for `!prerelease`. Fixed.

MarkusZoppelt:
> Do you know why the E2E tests are failing?

Ah, I forgot to remove this e2e testcase which is expected to not fail / result in an error any more:
https://github.com/CodeIntelligenceTesting/cifuzz/blob/26d8c566ec7a186eff671f8e29918e5a2d6271f2/e2e-tests/init/init_nodejs_test.go#L13-L23

I removed the testcase (since it is no longer needed) and tests should pass now.

internal/config/types.go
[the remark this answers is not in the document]
mlsandnerCI: 👌

mlsandnerCI (approved):

jochil:
Maybe this is a good time to add JS/TS logos in the README too
image

PR #972, opened 2023-08-07
CLI-1175 Test `cifuzz container run` with Bazel

adombeck:
**Based on https://github.com/CodeIntelligenceTesting/cifuzz/pull/961, please review after that.**

### Motivation/Context

We want to support the container run command with Bazel and have a test for that.

### Description

Some changes were necessary to get `cifuzz container run` to work with Bazel. Please see the commit messages for explanations.

### How to use/reproduce

To run the new test:
```
go test -v ./integration-tests/bazel/... -run TestIntegration_Bazel/containerRun
```

PR #983, opened 2023-08-15
CLI-1167 Print bundle metadata

adombeck:
**Based on https://github.com/CodeIntelligenceTesting/cifuzz/pull/972, please review after that.**

### Motivation/Context

Requested by the backend team to support debugging and later on using information for displaying in UI.

### Description

This makes the `cifuzz execute` command print the bundle metadata in JSON format if the `--json` option is used. It's not printed by default to not clutter the output of the `cifuzz container run` command by default. When called by the backend, the `--json` option should be used anyway to receive parseable output. 

When `cifuzz container run` command is called with `--json`, it now passes that option on to `cifuzz execute`.

### How to use/reproduce

```
make build-container-image && cifuzz -C examples/cmake container run my_fuzz_test --interactive=false --no-notifications --docker-image cifuzz-test-container-run-other:latest --json
```


internal/container/container.go
```diff
@@ -34,6 +34,9 @@ func Create() (string, error) {
 	if viper.GetBool("verbose") {
 		containerConfig.Cmd = append(containerConfig.Cmd, "-v")
 	}
+	if printJSON {
```
mlsandnerCI, line 37: Would it make sense to use `viper.GetBool("print-json")` here instead of giving it as a parameter to the function? We do the same with verbose too and I'm wondering if there is a reason to handle both flags differently?

adombeck: In general we don't use viper throughout our code to access the config options. Instead, we use viper once to write the config options into an options struct and pass the values from that on to the functions which need them (which should make the code easier to understand, because we're not accessing global state everywhere). The one exception we have is `viper.GetBool("verbose")`, because we would have to pass the verbose option on to all functions which print debug messages. For that option we actually want to access global state.

adombeck:
>  would maybe squash the commits 

done

jochil:
> This makes the `cifuzz execute` command print the bundle metadata in JSON format if the `--json` option is used. It's not printed by default to not clutter the output of the `cifuzz container run` command by default.

Shouldn't the amount of information be the same, no matter of the format?

jochil:
> ### How to use/reproduce
> ```
> make build-container-image && cifuzz -C examples/cmake container run my_fuzz_test --interactive=false --no-notifications --docker-image cifuzz-test-container-run-other:latest --json
> ```

Unfortunately this does not work for me
image

adombeck:
> Shouldn't the amount of information be the same, no matter of the format?

I don't think the bundle metadata is useful information to the user when running `cifuzz container run`. If it's important to us that we don't show more information in JSON mode than in plain mode, we could hide it behind an additional flag, like `--print-bundle-metadata`.

adombeck:
You need to `export CIFUZZ_PRERELEASE=1` to use the `container` command.

jochil:
> **Based on #972, please review after that.**
> 
> ### Motivation/Context
> Requested by the backend team to support debugging and later on using information for displaying in UI.
> 
> ### Description
> This makes the `cifuzz execute` command print the bundle metadata in JSON format if the `--json` option is used. It's not printed by default to not clutter the output of the `cifuzz container run` command by default. When called by the backend, the `--json` option should be used anyway to receive parseable output.
> 
> When `cifuzz container run` command is called with `--json`, it now passes that option on to `cifuzz execute`.
> 
> ### How to use/reproduce
> ```
> make build-container-image && cifuzz -C examples/cmake container run my_fuzz_test --interactive=false --no-notifications --docker-image cifuzz-test-container-run-other:latest --json
> ```

Still an error... I think it would also be helpful to add `make install` to the command:
> ❌ Failed to build image from bundle my_fuzz_test.tar.gz: pull access denied for cifuzz-test-container-run-other, repository does not exist or may require 'docker login': denied: requested access to the resource is denied

jochil:
> > Shouldn't the amount of information be the same, no matter of the format?
> 
> I don't think the bundle metadata is useful information to the user when running `cifuzz container run`. If it's important to us that we don't show more information in JSON mode than in plain mode, we could hide it behind an additional flag, like `--print-bundle-metadata`.

I think another flag is not necessary... maybe put it behind the verbose flag. 
But my point is, that the selected format should not affect the amount of information... otherwise it just a hidden kind of verbosity level

adombeck:
> Still an error... 

You need to provide a docker image with cmake installed. The `cifuzz-test-container-run-other` image exists if you executed the cmake integration test locally since we added the container run test there. You can do that via:

```
go test ./integration-tests/other/... -run TestIntegration_Other/containerRun
```

adombeck:
> I think another flag is not necessary... maybe put it behind the verbose flag.

I don't think the verbose flag should affect the output of JSON messages on stdout, it should only affect the output of log messages on stderr.

> But my point is, that the selected format should not affect the amount of information... otherwise it just a hidden kind of verbosity level

I understand your point but which solution do you propose? I don't see another than an additional flag. Actually, I like the idea of putting it behind an additional flag, because the current behavior makes the output of `cifuzz container run --json` inconsistent with the one of `cifuzz run --json`, there is no explanation for what this first JSON message is about and (most importantly IMO), the message is not useful to users of `cifuzz container run`, it's only useful to the backend team when they execute `cifuzz execute` directly.

internal/container/container.go
mlsandnerCI: Thanks for the explanation 👍

adombeck:
I pushed a commit which adds a `--print-bundle-metadata` flag

jochil:
> > Still an error...
> 
> You need to provide a docker image with cmake installed. The `cifuzz-test-container-run-other` image exists if you executed the other build system integration test locally since we added the container run test there. You can do that via:
> 
> ```
> go test ./integration-tests/other/... -run TestIntegration_Other/containerRun
> ```

can you please update the Reproduce command in the PR description accordingly?

jochil:
> > I think another flag is not necessary... maybe put it behind the verbose flag.
> 
> I don't think the verbose flag should affect the output of JSON messages on stdout, it should only affect the output of log messages on stderr.
> 
> > But my point is, that the selected format should not affect the amount of information... otherwise it just a hidden kind of verbosity level
> 
> I understand your point but which solution do you propose? I don't see another than an additional flag. Actually, I like the idea of putting it behind an additional flag, because the current behavior makes the output of `cifuzz container run --json` inconsistent with the one of `cifuzz run --json`, there is no explanation for what this first JSON message is about and (most importantly IMO), the message is not useful to users of `cifuzz container run`, it's only useful to the backend team when they execute `cifuzz execute` directly.

I got your point with the JSON output and from my point of view it is fine having a "fixed" json output independent from the `--verbose` flag, too... just a different type of thinking about it, but as long we have the same model from now it is fine for me.
 
But I think we should be careful adding more and more flags for specific use cases... I would prefer just adding the bundle metadata output to the `--verbose` stderr output. If there are any complains we can still change it... but by just adding it we would also be in line what we discussed in the CS/Bug Guideline meeting on Wednesday. And as far as I know there is no actual need to have the `--json` output for `container run` and `run` in sync.

adombeck:
> But I think we should be careful adding more and more flags for specific use cases... 

Why? IIUC `cifuzz execute` is not meant to be called directly by the user, so I think we can add all the flags we want to make it work well both when called via `cifuzz container run` and when called by the backend worker.

>  but by just adding it we would also be in line what we discussed in the CS/Bug Guideline meeting on Wednesday

That would be the case if we expect printing the bundle metadata could help us debug a failed `cifuzz container run` command. That's not the reason why we're printing it though, the reason is that the backend team wants this output when calling `cifuzz execute`.

jochil:
> > But I think we should be careful adding more and more flags for specific use cases...
> 
> Why? IIUC `cifuzz execute` is not meant to be called directly by the user, so I think we can add all the flags we want to make it work well both when called via `cifuzz container run` and when called by the backend worker.

I think you are right... we could be a bit more flexible with `cifuzz execute` than with mainly user facing commands

> > but by just adding it we would also be in line what we discussed in the CS/Bug Guideline meeting on Wednesday
> 
> That would be the case if we expect printing the bundle metadata could help us debug a failed `cifuzz container run` command. That's not the reason why we're printing it though, the reason is that the backend team wants this output when calling `cifuzz execute`.

I think it can help us debugging as it is an easy way to see the settings without digging deeper into the created bundle... IIRC this is also the reason the backend team asked for it.
On the other hand... if we add this information to the verbose output when creating a bundle (see CLI-1211) it might be obsolete

jochil (approved):

PR #984, opened 2023-08-15
CLI-1169 Make `cifuzz execute` always create corpus in `/cifuzz/corpus`

adombeck:
### Motivation/Context

The worker should to be able to provide a corpus directory to the container. `cifuzz execute` should use existing corpus seeds from that directory and add new corpus seeds to that directory.

### Description

`cifuzz execute` already sets the `libfuzzer.RunnerOptions.GeneratedCorpusDir` option to `/cifuzz/corpus`: https://github.com/CodeIntelligenceTesting/cifuzz/blob/ffc3d9304b0a62d470f4845c981e9a88244635c5/internal/cmd/execute/execute.go#L207-L207

The only problem was that this option was ignored by `jazzer.Runner`. That's fixed by this PR.

Note that there is currently no test which checks that the corpus directory is used. Such a test would need to start the container created by `cifuzz container run` with a bind-mount mounted to `/cifuzz/corpus` in the container, so that it can check if files were created there. This would be a lot easier if the `cifuzz container run` command would just mount the existing corpus directory (`.cifuzz-corpus`) from the host to `/cifuzz/corpus` in the container. So I propose that we add this feature and then a test.

### How to use/reproduce

To have Jazzer create some files in the corpus, we first need to patch the fuzz test to not crash immediately:

```diff
diff --git a/examples/maven/src/main/java/com/example/ExploreMe.java b/examples/maven/src/main/java/com/example/ExploreMe.java
index 22923029..973794ba 100644
--- a/examples/maven/src/main/java/com/example/ExploreMe.java
+++ b/examples/maven/src/main/java/com/example/ExploreMe.java
@@ -14,11 +14,7 @@ public class ExploreMe {
                 if (b - a < 100000) {
                     // Trigger Remote Code Execution Bug
                     if (c.startsWith("@")) {
-                        String className = c.substring(1);
-                        try {
-                            Class.forName(className).newInstance();
-                        } catch (ClassNotFoundException | InstantiationException | IllegalAccessException ignored) {
-                        }
+                        System.out.println(c);
                     }
                 }
             }
```

Now we can start a container for the maven example:
```
make build-container-image && CIFUZZ_CONTAINER_SLEEP=1 CIFUZZ_PRERELEASE=1 cifuzz -C examples/maven container run com.example.FuzzTestCase::myFuzzTest --interactive=false --no-notifications
```

Attach to the started container by following the on-screen message.

In the container, execute the original command via `eval $CMD`. After 2-3 seconds, cancel the run via Ctrl+C. Now you can inspect the content of `/cifuzz/corpus`:

```
ls -l /cifuzz/corpus
```

On this branch, there should be files in that directory. On main, the directory is empty and the files are in `.cifuzz-corpus/com.example.FuzzTestCase/myFuzzTest/` instead.


MarkusZoppelt (approved):
Thank you for the very helpful description on how to test/reproduce this behavior!

PR #986, opened 2023-08-16
Use Java fuzz test listing tool for bundles

florianGla:
The new fuzz test listing tool for Java is used for bundles in order to add the correct fuzz tests. The previous regex based solution was not able to pick up the correct fuzz tests in all cases.

### Motivation/Context

Users reported that not all fuzz tests were added to the bundle

### How to use/reproduce

Create bundles for Java example projects.


internal/builder/builder.go
```diff
@@ -284,6 +290,54 @@ func (i *CIFuzzBuilder) BuildProcessWrapper() error {
...
+	}()
+
+	// Build tool in temp dir
+	tempDir, err := os.MkdirTemp("", "list-fuzz-tests-")
+	if err != nil {
+		return errors.WithStack(err)
+	}
+	defer fileutil.Cleanup(tempDir)
+
+	err = copy.Copy(filepath.Join(i.projectDir, "tools", "list-fuzz-tests"), tempDir)
```
adombeck, line 313: This error is not checked

internal/builder/builder.go
```diff
@@ -284,6 +290,54 @@ func (i *CIFuzzBuilder) BuildProcessWrapper() error {
...
+		return err
+	}
+	defer func() {
+		err = i.Unlock()
+		if err != nil {
+			log.Printf("error: %v", err)
+		}
+	}()
+
+	// Build tool in temp dir
```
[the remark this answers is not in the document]
florianGla: I did not want to pollute our project tree with build artifacts, but I'm open to change that.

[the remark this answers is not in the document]
adombeck: Unless they cause any issues IMO it's fine to have the build artifacts in the project tree (and have them ignored in the `.gitignore`, but that's already the case). We also have build artifacts in `build/` and `third-party/minijail` which don't cause any problems AFAICT.

internal/build/maven/maven.go
```diff
@@ -189,7 +189,17 @@ func GetTestDir(projectDir string) (string, error) {
 	}
 
 	log.Debugf("Found maven test source at: %s", project.Build.TestSourceDirectory)
-	return strings.TrimSpace(project.Build.TestSourceDirectory), nil
+	testDir := strings.TrimSpace(project.Build.TestSourceDirectory)
+
+	exists, err := fileutil.Exists(testDir)
+	if err != nil {
+		return "", errors.Wrapf(err, "Error checking if Maven test dir %s exists", testDir)
```
adombeck, line 196: Using `errors.Wrapf` adds a duplicate stack trace to the error. If you want to attach a message here, use `errors.WithMessagef` instead.

internal/bundler/jazzer_bundler.go
```diff
@@ -403,9 +393,18 @@ func (b *jazzerBundler) getTestDirs() ([]string, error) {
...
-		testDirs = append(testDirs, testDir)
+		if testDir != "" {
+			testDirs = append(testDirs, testDir)
+		}
 	} else {
-		testDirs = append(testDirs, filepath.Join(b.opts.ProjectDir, "src", "test"))
+		defaultTestDir := filepath.Join(b.opts.ProjectDir, "src", "test")
+		exists, err := fileutil.Exists(defaultTestDir)
+		if err != nil {
+			return nil, errors.Wrapf(err, "Error checking if default test dir %s exists", defaultTestDir)
```
adombeck, line 403: Same here regarding the duplicate stack trace.

internal/bundler/jazzer_bundler.go
```diff
@@ -424,9 +423,18 @@ func (b *jazzerBundler) getSourceDirs() ([]string, error) {
...
-		sourceDirs = append(sourceDirs, sourceDir)
+		if sourceDir != "" {
+			sourceDirs = append(sourceDirs, sourceDir)
+		}
 	} else {
-		sourceDirs = append(sourceDirs, filepath.Join(b.opts.ProjectDir, "src", "main"))
+		defaultSourceDir := filepath.Join(b.opts.ProjectDir, "src", "main")
+		exists, err := fileutil.Exists(defaultSourceDir)
+		if err != nil {
+			return nil, errors.Wrapf(err, "Error checking if default source dir %s exists", defaultSourceDir)
```
adombeck, line 433: Same here regarding the duplicate stack trace.

internal/build/maven/maven.go
```diff
@@ -203,5 +213,15 @@ func GetSourceDir(projectDir string) (string, error) {
 	}
 
 	log.Debugf("Found maven source at: %s", project.Build.SourceDirectory)
-	return strings.TrimSpace(project.Build.SourceDirectory), nil
+	sourceDir := strings.TrimSpace(project.Build.SourceDirectory)
+
+	exists, err := fileutil.Exists(sourceDir)
+	if err != nil {
+		return "", errors.Wrapf(err, "Error checking if Maven source dir %s exists", sourceDir)
```
adombeck, line 220: Same here regarding the duplicate stack trace.

internal/build/maven/maven.go
```diff
@@ -189,7 +189,17 @@ func GetTestDir(projectDir string) (string, error) {
 	}
 
 	log.Debugf("Found maven test source at: %s", project.Build.TestSourceDirectory)
```
adombeck, line 191: If we print a debug message `Found maven test source at: %s`, maybe we should also print a message below in the case that this directory doesn't exist and we ignore it therefore, like: `Ignoring maven test source directory %s: directory does not exist.

internal/bundler/jazzer_bundler.go
```diff
@@ -286,16 +283,14 @@ func (b *jazzerBundler) runBuild() ([]*build.Result, error) {
...
 			return nil, err
 		}
-		for i := range fuzzTests {
-			buildResult, err := builder.Build(fuzzTests[i], targetMethods[i])
-			if err != nil {
-				return nil, err
-			}
-			buildResults = append(buildResults, buildResult)
+
+		buildResult, err = builder.Build("", "")
```
adombeck, line 287: I think this could use some refactoring. AFAICT we don't have to pass the arguments `targetClass string, targetMethod string` to `builder.Build`, because it doesn't use them, it only sets them in the `buildResult` struct which the caller could do instead. Looking further, it seems like the `buildResult.Name` and `buildResult.TargetMethod` fields are not used anywhere in the Jazzer bundler. Could we delete them altogether?

internal/bundler/jazzer_bundler.go
```diff
@@ -503,3 +511,48 @@ func getUniqueArtifactName(dependency string, artifactsMap map[string]uint) stri
...
+
+	javaHome, err := runfiles.Finder.JavaHomePath()
+	if err != nil {
+		return nil, err
+	}
+	javaBin := filepath.Join(javaHome, "bin", "java")
+	if runtime.GOOS == "windows" {
+		javaBin = filepath.Join(javaHome, "bin", "java.exe")
+	}
+
```
adombeck, line 539: This duplicates the code from https://github.com/CodeIntelligenceTesting/cifuzz/blob/CLI-1180-list-fuzz-tests-in-bundle/pkg/runner/jazzer/jazzer_runner.go#L63-L71 and I think https://github.com/CodeIntelligenceTesting/cifuzz/blob/CLI-1180-list-fuzz-tests-in-bundle/pkg/dependencies/version.go#L158-L162 should also do the same but it doesn't handle the Windows case. Maybe we could move this code to the currently unused `runfiles.Finder.JavaPath()` and call that?

internal/bundler/jazzer_bundler.go
```diff
@@ -503,3 +511,48 @@ func getUniqueArtifactName(dependency string, artifactsMap map[string]uint) stri
...
+	log.Debugf("Command: %s", cmd.String())
+	err = cmd.Run()
+	if err != nil {
+		return nil, cmdutils.WrapExecError(errors.WithStack(err), cmd)
+	}
+
+	output, err := io.ReadAll(buffer)
+	if err != nil {
+		return nil, errors.WithStack(err)
+	}
```
adombeck, line 553: This can be simplified:

```suggestion
cmd := exec.Command(javaBin, args...)
	cmd.Stderr = os.Stderr
	log.Debugf("Command: %s", cmd.String())
	output, err := cmd.Output()
	if err != nil {
		return nil, cmdutils.WrapExecError(errors.WithStack(err), cmd)
	}
```

internal/builder/builder.go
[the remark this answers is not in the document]
florianGla: Thank you, I will change that.

internal/bundler/jazzer_bundler.go
florianGla: Yes, I did not want to refactor it further, because we wanted to do a refactoring session with the whole team to improve the `buildResult` usage. But I can do it in this PR if you think that makes more sense. `buildResult.Name` and `buildResult.TargetMethod` are not used in the Jazzer bundler, but those are needed in other commands. We should think about a better approach.

internal/bundler/jazzer_bundler.go
florianGla: Great idea! I will do that

internal/bundler/jazzer_bundler.go
adombeck: > buildResult.Name and buildResult.TargetMethod are not used in the Jazzer bundler, but those are needed in other commands.

I can't find any place where they are used, can you point me to it?

florianGla: `buildResult.TargetMethod` is indeed not needed anymore. I will remove it. `buildResult.Name` is used in libfuzzer_bundler.go (lines 502, 555, 616)

florianGla: `buildResult.TargetMethod` is removed

adombeck: Please remove the now unused `targetMethod` argument as well.

adombeck: > buildResult.Name is used in libfuzzer_bundler.go (lines 502, 555, 616)

If it's only used in `libfuzzer_bundler.go`, there is no reason to set it in `jazzer_bundler.go` IMO. To make it more clear which fields need to be set where, I propose we refactor this so the builders which are called by `libfuzzer_bundler.go` and `jazzer_bundler.go` return different structs and extract the fields which are needed by both (and which are needed by `runCmd.runFuzzTest` in the `run` package) into a third struct which is embedded by the other two:

```
// BuildResult contains fields which are needed to run the fuzz test
type BuildResult struct {
	// A name which uniquely identifies the fuzz test and is a valid path
	Name string
	// Canonical path of the fuzz test executable
	Executable string
	// Canonical path of the fuzz test's generated corpus directory
	GeneratedCorpus string
	// Canonical path of the fuzz test's default seed corpus directory
	SeedCorpus string
	// Canonical path of the build directory
	BuildDir string
	// The canonical paths of the fuzz test's runtime dependencies
	RuntimeDeps []string
}

// CBuildResult contains the fields needed to run or bundle a C project which has been built
type CBuildResult struct {
	BuildResult
	// A name which uniquely identifies the fuzz test and is a valid path
	Name string
	// The sanitizers with which the fuzz test was built
	Sanitizers []string
	// Canonical path of the directory to which source file paths should
	// be made relative
	ProjectDir string
}

// JavaBuildResult contains the fields needed to run or bundle a Java (or other JVM language) project which has been built
type JavaBuildResult struct {
	BuildResult
}
```

What do you think?
```

adombeck (approved):

PR #988, opened 2023-08-16
feat(examples): extended maven & gradle project

mlsandnerCI:
### Motivation/Context
We noticed that our example projects are sometimes too simple and don't catch problems (like with multi module setups).

### Description

New maven & gradle projects with multi module setups, multiple fuzz tests and other edge cases.


jochil:
> I thought the plan for [CLI-1171](https://code-intelligence.atlassian.net/browse/CLI-1171) was to find existing, actual projects which we can run some tests on.

not at all... the goal is to have projects that are covering more complex setups and use cases, not necessarily real projects

test/projects/gradle/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+### Remote Code Execution
+
+```
+cifuzz run -C app com.example.TestCases::remoteCodeFuzzTest
+```
+
+```
+cifuzz run -C testsuite com.example.RCEFuzzTestCase::fuzzTestTrigger
+```
```
jochil, line 21: I guess this can be one code block

test/projects/gradle/testsuite/build.gradle.kts
```diff
@@ -0,0 +1,18 @@
+plugins {
+    id("java-library")
+    id("com.code-intelligence.cifuzz") version "1.3.0"
```
jochil, line 3: we can use version 1.4.0 of the plugin already

jochil (approved):

PR #991, opened 2023-08-17
fix(errors): handle all error types correctly in root 

mlsandnerCI:
### Motivation/Context

Usage errors were falsely printed with a support message indicating to run the command in verbose mode.

### Description

Refactored the if-clauses in the root command handling the errors to a switch/case to make sure that errors only get printed once. 

In case of the usage error mentioned in the ticket, the error did not get printed at all in the `IncorrectUsageError` handling, only the help message and then it was handled again as "not a `SilentError`" where the error message with the support message + stack trace was printed.

Handling of `ExecError` and `APIError` also got moved to the root command, the stack trace will be printed in verbose mode. 

Edit: The support message got removed.

#### Some error messages for reference
Old `ExecError` message in verbose:
execErr_old

New `ExecError` message:
Bildschirmfoto 2023-08-17 um 14 26 34
Bildschirmfoto 2023-08-17 um 14 21 41 
Edit: Stacktrace is now gray and there is no support message anymore.

`IncorrectUsageError` message, wrong arguments:
usageErr_new

`IncorrectUsageError` message, not a cifuzz project:
Bildschirmfoto 2023-08-17 um 14 45 51


adombeck:
> New ExecError message:
> image

I find the "More information can be acquired running the command in verbose mode (--verbose)" confusing here. It as if the user could get some useful information when re-running with `--verbose`, but it will only show them the stack trace, which is not useful information for the user. The reason we print a stack trace in verbose mode is to make things easier for us developers, not because it's useful information for the user.

adombeck:
This looks good!

internal/cmd/root/root.go
```diff
@@ -129,30 +130,44 @@ func Execute() {
...
 	var cmd *cobra.Command
 	if cmd, err = rootCmd.ExecuteC(); err != nil {
-		// We only want to print the usage message if an ErrIncorrectUsage
-		// was returned or it's an error produced by cobra which was
-		// caused by incorrect usage
-		var usageErr *cmdutils.IncorrectUsageError
-		if errors.As(err, &usageErr) ||
-			strings.HasPrefix(err.Error(), "unknown command") ||
-			regexp.MustCompile(`(accepts|requires).*arg\(s\)`).MatchString(err.Error()) {
+		// Configure error formatting because we can't use the log package here
```
adombeck, line 133: Why can't we use the log package here?

internal/cmd/root/root.go
```diff
@@ -129,30 +130,44 @@ func Execute() {
...
-			}
+		// Error types we want to handle differently
+		var usageErr *cmdutils.IncorrectUsageError
+		var couldBeSandboxError *cmdutils.CouldBeSandboxError
+		var signalErr *cmdutils.SignalError
+		var execErr *cmdutils.ExecError
+		var apiErr *api.APIError
+		var silentErr *cmdutils.SilentError
 
+		switch {
```
adombeck, line 151: Using `switch` here instead of multiple `if` statements has the effect that only the first `case` which matches is executed. The error types are not mutually exclusive though, e.g. it's possible that an error is wrapped as both an `ExecError` and a `CouldBeSandboxError`. Either we add `fallthrough` statements to each case or we revert this back to use `if` statements.

adombeck: We also have to make sure that the error is not printed twice in these cases.

mlsandnerCI: I did not know that this case exists, thanks for pointing it out. I will adjust the code.

internal/cmd/root/root.go
mlsandnerCI: `log.Error()` would always print a stack trace in verbose mode and not the stack trace of the root error but of the last one that wrapped it. 
And we didn't use it before I refactored it, I'm assuming there is another reason why we decided not to use it here.

internal/cmd/root/root.go
mlsandnerCI: I updated the code but I'm unsure if this is okay now. Could you recheck?

internal/cmd/root/root.go
```diff
@@ -129,36 +130,45 @@ func Execute() {
...
-			// and the following message
-			if !strings.HasSuffix(err.Error(), "\n") {
-				_, _ = fmt.Fprintln(cmd.ErrOrStderr())
-			}
-			msg := `Note: If you don't expect this fuzz test to do any harm to the system
+			msg := `
+Note: This error could have occurred due to sandboxing.
+If you don't expect this fuzz test to do any harm to the system
 accidentally (like overwriting files), you might want to try
 running it without sandboxing:
```
adombeck, line 173: We want to print this note after the error message, so this if block should be moved below the one which prints non-silent errors.

It (accidentally) worked as expected before, because we handled all errors which we wrapped as `CouldBeSandboxError` on an upper layer and returned a `SilentError`.

adombeck: While testing this, I noticed that we didn't print the note for the `CouldBeSandboxError` returned here: https://github.com/CodeIntelligenceTesting/cifuzz/blob/mlsandnerCI%2FCLI-1189-error-no-stacktrace/internal/cmd/coverage/llvm/llvm.go#L89-L89

I fixed that in a commit which I pushed to this branch, @mlsandnerCI please review that commit.

internal/cmd/root/root.go
```diff
@@ -167,55 +177,51 @@ running it without sandboxing:
...
-		if !errors.As(err, &silentErr) {
-			icon := "❌ "
-			style := pterm.Style{pterm.Bold, pterm.FgRed}
-			if log.PlainStyle() {
-				icon = ""
-				style = pterm.Style{}
-			}
+		if errors.As(err, &execErr) ||
+			errors.As(err, &apiErr) ||
+			!errors.As(err, &silentErr) {
```
adombeck, line 194: This has the effect that an `ExecError` or `ApiError` which was wrapped as `SilentError` is still printed. We don't want to print any `SilentError` though, so lets remove the check for `ExecError` and `ApiError` here.

adombeck (changes requested):

internal/cmd/root/root.go
adombeck: > log.Error() would always print a stack trace in verbose mode and not the stack trace of the root error but of the last one that wrapped it.

Ah that's because of the `errors.As(errors.Cause(err), &st)`. Took me a while to understand, a comment would have been helpful there. But I think this behavior is a problem actually: If the original error doesn't have a stack trace, we don't print a stack trace even if `err` does have one. That's the case when we return an error that's wrapped via `errors.WithStack` or `errors.Wrap` but not as any other error (like `ExecError`). To test this, you can for example make the creation of the build dir fail:
```
rm -rf examples/cmake/.cifuzz-build
touch examples/cmake/.cifuzz-build
make install && cifuzz -C examples/cmake run my_fuzz_test -v
```

internal/cmd/root/root.go
```diff
@@ -129,36 +130,45 @@ func Execute() {
...
 			// Make cmd.Help() print to stderr
 			cmd.SetOut(cmd.ErrOrStderr())
+
 			// Print the usage message of the command. We use cmd.Help()
 			// here instead of cmd.UsageString() because the latter
 			// doesn't include the long description.
 			_ = cmd.Help()
+
+			// Add "Invalid Usage" in front of error message
+			err = errors.Wrap(err, "Invalid usage")
```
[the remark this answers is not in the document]
mlsandnerCI: I would argue that using cifuzz outside of an initialized project is an incorrect usage of cifuzz from the user. I added the "Invalid Usage" message to show the user that this is something they can fix, in this case either changing the directory or running `cifuzz init`. Homebrew does this too and I think it provides value for the user. 

And I do believe that it is easier to read the error message if the "use cifuzz init" solution is separately. The first error message is very long plus it seems more logical to first print the error and then the solution.  

I would prefer to keep it like this. If however this is something you strongly feel about and because this is not a right vs wrong decision but a style decision, I would suggest a voting from the team and see what everyone else thinks?

internal/cmd/root/root.go
mlsandnerCI: I wrote a comment but I can elaborate it if needed. 
```
// If running in verbose mode, check if the root error has a
// stacktrace attached and print it (except for usage errors)
```

This was a deliberate decision in the last PR that only the original stack trace gets printed. For example the stack trace pointing to a go routine which returned an error which actually happened somewhere else is useless IMO. If we need a stack trace at the point where the error occurs, we should add it right there.

adombeck: The problem is that if the original error originates in the standard library or a third-party package, it does not have a stack trace. That's why we call `errors.WithStack` or `errors.Wrap` when we get an error from such a package. `errors.Cause(err)` returns the original error from the third-party package, so it strips the stack trace we add and doesn't print any stack trace.

IMO the correct solution is to print the stack trace of `err` instead and ensure that we never add more than one stack trace to an error.

mlsandnerCI: That is true with the librarys/external packages. We can try and see if your solution works out 👍 Wrapping always adds a stack trace that's why I added this part to make sure we print the correct one. If we have a situation where the error does not print the correct stack trace, we can always come back to this and see how we can adjust this.

adombeck: > that's why I added this part to make sure we print the correct one.

Note that the only reason that most of our errors were still printed with a stack trace after that part was implemented is that we wrap most errors with one of our own error types, like `IncorrectUsageError` or `ExecError`, which don't implement the `Cause` method, so `errors.Cause` returns that error instead of the original.

internal/cmd/root/root.go
```diff
@@ -167,55 +174,49 @@ running it without sandboxing:
...
-					_, _ = fmt.Fprint(cmd.ErrOrStderr(), style.Sprintf("%s%s\n%s", icon, err, supportMsg))
-				}
-			} else {
-				// Catch cases where we either did not add any stacktrace/wrapped the error
-				// or the error does not implement the interface for the stacktracer e.g. os.ErrExist
-				_, _ = fmt.Fprint(cmd.ErrOrStderr(), style.Sprintf("\n%s%v\n", icon, err))
+			if viper.GetBool("verbose") &&
+				errors.As(err, &st) &&
+				!errors.As(err, &usageErr) {
+				_, _ = fmt.Fprint(cmd.ErrOrStderr(), stackTraceStyle.Sprintf("%+v", st.StackTrace()))
```
adombeck, line 208: The error + stack trace we print here look great now! I think we should use the same style in the `log.Error` package. And if we move the respective code there, we might be able to just call `log.Error` here. What do you think?

mlsandnerCI: Sounds great 👍 Should I add it in this PR?

internal/cmd/root/root.go
mlsandnerCI: Can a signal error also be a sandbox error? Because then I would need to remove the os.Exit from that if-clause so that we can still print this error message if it is checked after the non-silent error check 🤔

adombeck: No, we return a `SignalError` when cifuzz catches a terminating signal and a `CouldBeSandboxError` when a command that's executed in Minijail fails. But it's good that you considered that case 👌

internal/cmd/root/root.go
mlsandnerCI: The only thing I am struggling now is the IncorrectUsage error. I would like to use log.Error but I don't want to add a check in the function for this special error so we don't print a stack trace and as far as I know there is now way to remove a stack trace from an error. 

Either we never add a stack trace to this kind of error which means I would have to check everywhere that we always initialize the error correctly. This would fit in our guidelines to only add stack traces where we need them but this approach could be error prone if we accidentally use the wrong function and add a stack trace. 

We could also have another function like `SimpleError()` that only prints the error message itself and use that but I don't think that is really an optimal solution. What do you think?

adombeck: Ah yes I see the problem. I think an additional function to print an error without the stack trace makes sense, something like this:

```diff
diff --git a/pkg/log/log.go b/pkg/log/log.go
index 728c0077..e0ba0d01 100644
--- a/pkg/log/log.go
+++ b/pkg/log/log.go
@@ -133,13 +133,17 @@ func Errorf(err error, format string, a ...any) {
 	Error(err, fmt.Sprintf(format, a...))
 }
 
-func Error(err error, a ...any) {
+func printError(err error, withStack bool, a ...any) {
 	// If no message is provided, print the message of the error
 	if len(a) == 0 {
 		a = []any{err.Error()}
 	}
 	log(pterm.Style{pterm.Bold, pterm.FgRed}, "❌ ", a...)
 
+	if !withStack {
+		return
+	}
+
 	type stackTracer interface {
 		StackTrace() errors.StackTrace
 	}
@@ -149,6 +153,14 @@ func Error(err error, a ...any) {
 	}
 }
 
+func Error(err error, a ...any) {
+	printError(err, true, a...)
+}
+
+func ErrorWithoutStack(err error, a ...any) {
+	printError(err, false, a...)
+}
+
 // Infof outputs a regular user message without any highlighting
 func Infof(format string, a ...any) {
 	Info(fmt.Sprintf(format, a...))
```

mlsandnerCI: Cool, thanks for the quick feedback!

pkg/log/log.go
```diff
@@ -128,24 +128,43 @@ func Note(a ...any) {
...
+
 	type stackTracer interface {
 		StackTrace() errors.StackTrace
 	}
 	var st stackTracer
-	if errors.As(err, &st) {
-		Debugf("%+v", st)
+	if viper.GetBool("verbose") &&
+		errors.As(err, &st) {
+		Info(fmt.Sprintf("%+v", st.StackTrace()))
```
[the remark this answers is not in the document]
mlsandnerCI: I already gave up hope on this because I could not find a way to make this work. I always thought I would have to remove the suffix from the error message which just resulted in weird code, not the prefix from the stack trace. Thanks so much!

adombeck:
> But we do want it to be printed, right? I would suggest wrapping the error in the provided message before logging it 🤔

The problem I see with automatically wrapping the error in the provided message in `log.Error` / `log.Errorf` is that the caller can't decide the format of that wrapping. For example, if the provided message is long, we probably want to show the error message on a separate line. If the provided message is short, we want to show it in the format `%s: %s`. Currently, the caller can decide the format, which provides the needed flexibility.

I suggest we keep the current behavior and explain in the function's doc comment that the error message will not be printed unless `err` is the only argument. This should also require fewer adaptions of existing calls to `log.Errorf`, because in most of those we already include `err.Error` in the provided message.

This has consequences for the `ErrorWithoutStack` function: It doesn't make sense to provide it an `err` argument if neither the message nor the stack trace of it is printed. I suggest we rename it and remove the `err` argument.

I already implemented my suggestions and adapted existing calls to `log.Error` and `log.Errorf` to test the feasibility. I'm pushing two commits, please take a look and tell me what you think.

internal/cmd/root/root.go
```diff
@@ -129,36 +128,53 @@ func Execute() {
...
 			// Make cmd.Help() print to stderr
 			cmd.SetOut(cmd.ErrOrStderr())
+
 			// Print the usage message of the command. We use cmd.Help()
 			// here instead of cmd.UsageString() because the latter
 			// doesn't include the long description.
 			_ = cmd.Help()
+
+			// Add "Invalid Usage" in front of error message
+			err = errors.Wrap(err, "Invalid usage")
```
adombeck, line 149: We don't want to attach a stack trace here
```suggestion
			err = errors.WithMessage(err, "Invalid usage")
```

adombeck (approved):

PR #993, opened 2023-08-21
style: use go-idiomatic package naming and structure

MarkusZoppelt:
### Motivation/Context

The `e2e` package and tests had a package structure that was not in line with the rest of our naming and project structure.

### Description

This PR renames the packages and directories to be more go-idiomatic, e.g., 

`e2e-tests/container-remote-run_test.go` -> `e2e/tests/container/remoterun_test.go`

etc.

*I also removed the `e2e-tests` entries from the buckets and let the rebalancer add them in its next PR cycle*

### How to use/reproduce

`make test/e2e`


e2e/tests/container/container_test.go
```diff
@@ -1,9 +1,9 @@
-package e2e
+package container_test
```
mlsandnerCI, line 1: Since these packages are in a `tests` directory, I don't think they need to have a the suffix `_test`. Or what do you think would be the benefit of this?

MarkusZoppelt: Hm, I wondered this too. Although I am not entirely sure.
I found some contradictory sources online.

E.g., look at this package name, from the Go sources itself:
https://github.com/golang/go/blob/e9188d1d18040bc6cb46065b2474664b8728a6df/src/fmt/stringer_test.go#L5

They are using the `_test` suffix to make the tests explicitly for that package under test.

Furthermore, [this StackOverflow answer](https://stackoverflow.com/questions/19998250/proper-package-naming-for-testing-with-the-go-language/31443271#31443271) describes packages with the `_test` suffix as **black-box tests**. This makes sense in this case, since our `e2e` tests only use exported functions and commands during test.

mlsandnerCI: Alright, if it is convention then let's keep it 👍

mlsandnerCI (approved):

PR #995, opened 2023-08-22
feat(finding): Rename original input file to contain finding name

fmeum:
### Description

By renaming the original fuzzer-produced input file from `crash-1234...` to `funky_kangaroo-1234...`, it becomes easier for users to match a failing IDE regression test case to a cifuzz finding.

### How to use/reproduce

In `examples/maven`, run `cifuzz run com.example.FuzzTestCase::myFuzzTest` and observe that an input file referencing the finding name is added.


fmeum:
I wasn't sure about how to add test coverage for this new behavior. If you have suggestions, please let me know.

florianGla (approved):
This feature does not make sense for C/C++ projects. In that case the input file stored in a temp dir gets renamed, which is not really a problem. I'm not sure if it makes sense to exclude C/C++ projects.

fmeum:
> This feature does not make sense for C/C++ projects. In that case the input file stored in a temp dir gets renamed, which is not really a problem. I'm not sure if it makes sense to exclude C/C++ projects.

I thought about that as well and ultimately came to the conclusion that renaming a temporary file is simpler than making a decision based on language. But I am open to change this.

PR #997, opened 2023-08-22
fix(cmd): check connection error after getting auth status

florianGla:
If it is not possible to check the authentication status due to a missing internet connection, CI Fuzz crashes with an error message. We should print a warning informing the user about the missing authentication with CI Sense, but continue with the command if possible. 

### Motivation/Context

Otherwise some cifuzz commands cannot be used offline.

### How to use/reproduce

Turn off your internet connection and either run `cifuzz run` or `cifuzz finding` or `cifuzz container run`.


internal/cmd/remoterun/remoterun.go
```diff
@@ -204,10 +203,14 @@ variable or by running 'cifuzz login' first.
 func (c *runRemoteCmd) run() error {
 	authenticated, err := auth.GetAuthStatus(c.apiClient.Server)
 	if err != nil {
-		return err
+		var connErr *api.ConnectionError
+		if !errors.As(err, &connErr) {
+			return err
+		}
```
MarkusZoppelt, line 209: Should we add a debug output, that there was a connection error?
And not only here but every time we run into one?

MarkusZoppelt (approved):

internal/cmd/remoterun/remoterun.go
florianGla: Great idea!

PR #1005, opened 2023-08-25
Skip failing UBSan test on Windows

adombeck:
### Motivation/Context

Since yesterday (2023-08-24), the test keeps failing in our CI because symbols needed by clang_rt.ubsan_standalone-x86_64.lib can't be found. This seems to be a known issue with newer Visual Studio versions:  https://github.com/microsoft/STL/issues/3568  

I suspect that the Visual Studio installed on the GitHub runner has been updated and that we can't do anything but wait until the issue was fixed.


adombeck:
I fixed two cases where `errors.Wrap` was called on errors which already had a stack trace. I started to look through some more occasions of `errors.Wrap` and found that there are more cases, for example:
* https://github.com/CodeIntelligenceTesting/cifuzz/blob/2f3a033c7245a63fe4922ad2dd9af99e5ed98ef0/internal/api/api.go#L332-L332
* https://github.com/CodeIntelligenceTesting/cifuzz/blob/2f3a033c7245a63fe4922ad2dd9af99e5ed98ef0/internal/build/bazel/bazel.go#L346-L346
* https://github.com/CodeIntelligenceTesting/cifuzz/blob/2f3a033c7245a63fe4922ad2dd9af99e5ed98ef0/internal/build/bazel/bazel.go#L387-L387

@mlsandnerCI I think most of those have been added by 3652838a3131dd6bec9ab6c8eb4d1d2536609ccb. We must have missed those when we merged the PR 😬 
Could you maybe go through the `errors.Wrap` occurrences and check that they don't add a duplicate stack trace?

jochil:
The failing jobs are definitly using the "broken" windows 2022 runner image
image

While the successful ones are using an old one
image

I think they are rolling it out incrementally

adombeck:
This PR now only contains the commits which skip the failing tests. I moved the other commits to https://github.com/CodeIntelligenceTesting/cifuzz/pull/1008.

jochil (approved):

adombeck:
@jochil: The compability check in the "build installer + upload as artifact (Windows)" pipeline is also broken because of the same issue, I pushed another commit which also disables that.

.github/workflows/workflow_installer.yml
```diff
@@ -34,7 +34,8 @@ jobs:
         run: make version=${{ inputs.version }} installer/darwin-arm64
 
       - name: Check Compatibility
-        if: ${{ inputs.compatibility_check == 'true' }}
+        # TODO: Enable on Windows again once https://github.com/microsoft/STL/issues/3568 is fixed
+        if: ${{ inputs.compatibility_check == 'true' && !contains(inputs.os, 'windows')" }}
```
jochil, line 38: This is causing an syntax error on the pipeline
image

I think the last `"` is unnecessary

adombeck: fixed

PR #1009, opened 2023-08-28
CLI-1237 Fix pterm signal handling

adombeck:
### Motivation/Context

An issue in pterm (https://github.com/pterm/pterm/issues/562) prevents the default behavior of SIGINT and SIGTERM when pterm is imported. In effect, Ctrl-C and `kill` don’t terminate cifuzz unless it currently registered its own handlers for those signals (which is the case when running a fuzzer).

### Description

This PR configures `go.mod` to use a pterm fork which fixes https://github.com/pterm/pterm/issues/562 by re-raising the handled signals.

### How to use/reproduce

To reproduce the issue on main, run cifuzz and immediately press Ctrl-C. The command will keep on running:
```
cifuzz -C examples/cmake run my_fuzz_test --interactive=false --no-notifications
# Press Ctrl-C
```

On this branch, pressing Ctrl-C will abort the command:
```
make install
cifuzz -C examples/cmake run my_fuzz_test --interactive=false --no-notifications
# Press Ctrl-C
```


MarkusZoppelt (approved):
Nice! Do you also plan on opening a PR upstream?

adombeck:
> Nice! Do you also plan on opening a PR upstream?

I proposed a better but more complex solution on the ticket and are waiting for feedback from the maintainers before I start working on it.

jochil:
I guess we should postpone this PR then https://github.com/CodeIntelligenceTesting/cifuzz/pull/1010, right?

adombeck:
> I guess we should postpone this PR then #1010, right?

Hah, I just wrote on that PR that I think we should close it 😄

PR #1011, opened 2023-08-29
Improve verbose output for bundling

mlsandnerCI:
### Motivation/Context

To improve the debugging process for CS and us, we discussed what information could be important that we not yet display in verbose mode especially in the context if we have to debug a failed bundle command but do not have access to the bundle.

### Description

- Printing archive content with size and file mode
- Printing content of bundle.yaml
- Including a "Creating bundle..." info message and refactoring the exisiting "Building for x..." info messages to not be prefixed with a new line for consistency because we do not add them in the build log file and it broke the visual flow of the debug messages. Since the info messages do not have an icon as our debug messages, they still stand out without any additional new line.

### How to use/reproduce

Creating a bundle in verbose mode will print the new information on the console.


internal/bundler/bundler.go
```diff
@@ -178,6 +186,9 @@ func (b *Bundler) createMetadataFileInArchive(fuzzers []*archive.Fuzzer, archive
 		return err
 	}
 
+	// Print bundle.yaml content for debugging purposes
+	log.Debugf("bundle.yaml\n%s", metadataYamlContent)
```
adombeck, line 190:
```suggestion
	log.Debugf("Content of bundle.yaml:\n%s", metadataYamlContent)
```

internal/bundler/bundler.go
```diff
@@ -94,6 +94,14 @@ func (b *Bundler) Bundle() (string, error) {
 		}
 	}
 
+	// List contents of archive in verbose mode for easier debugging
+	// when we do not have access to the bundle itself
+	var content []string
+	for _, h := range archiveWriter.Headers() {
+		content = append(content, fmt.Sprintf("%s %12d %s", h.FileInfo().Mode().String(), h.Size, h.Name))
+	}
+	log.Debugf("Archive content:\n%s", strings.Join(content, "\n"))
```
adombeck, line 103:
```suggestion
	log.Debugf("Content of bundle %s:\n%s", bundle.Name(), strings.Join(content, "\n"))
```

adombeck, line 103: We can use the builtin `tabwriter` package here to ensure that the width of the columns fits the string lengths:
```suggestion
	w := tabwriter.NewWriter(os.Stderr, 0, 0, 1, ' ', tabwriter.AlignRight)
	for _, h := range archiveWriter.Headers() {
		_, err := fmt.Fprintf(w, "%s\t%d\t %s\n", h.FileInfo().Mode().String(), h.Size, h.Name)
		if err != nil {
			return "", errors.WithStack(err)
		}
	}
	log.Debugf("Content of bundle %s:", bundle.Name())
	err = w.Flush()
	if err != nil {
		return "", errors.WithStack(err)
	}
```

internal/bundler/jazzer_bundler.go
```diff
@@ -54,6 +54,8 @@ func (b *jazzerBundler) bundle() ([]*archive.Fuzzer, error) {
 		return nil, err
 	}
 
+	log.Infof("Creating bundle...")
```
adombeck, line 57:
```suggestion
	log.Info("Creating bundle...")
```

internal/bundler/libfuzzer_bundler.go
```diff
@@ -70,6 +69,8 @@ func (b *libfuzzerBundler) bundle() ([]*archive.Fuzzer, error) {
 		return nil, err
 	}
 
+	log.Infof("Creating bundle...")
```
adombeck, line 72:
```suggestion
	log.Info("Creating bundle...")
```

adombeck (approved):

PR #1013, opened 2023-08-29
Remove duplicate stack traces

mlsandnerCI:
#991 should be merged first

### Motivation/Context

We are in the process of optimizing our internal error handling. The first changes involved wrapping errors instead of logging and adding stack traces if we use external packages. When logging the error, only the stack trace of the original error was printed. We decided against this specific approach so we now have to check that we only add a stack trace where we actually need it -> with external packages or custom created errors.

### Description
- replaced `errors.Wrap()` with `errors.WithMessage()` where we do not need a new stack trace 
- removed duplicate `errors.WithStack()` calls

In the process I found an error in the create command where the wrong error was returned and found a few errors that could be simplified.


pkg/stubs/stubs.go
```diff
@@ -51,7 +51,7 @@ func Create(path string, testType config.FuzzTestType) error {
 			}
 			fileNameExtension, found := config.TestTypeFileNameExtension(testType)
 			if !found {
-				return errors.WithStack(err)
+				return errors.Errorf("unknown file extension %s", testType)
```
adombeck, line 54: `testType` is not a file extension. Also, I think not finding a file extension here warrants a panic because it's clearly a programming error.

```suggestion
				panic(fmt.Sprintf("no file name extension found for test type %s", testType))
```

internal/cmd/run/run.go
```diff
@@ -912,7 +912,7 @@ func ExecuteRunner(runner Runner) error {
 	// Routines.Wait() returns an error created by us so it already has a
 	// stack trace and we don't want to add another one here
 	// nolint: wrapcheck
-	return err
+	return errors.WithMessage(err, "Failed to execute runner")
```
adombeck, line 915: IMO this message has more potential to be misleading than to be helpful, because it reads as if we failed to execute the fuzzer which is not necessarily the case, there are many other things which can result in `Runner.Run` returning an error.


```suggestion
	return err
```

adombeck (approved):

PR #1014, opened 2023-08-29
fix(remoterun): don't call auth.EnsureValidToken twice

MarkusZoppelt:
### Motivation/Context

The message “You are authenticated” appears twice when starting a remote run.

### Description

This PR removes an unecessary second call of `EnsureValidToken` since it is already called inside of `GetAuthStatus()` which itself gets called in the beginning of the `remote-run` command.
This caused a bug where some messages were printed twice, e.g., "You are authenticated."

### How to use/reproduce

Start a remote run and see that the message "You are authenticated" gets printed only once.


florianGla (approved):

adombeck:
I'm currently writing a comment requesting more changes

MarkusZoppelt:
> 1. We first call `auth.GetAuthStatus` and abort if it returns `false`. Then we check if `token == ""` and try to read it interactively. However, if `token == ""` we never get there because `auth.GetAuthStatus` returns `false` in that case.

I added a commit with improved behavior.

> 2. Why do we call `auth.GetAuthStatus` with `c.apiClient.Server` but `auth.GetToken` with `c.opts.Server`?

Both should be set correctly. I changed it to use `c.opts.Server` both times.

adombeck:
I think the case that `auth.GetAuthStatus` fails with an `api.ConnectionError` is not handled correctly. It will print a log message `Connection error: ...` (if --verbose is used) and then ask the user for an API token even if one was already provided.

MarkusZoppelt:
> I think the case that `auth.GetAuthStatus` fails with an `api.ConnectionError` is not handled correctly. It will print a log message `Connection error: ...` (if --verbose is used) and then ask the user for an API token even if one was already provided.

I agree. Because this is the `remote-run` cmd, we should return in case a (connection) error happens and not just log the error and continue. Added a commit with that change.

adombeck (approved):

PR #1017, opened 2023-08-30
Add source file location to jazzer stack traces

florianGla:
This PR uses a source map to convert the location of jazzer findings into file paths.

Consider our maven example project.
Previous finding location: com.example.ExploreMe:19
New finding location: src/main/java/com/example/ExploreMe.java:19

### Motivation/Context

This is needed to be able to map a finding to a location in the source code.

### How to use/reproduce

Start a fuzzing run for a java project.


pkg/parser/libfuzzer/stacktrace/stacktrace.go
```diff
@@ -265,3 +294,28 @@ func (p *parser) sourceLocationFromLine(line string) (*StackFrame, error) {
...
 }
+
+func (p *parser) setJavaSourceFilePath(frame *StackFrame) {
+	removeLastPart := func(javaName string) string {
+		sepIndex := strings.LastIndex(javaName, ".")
+		if sepIndex > 0 {
+			return javaName[0:sepIndex]
+		}
+		return ""
+	}
```
jochil, line 308: I would prefer moving this to a regular function and write a short test for it

jochil (commented):
I have the feeling that this change can use some more (unit) tests... there are a lot of changes to the code but little new tests

In general the solution with the `SourceFiles` field in the parser makes sense for me

internal/build/java/java.go
```diff
@@ -0,0 +1,55 @@
...
+			return nil, err
+		}
+		if sourceDir != "" {
+			sourceDirs = append(sourceDirs, sourceDir)
+		}
+	} else {
+		sourceDirs = append(sourceDirs, filepath.Join(projectDir, "src", "main"))
+	}
+
+	return sourceDirs, nil
```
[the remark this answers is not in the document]
florianGla: Yes, `sourceDirs` are only returned at the end of the function.

pkg/parser/libfuzzer/stacktrace/stacktrace.go
```diff
@@ -164,13 +186,21 @@ func (p *parser) stackFrameFromLine(line string) (*StackFrame, error) {
...
 		SourceFile:  filepath.ToSlash(sourceFile),
 		Line:        uint32(lineNumber),
 		Column:      uint32(column),
 		FrameNumber: uint32(frameNumber),
-		Function:    matches["function"],
-	}, nil
+		Function:    function,
+	}
+
+	log.Printf("xxxxxxxxxxxx %v", stackFrame)
```
jochil, line 200: I guess this is a left over

pkg/java/source_map.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/pkg/errors"
+)
+
+// SourceMap provides a mapping from package names
+// into the corresponding source file locations
+type SourceMap struct {
+	JavaPackages map[string][]string `json:"java_packages,omitempty"`
+}
+
+func CreateSourceMap(projectDir string, sourceDirs []string) (*SourceMap, error) {
```
jochil, line 18: I think we should cover this function by a test too... there is a lot of stuff happening here

pkg/java/source_map.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+	for _, file := range sourceFiles {
+		packageName, err := func() (string, error) {
+			fd, err := os.Open(file)
+			if err != nil {
+				return "", errors.WithStack(err)
+			}
+			defer fd.Close()
+			return GetPackageFromSource(fd), nil
+		}()
```
jochil, line 40: you are basically re-declaring this function over and over in a loop... why not extracting it and using `file` as a parameter?

pkg/java/source_map.go
florianGla: I can do that. I hoped that it is already covered by the integration tests for maven/gradle and jazzer_bundler. We would need an integration test with real test data.

pkg/java/source_map.go
florianGla: Good idea. I used a lambda function to get `fd.Close()` to work correctly.

jochil (approved):
thanks for all the adjustments

PR #1020, opened 2023-08-30
CLI-1242 Fix cifuzz init argument order

adombeck:
### Motivation/Context

The order of the arguments in the usage message of `cifuzz init` was undeterministic.

### Description

Use a slice which contains the supported arguments in the correct order.

### How to use/reproduce

```
make install && for run in {1..10}; do cifuzz init --help | grep "cifuzz init"; done
```


florianGla (approved):

adombeck:
@JackuB I implemented here what I proposed in a thread on slack regarding the argument order. Are you fine with that change?

adombeck:
> 



> @JackuB I implemented here what I proposed in a thread on slack regarding the argument order. Are you fine with that change?

@JackuB ping

JackuB (approved):

PR #1025, opened 2023-09-01
Fix CMake integration on Windows

florianGla:
Our CMake integration on Windows links libraries, which are contained in the Visual Studio folder.  These libraries are contained in a subfolder, which is named after the used clang version. We assumed that this name is equal to the full compiler version. It turns out that only the major version part is used. This PR fixes this issue and enables the skipped integration tests.


adombeck (approved):
Great find!

PR #1026, opened 2023-09-01
feat: --stop-signal-file for cifuzz execute

mlsandnerCI:
### Motivation/Context

> When executing fuzz containers in a k8s setup we need to signal to sidecar containers that the main process has finished to stop any sidecar containers running alongside the main cifuzz execution.

### Description
`cifuzz execute` now has a boolean flag `--stop-signal-file` that if set, creates an empty file called `cifuzz-execution-finished` on exit.

### How to use/reproduce
- create bundle e.g. for maven 
- run cifuzz execute in unzipped bundle with `--stop-signal-file`


jochil:
Can we add the flag already to the default `cifuzz execute` call? Or is there a reason against it?

internal/cmd/execute/execute.go
```diff
@@ -117,6 +128,7 @@ It is currently only intended for use with the 'cifuzz container' subcommand.
 
 	cmd.Flags().Bool("single-fuzz-test", false, "Run the only fuzz test in the bundle (without specifying the fuzz test name).")
 	cmd.Flags().Bool("print-bundle-metadata", false, "Print the bundle metadata as JSON.")
+	cmd.Flags().Bool("stop-signal-file", false, "CI Fuzz will create a file 'cifuzz-execution-finished' upon exit")
```
jochil, line 131: In the jira task it was defined as `--stop-signal-file=<filename>` not just a boolean flag... did this requirement changed somehow?

internal/cmd/execute/execute.go
```diff
@@ -55,11 +56,21 @@ It is currently only intended for use with the 'cifuzz container' subcommand.
...
 			cmdutils.ViperMustBindPFlag("single-fuzz-test", cmd.Flags().Lookup("single-fuzz-test"))
 			cmdutils.ViperMustBindPFlag("print-bundle-metadata", cmd.Flags().Lookup("print-bundle-metadata"))
+			cmdutils.ViperMustBindPFlag("stop-signal-file", cmd.Flags().Lookup("stop-signal-file"))
 			opts.SingleFuzzTest = viper.GetBool("single-fuzz-test")
 			opts.PrintBundleMetadata = viper.GetBool("print-bundle-metadata")
 			opts.PrintJSON = viper.GetBool("print-json")
 		},
 		RunE: func(c *cobra.Command, args []string) error {
+			if viper.GetBool("stop-signal-file") {
+				defer func() {
```
jochil, line 66: I think we should cover the creation of the stop file with a small test

adombeck (changes requested):
Citing https://code-intelligence.atlassian.net/browse/CLI-1235:
> The specified file should be written on exit of the cifuzz command and contain information if cifuzz exited with an error or as expected (e.g. write the exit code into the file). 

AFAICT the file that's written here does not tell the user anything about the exit status.

adombeck:
> Citing https://code-intelligence.atlassian.net/browse/CLI-1235:
> 
> > The specified file should be written on exit of the cifuzz command and contain information if cifuzz exited with an error or as expected (e.g. write the exit code into the file).
> 
> AFAICT the file that's written here does not tell the user anything about the exit status.

@simonresch clarified out of band that the backend team does not need the file to include the exit status.

internal/cmd/execute/execute.go
mlsandnerCI: As discussed in the planning, right now there is no reason for the filename to be configurable so it should be fine to keep it simple for now

internal/cmd/execute/execute.go
mlsandnerCI: I created a very simple test that calls the command with the flag and checks if the file has been created, no matter if the command succeeded or not. Is this okay?

jochil: works for me. thanks

jochil:
> > Can we add the flag already to the default `cifuzz execute` call? Or is there a reason against it?
> 
> You mean including it in the container run command where we call execute? I don't know if every user needs this file but of course it could be added 🤔

I mean if the file is there it creates no harm but the backend team don't have to take care of the flag

jochil:
> > > > Can we add the flag already to the default `cifuzz execute` call? Or is there a reason against it?
> > > 
> > > 
> > > You mean including it in the container run command where we call execute? I don't know if every user needs this file but of course it could be added 🤔
> > 
> > 
> > I mean if the file is there it creates no harm but the backend team don't have to take care of the flag
> 
> True but would we need the flag then at all? Then it could just be part of the regular execute workflow.

Good question... @simonresch @adombeck WDYT?

simonresch:
> > > > > Can we add the flag already to the default `cifuzz execute` call? Or is there a reason against it?
> > > > 
> > > > 
> > > > You mean including it in the container run command where we call execute? I don't know if every user needs this file but of course it could be added 🤔
> > > 
> > > 
> > > I mean if the file is there it creates no harm but the backend team don't have to take care of the flag
> > 
> > 
> > True but would we need the flag then at all? Then it could just be part of the regular execute workflow.
> 
> Good question... @simonresch @adombeck WDYT?

Setting the flag in the backend is not a problem. We would like to control where the file is saved anyways to make that uniform with the behavior of `fuzzer-runner` calls.

simonresch:
> > Setting the flag in the backend is not a problem. We would like to control where the file is saved anyways to make that uniform with the behavior of `fuzzer-runner` calls.
> 
> Does that mean you do need the flag to be configurable, so a string instead of boolean? @simonresch

A string flag where we can pass the file name would be ideal. We can't be sure that the directory where cifuzz is executed is writable in the k8s pod. Making this configurable would help there.

jochil (approved):

PR #1034, opened 2023-09-05
feat: access remote findings from CI Sense

MarkusZoppelt:
### Motivation/Context

Users should view / investigate their remote findings from CI Sense in cifuzz.

### Description

* ability to query remote findings via CLI
* works if authenticated and `project` is set
* if authenticated and `project` ist **NOT** set, show interactive dialog
* respects `--interactive`  flag

*Missing:*
Command completion for remote findings. Help appreciated.

### How to use/reproduce

Try out the `cifuzz finding` command when authenticated.

Integration test:

    go test ./internal/cmd/finding/... -v

e2e tests:

    go test ./e2e/tests/finding/... -v


internal/cmd/finding/finding.go
```diff
@@ -86,50 +90,131 @@ func newWithOptions(opts *options) *cobra.Command {
...
-	errorDetails, err := auth.TryGetErrorDetails(cmd.opts.Server)
+	errorDetails, token, err := auth.TryGetErrorDetailsAndToken(cmd.opts.Server)
+	if err != nil {
+		var connErr *api.ConnectionError
+		if !errors.As(err, &connErr) {
+			return err
+		} else {
+			log.Debugf("Connection error: %v", connErr)
+		}
+	}
```
adombeck, line 108: The `api.ConnectionError` is already handled by auth.TryGetErrorDetailsAndToken
```suggestion
	errorDetails, token, err := auth.TryGetErrorDetailsAndToken(cmd.opts.Server)
	if err != nil {
		return err
	}
```

pkg/finding/finding.go
```diff
@@ -35,6 +37,8 @@ type Finding struct {
 	MoreDetails        *ErrorDetails `json:"more_details,omitempty"`
 	Tag                uint64        `json:"tag,omitempty"`
```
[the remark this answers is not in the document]
MarkusZoppelt: According to https://gitlab.code-intelligence.com/code-intelligence/core/-/blob/master/ui/shared/api/src/swagger.json (see `GET /v1/{project}/findings`), there is a `Tag` field of type `string`. I changed the local finding struct and added the field during conversion.

internal/cmd/finding/finding.go
MarkusZoppelt: Fixed, thx!

pkg/finding/finding.go
```diff
@@ -35,6 +37,8 @@ type Finding struct {
...
 
+	Location string
+
```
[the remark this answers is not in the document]
MarkusZoppelt: I added a commit that should use `Finding.StackTrace` instead of `Location` (32125a380545d7917c36a857b1f2c7585d4ee503)

internal/cmd/finding/finding.go
```diff
@@ -284,3 +380,28 @@ func wrapLongStringToMultiline(s string, maxLineLength int) string {
...
+	// add option to cancel
+	items["<Cancel>"] = "<cancel>"
+
+	projectName, err := dialog.Select("Select a remote project:", items, true)
+	if err != nil {
+		return "<cancel>", err
+	}
+
+	return projectName, nil
+}
```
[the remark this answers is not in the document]
MarkusZoppelt: They are similar, but not *entirely* the same, e.g. we do not want users to create a new project during `cifuzz finding` with the `<<new>>` selection. It's also different wording in the two functions and I would argue that having them both separated makes sense for these different commands, since they offer different dialogs.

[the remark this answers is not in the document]
adombeck: Makes sense, thanks for explaining

internal/cmd/finding/finding.go
```diff
@@ -86,50 +91,130 @@ func newWithOptions(opts *options) *cobra.Command {
...
 		}
 
-		for _, f := range findings {
+		for _, f := range allFindings {
 			score := "n/a"
 			locationInfo := "n/a"
 			// add location (file, function, line) if available
-			if len(f.ShortDescriptionColumns()) > 1 {
+			if f.StackTrace != nil && len(f.StackTrace) > 0 {
+				locationInfo = fmt.Sprintf("%s:%s:%d", f.StackTrace[0].SourceFile, f.StackTrace[0].Function, f.StackTrace[0].Line)
```
adombeck, line 216: This produces output like `src/explore_me.cpp:exploreMe:13`. I would prefer omitting the function name here to produce the more common format `$FILE:$LINE` which is understood by existing tools, for example the "Go to file" action in JetBrains IDEs will bring you directly to the line in that file if you copy paste the string.

```suggestion
				locationInfo = fmt.Sprintf("%s:%d", f.StackTrace[0].SourceFile, f.StackTrace[0].Line)
```

adombeck: What I would like even better is to extract a method `finding.SourceLocation` which can be used here and in the `finding.ShortDescriptionColumns` method.

MarkusZoppelt: Thanks! I added 8b50eea that should do this.

adombeck (approved):

PR #1035, opened 2023-09-05
chore(deps): update examples to jazzerjs 1.7.0

jochil:
### Motivation/Context
we should keep our dependencies up to date

### Description
Adjusting the `./tools/updater/updater.go` tool to support updating npm based project to a specific jazzer.js version and then updating jazzer to 1.7.0

### How to use/reproduce
You can use the updater tool by 
```
go run ./tools/updater/updater.go --dependency jazzerjs --version 1.7.0
```


MarkusZoppelt (approved):
The updater is nice but I constantly forget that we have it.
Could we maybe utilise it more/automatically? Or put it in the `CONTRIBUTING.md`?

jochil:
> The updater is nice but I constantly forget that we have it. Could we maybe utilise it more/automatically? Or put it in the `CONTRIBUTING.md`?

I was thinking we can use it more when automating the dependency updates... we have a meeting about this next week

PR #1041, opened 2023-09-07
Validate java fuzz tests

florianGla:


internal/cmd/run/run.go
```diff
@@ -992,3 +1000,37 @@ func (c *runCmd) prepareCorpusDirs(buildResult *build.BuildResult) error {
 
 	return nil
 }
+
+// validateJVMFuzzTest checks if the given fuzz test is valid.
+// If no target method is specified, it will be added.
+func (c *runCmd) validateJVMFuzzTest(deps []string) error {
```
jochil, line 1015: Isn't this the exact same code as added in `internal/cmd/coverage/coverage.go`.
Can we reuse it somehow instead of duplicating it?

jochil (approved):

internal/cmd/run/run.go
florianGla: It is possible by passing the target method as a pointer. I tried to avoid that, but it is probably the better solution for now. Maybe we find a better solution after more refactoring steps.

PR #1042, opened 2023-09-07
Minor fixes and improvements

adombeck:
### Motivation/Context

I found this old branch which contains some minor unmerged fixes and improvements.

### Description

See commit messages


internal/container/client.go
```diff
@@ -7,8 +7,8 @@ import (
 
 var dockerClient *client.Client
 
-// getDockerClient returns a docker client and will also handle its closing. It will take configuration options in the future.
-func getDockerClient() (*client.Client, error) {
+// GetDockerClient returns a docker client and will also handle its closing. It will take configuration options in the future.
+func GetDockerClient() (*client.Client, error) {
```
jochil, line 11: Maybe using the opportunity to rename it to `DockerClient` (https://go.dev/doc/effective_go#Getters)

jochil (approved):

internal/container/client.go
adombeck: The function returns an `error` which IMO would be unexpected if it was just called `DockerClient()`. Also https://go.dev/doc/effective_go#Getters refers to getter *methods* but we're not talking about a method here.

jochil: You are right

PR #1043, opened 2023-09-07
fix(ci): enable CMake compatibility test on Windows

florianGla:
Now that cifuzz is released with a fix for the CMake issue on Windows, the compatibility check can be enabled again.


MarkusZoppelt (approved):

jochil:
According to https://github.com/CodeIntelligenceTesting/cifuzz/pull/1005/files more files where affected than in this change? Any reason for not changing them back, too?

florianGla:
They were already changed back in https://github.com/CodeIntelligenceTesting/cifuzz/pull/1025

jochil:
> They were already changed back in #1025

great, thanks

PR #1045, opened 2023-09-08
Fix race condition when executing multiple `cifuzz container run` commands in parallel

adombeck:
### Motivation/Context

The cifuzz container run command always tagged the created Docker image as "cifuzz" and used that tag to create the container. That caused a race condition when multiple tests which use that command run in parallel, which was happening in our CI.

Closes CLI-1266


internal/container/container.go
```diff
@@ -17,15 +17,15 @@ import (
...
+func Create(imageID string, printJSON bool) (string, error) {
 	cli, err := GetDockerClient()
 	if err != nil {
 		return "", err
 	}
 
 	hostConfig := &container.HostConfig{}
 	containerConfig := &container.Config{
-		Image:        "cifuzz",
+		Image:        imageID,
```
jochil, line 28: what do you thing about tagging the container to make it easier to find?

jochil (approved):

internal/container/container.go
adombeck: What do you mean with tagging the container? Docker only supports tags for images. We already set the tag "cifuzz" for the images we create. For containers, [`name`](https://docs.docker.com/engine/reference/run/#name---name) and [`label`](https://docs.docker.com/config/labels-custom-metadata/) is supported.

jochil: ah got a wrong turn... I meant tagging the image. 
If we are already doing it, fine :+1:

PR #1046, opened 2023-09-08
fuzz container: Use user-specified options (--dict, --engine-args, --seed-corpus, --timeout, --plain)

adombeck:
### Motivation/Context

The fuzz container command ignored many of the options which the user can specify.

### Description

Use the user-specified options.

### How to use/reproduce

Try using the options in the `cifuzz container run` command. For example the `--engine-arg` option can be easily tested via:

```
make install
make build-container-image
cifuzz -C examples/maven container run com.example.FuzzTestCase::myFuzzTest --interactive=false --no-notifications -v --engine-arg=-help=1
```

which should print the libFuzzer usage message.


internal/cmd/execute/execute.go
```diff
@@ -223,6 +223,16 @@ func (c *executeCmd) run(metadata *archive.Metadata) error {
 		EnvVars:            []string{"NO_CIFUZZ=1"},
 	}
 
+	// Specify the dictionary file if the bundle includes one.
+	dictFileName := "dict"
```
jochil, line 231: Isn't this something that can be defined by the user? Or is `dict` the default name?

jochil (approved):

internal/cmd/execute/execute.go
adombeck: The user-specified dictionary is always stored in a file named `dict`:
https://github.com/CodeIntelligenceTesting/cifuzz/blob/bfc539643a46233bd008b241d5afb1001a0f6d14/internal/bundler/libfuzzer_bundler.go#L518-L522
https://github.com/CodeIntelligenceTesting/cifuzz/blob/bfc539643a46233bd008b241d5afb1001a0f6d14/internal/bundler/jazzer_bundler.go#L75-L79

PR #1047, opened 2023-09-08
refactor: split run command into build system specific runner

jochil:
### Motivation/Context
Adding new build systems and languages became a bit complicated as a lot things where happening inside the run command itself. We wanted to improve readability and separate the build specific code into separate files/packages

### Description
We introduced the `Runner` interface and created a single runner for each build system. Code that is used by multiple build systems is shared (see `runner.go` & `shared.go`)

### How to use/reproduce
When done, everything should work as before ;)


internal/cmd/run/runner/runner.go
```diff
@@ -0,0 +1,206 @@
...
+	"code-intelligence.com/cifuzz/pkg/log"
+	"code-intelligence.com/cifuzz/pkg/runner/jazzer"
+	"code-intelligence.com/cifuzz/pkg/runner/libfuzzer"
+	"code-intelligence.com/cifuzz/util/fileutil"
+)
+
+type Runner interface {
+	Run(*RunOptions) (*reporthandler.ReportHandler, error)
+	CheckDependencies(string) error
+}
```
adombeck, line 30: Can we find another name for the new interface to avoid confusion with the existing `Runner` (renamed to `FuzzerRunner` in this PR) interface?

jochil: Sure... any proposals? Off the top of my head: `CommandRunner`, `FuzzTestRunner`, `BuildSystemRunner`, `TestRunner`
Or another base name like `*Executor`, `*Handler`, `*Adapter`

adombeck: I like `BuildSystemAdapter` as the interface name and `CMakeAdapter`, `MavenAdapter`, etc as the implementations. It doesn't really conform to the definition of the [adapter pattern](https://en.wikipedia.org/wiki/Adapter_pattern) though, which might be confusing..

jochil: I like it too... would mean we rename the whole package to adapter.
Honestly I see no problem with the adapter pattern. I would not expect a go project implementing a design pattern from the OOP world. At least in my experience is not that common for go

jochil: I added a commit at the end of the PR

adombeck (approved):
I see a lot of things which I would like to refactor but I think the easiest thing to do to avoid further merge conflicts is to merge it first like this and then I can do some more refactoring in a follow-up PR.

jochil:
> I see a lot of things which I would like to refactor but I think the easiest thing to do to avoid further merge conflicts is to merge it first like this and then I can do some more refactoring in a follow-up PR.

I see this change more as a starting point, too. Having the build system specific things separated might open the view for other improvements

PR #1052, opened 2023-09-12
Print Jazzer/Jazzer.js version infos in verbose mode

florianGla:
In order to better understand the behavior of fuzzing runs with Jazzer and Jazzer.js we print the used versions of Jazzer/JUnit and Jazzer.js/Jest in verbose mode.

### Motivation/Context

This shall help to debug user issues.

### Description

For jazzer/JUnit the versions are extracted from the class path. This has the advantage that no build system specific functionality has to be added.

### How to use/reproduce

Start of fuzzing run for Jazzer and Jazzer.js in verbose mode.


pkg/dependencies/version.go
```diff
@@ -181,6 +184,32 @@ func gradleVersion(dep *Dependency, projectDir string) (*semver.Version, error)
...
+}
+
+func JestVersion() (*semver.Version, error) {
+	cmd := exec.Command("npx", []string{"jest", "--version"}...)
+	output, err := cmd.CombinedOutput()
+	if err != nil {
+		return nil, errors.WithStack(err)
+	}
+	return extractVersion(string(output), nodeRegex, "jest")
+}
```
adombeck, line 211: By using `cmd.Output` instead of `cmd.CombinedOutput` and wrapping the error in `cmdutils.ExecError`, the stderr of the command is printed on failure, which allows to debug failures.

```suggestion
func JazzerJSVersion() (*semver.Version, error) {
	cmd := exec.Command("npx", []string{"jazzer", "--version"}...)
	output, err := cmd.Output()
	if err != nil {
		return nil, cmdutils.WrapExecError(errors.WithStack(err), cmd)
	}
	return extractVersion(string(output), nodeRegex, "jazzer.js")
}

func JestVersion() (*semver.Version, error) {
	cmd := exec.Command("npx", []string{"jest", "--version"}...)
	output, err := cmd.Output()
	if err != nil {
		return nil, cmdutils.WrapExecError(errors.WithStack(err), cmd)
	}
	return extractVersion(string(output), nodeRegex, "jest")
}
```

adombeck (approved):

PR #1056, opened 2023-09-12
fix: Adapt usage message of `--docker-image` for container command

adombeck:
### Motivation/Context

The usage message was written for the `cifuzz bundle` command and reused for `cifuzz container run` and `cifuzz container remote-run` but did not describe the flag's behavior correctly for those commands.

### How to use/reproduce

```
make install && cifuzz container run --help
```


internal/cmdutils/pflag.go
```diff
@@ -115,13 +115,26 @@ func AddDictFlag(cmd *cobra.Command) func() {
...
 }
 
-func AddDockerImageFlag(cmd *cobra.Command) func() {
+func AddDockerImageFlagForContainerCommand(cmd *cobra.Command) func() {
+	// Default was originally set to "ubuntu:rolling", but this is not correct
+	// It will be set by the bundle command depending on the build system, unless user overrides it
+	cmd.Flags().String("docker-image", "",
+		"A Docker image which is used as the base for the container image.\nThe image must contain all the dependencies required to run the fuzz test.\n"+
+			"By default, the image is chosen automatically based on the build system\n"+
+			"(\"eclipse-temurin:20\" for Java build systems, \"ubuntu:rolling\" for others).")
```
MarkusZoppelt, line 124: This looks awkwardly formatted.

How about 
```suggestion
	cmd.Flags().String("docker-image", "",
`A Docker image which is used as the base for the container image.
The image must contain all the dependencies required to run the fuzz test.
By default, the image is chosen automatically based on the build system
("eclipse-temurin:20" for Java build systems, "ubuntu:rolling" for others).`)
```

MarkusZoppelt (approved):

internal/cmdutils/pflag.go
adombeck: applied

PR #1062, opened 2023-09-13
fix(coverage): use output format for maven and gradle

florianGla:
This PR passes the output format to the coverage generators for maven and gradle.

### How to use/reproduce

Run `cifuzz coverage --output=cov --format=jacocoxml com.example.FuzzTestCase::myFuzzTest`.


MarkusZoppelt (commented):
This looks like something we could test easily. Should we add a test?

florianGla:
> This looks like something we could test easily. Should we add a test?

I will try to add tests to our integration tests for maven and gradle.

PR #1063, opened 2023-09-14
ci: install docker-ce with buildx support on ubuntu

MarkusZoppelt:
### Motivation/Context

`docker.io` is an unofficial package and we should rely on `docker-ce`.

This causes our E2E Source Installation Pipeline to fail on Ubuntu:
```
ERROR: BuildKit is enabled but the buildx component is missing or broken.
       Install the buildx component to build images with BuildKit:
       https://docs.docker.com/go/buildx/
```

### Description

Install `docker-ce` according to https://docs.docker.com/engine/install/ubuntu/#install-using-the-repository

### How to use/reproduce

E2E Source Installation Pipeline will no longer fail on Ubuntu.


jochil (approved):
thanks for the fix, but if this test continue to make troubles we can delete it... IMO it is somehow redundant

PR #1066, opened 2023-09-14
feat: Ignore CIFUZZ_UID not set

adombeck:
### Motivation/Context

The backend team doesn't currently set `CIFUZZ_UID`.

### Description

Instead of requiring that `CIFUZZ_UID` is set, `cifuzz execute` now simply doesn't change the UID if it's not set.


adombeck:
Superseded by https://github.com/CodeIntelligenceTesting/cifuzz/pull/1077

PR #1067, opened 2023-09-15
ci: remove archlinux from workflows and actions

MarkusZoppelt:
### Motivation/Context

The E2E Source Installation pipeline is broken for `archlinux` and we probably don't need to test building from source in our pipeline, since building from source is usually covered by our devs.

### Description

Remove `archlinux` from workflows and actions.

### How to use/reproduce

E2E Source Installation pipeline should be gone.


jochil (approved):
goodbye arch :cry:

PR #1068, opened 2023-09-15
Add commands to print build flags

florianGla:
This PR adds hidden commands to print build flags. 

### Motivation/Context

Users that build their executables outside of cifuzz need to compile them with specific flags to ensure that cifuzz can fuzz them.

### How to use/reproduce

`cifuzz print-cflags`
`cifuzz print-cxxflags`
`cifuzz print-ldflags`
`cifuzz print-fuzz-test-cflags`
`cifuzz print-fuzz-test-cxxflags`

Each command has a `--coverage` flag to print the flags for a coverage build.


internal/cmd/print-flags/printflags.go
```diff
@@ -0,0 +1,59 @@
...
+				env, err = other.SetLibFuzzerEnv(env, runfiles.Finder)
+				if err != nil {
+					return err
+				}
+			}
+			fmt.Print(envutil.Getenv(env, flags))
+			return nil
+		},
+	}
+	coverage = cmd.Flags().Bool("coverage", false, "print flags for coverage")
```
adombeck, line 56: Concise and clean solution, love it!

adombeck (approved):

PR #1070, opened 2023-09-15
fix: trim/add `projects/` prefix correctly

MarkusZoppelt:
### Motivation/Context

The `/v1` / `/v2` API expects a `projects/` prefix for project ~~IDs~~ names but the `/v3` API will phase out the old project names (called `project_external_id`) for a newer format.

Since we already use the `/v3` API now (e.g. for `container remote-run`), we encountered some bugs were the `projects/` prefix was incorrectly set during API calls.

### Description

* Remove `projects/` prefix from all project names during `ListProjects`
* Add the prefix manually where necessary. This is cleaner and harder to overlook than trimming the prefix in a bunch of different places.


adombeck:
Can you explain (ideally in a code comment or commit message) in which case the `project` argument is prefixed with `projects/` and in which case it does not have that prefix?

MarkusZoppelt:
Converting to draft because I was thinking of a better fix.

MarkusZoppelt:
> Can you explain (ideally in a code comment or commit message) in which case the `project` argument is prefixed with `projects/` and in which case it does not have that prefix?

@adombeck I updated the PR / commit and tried to document everything as good as possible.
Hope this helps. There are some more discussions about the `v1` / `v2` vs. `v3` API in our Slack.

internal/api/container_run.go
```diff
@@ -49,11 +50,21 @@ type Job struct {
...
 func (client *APIClient) PostContainerRemoteRun(image string, project string, fuzzTests []string, token string) error {
 	tests := []*FuzzTest{}
 	for _, fuzzTest := range fuzzTests {
 		tests = append(tests, &FuzzTest{Name: fuzzTest})
 	}
+
+	// the /v3 project_external_id is the project name without the projects/
+	// prefix and url escaped
+	project = strings.TrimPrefix(project, "projects/")
+	project, err := url.QueryUnescape(project)
```
jochil, line 63: IMO we should use always the unescaped string and only when needed explicitly escape it

MarkusZoppelt: This is how I done it. It only needs to be unescaped here. It will fail if unescaped in other calls (to v1 and v2)

MarkusZoppelt: There's a fundamental difference in how v1/v2 and v3 handle project IDs / names.
v1 and v2 expect project names. These do not to be unescaped.

v3 expects either a `project_external_id` (like here) or a `project_nid` which will be used in the future, eventually. Those boot need to be unescaped.

jochil: I got this the first place, but maybe my comment was unclear. I meant that we should not have to call `url.QueryUnescape` here as the given `project` param shouldn't be escaped. Whenever we need we should call `url.QueryEscape`. For me it is confusing to explicitly unescape a string when using it somewhere else than a url

MarkusZoppelt: Sorry, I am still not sure what you mean.

We are getting `project` from a `GET` to `v1/projects/`. These `project`s will have special characters in them, i.e., they are always in a format that we need to unescape for later calls to `v3`.

Example: a project created in CI Sense with the name "my project" will be:
 * returned by `v1/projects` in the form of `projects/my%20project-01f80da4` (if we don't alter the resulting name in any way)
 * our first alteration should be to strip the prefix, so that we don't show users the prefix in dialogs, selections, tables, etc. Let's call that _p_: `my%20project-01f80da4`
 * When sending requests to `v1` and `v2`, we need to refer to `project` **with** the prefix: `projects/` + _p_: `projects/my%20project-01f80da4`
 
* **_However_**, for a `POST` to `v3/runs` (and only here), we need to transform _p_ so that it contains the correct special characters. So -only here- we need to use _QueryUnescape(p)_, so that it becomes: `my project-01f80da4`

This should be what I am trying to do in my PR. Can you give me a code snippet or a location what or where I should change things?

MarkusZoppelt: Also, this is not new bahavior, we merged using `url.QueryUnescape` in #1054 already:
https://github.com/CodeIntelligenceTesting/cifuzz/blob/2dcf95966d61a02b5752c1802df9cd93584e359e/pkg/dialog/dialog.go#L158

The place is now just moved to the single location where we need to call it.

jochil: My point is that we should convert `projects/my%20project-01f80da4` to `my project-01f80da4` as soon as possible... best case would be immediately after receiving it from the backend. If we need to encode it and/or add `/projects/` to it, we can do it in actual functions needing it for the specific requests. 

But I see the point that it came from an old PR... maybe you can think about it and for now we continue with merging this one :+1:

jochil (approved):

internal/api/container_run.go
MarkusZoppelt: Will do, thanks. We need to transition to the new `project_nid` format anyway, so we need to revisit this topic anyway.

PR #1071, opened 2023-09-15
chore(deps): remove PTerm replacement

MarvinJWendt:
### Motivation/Context

As you guys helped out to find the bug in PTerm, I thought I'd be so kind and update PTerm here with the fixed version.

### Description

Removed the PTerm fork replacement, as https://github.com/pterm/pterm/issues/562 is now fixed.
More context in in #1009 

### How to use/reproduce

Magic 🔮

(No usage changes)


adombeck:
Thanks a lot! Since the new pterm version requires Go 1.21, we'll first have to fix an issue in our code when building with Go 1.21 before we can merge this.

jochil:
We finished upgrading to go 1.21... so I rebased this branch against `main` and it should be ready to go
Thanks for contributing

jochil (approved):

PR #1075, opened 2023-09-18
fix: more `projects/` related fixes

MarkusZoppelt:
~~@adombeck tagging you because you implemented `func UploadBundle` with the `signalHandlerCtx, cancelSignalHandler`.
My commit fixes [this bug](https://codeintelligence.slack.com/archives/C03R639E2JJ/p1695050037887739) caused by my recent "project name" changes but I think we should change the function to use `api.sendRequest()` instead to prevent future bugs like this~~.

I refactored the entire project name handling by introducing conversion functions for all occurrences of project names so that we can make sure that they are used correctly for internal use, v1, v2, and v3 API calls.

Manually tested against app.code-intelligence.com and working for:

* cifuzz run

    `cifuzz run -C examples/gradle com.example.FuzzTestCase::myFuzzTest --project my%20project-01f80da4 -v`

* cifuzz finding 

    `cifuzz finding -C examples/gradle --project my%20project-01f80da4 -v`

* cifuzz remote-run

    `cifuzz remote-run -C examples/gradle com.example.FuzzTestCase::myFuzzTest --project my%20project-01f80da4 -v`

* cifuzz container remote-run

    c`ifuzz container remote-run -C examples/gradle --registry localhost:5000/test/cifuzz  com.example.FuzzTestCase::myFuzzTest --project my%20project-01f80da4 -v`


adombeck:
> @adombeck tagging you because you implemented `func UploadBundle` with the `signalHandlerCtx, cancelSignalHandler`.
> 
> My commit fixes [this bug](https://codeintelligence.slack.com/archives/C03R639E2JJ/p1695050037887739) caused by my recent "project name" changes but I think we should change the function to use `api.sendRequest()` instead to prevent future bugs like this.

I don't understand how this bug would have been prevented if `APIClient.sendRequest()` was used in that method (which would require a lot of changes to that small function to make it reusable by `APIClient.uploadBundle`).

internal/cmd/run/run.go
```diff
@@ -339,21 +339,6 @@ func (c *runCmd) uploadFindings(fuzzTarget, buildSystem string, firstMetrics *re
...
-				found = true
-				break
-			}
-		}
-
-		if !found {
-			message := fmt.Sprintf(`Project %s does not exist on server %s.
-Findings have *not* been uploaded. Please check the 'project' entry in your cifuzz.yml.`, project, c.opts.Server)
-			return errors.New(message)
-		}
```
adombeck, line 356: Why do we remove this check instead of fixing it?

MarkusZoppelt: Done

jochil (approved):
I think there is still room to improve this further... for example it might make sense to differentiate explicitly between things like `name`, `id`, `slug`, `path`, `label`, ... 
Currently there is still knowledge about v1,v2,v3 required to work with the project name, were it IMO should be encapsulated in the `api` package.

Maybe we can work on this together in another PR, but to get the problem fixed we can go on with this

PR #1077, opened 2023-09-19
feat(container): Support running fuzz container as non-root user

adombeck:
### Motivation/Context

The backend team wants to run the fuzz container as a non-root user. This is currently not supported by the fuzz container for multiple reasons:

* We require `CIFUZZ_UID` to be set and use `setuid` to change to that UID. Only root is allowed to call `setuid`, so this requires that the fuzz container is started as root.

* The fuzz container creates the `/cifuzz/seed` and `/cifuzz/corpus` directories which fails if it does not have write permissions on the `/cifuzz` directory. That directory is copied from the host during image build via the `COPY` instruction in the Dockerfile. Docker doesn’t support changing the mode or permission of the destination directory (unless we build with buildkit, which we currently don’t).

### Description

* Removes support for the `CIFUZZ_UID` environment variable. In `cifuzz container run`, we now set the `container.Config.User` field to make Docker run the container with the same UID as the host user.

* Uses the corpus directories `/tmp/generated-corpus` and `/tmp/managed-seed-corpus` instead of `/cifuzz/corpus` and `/cifuzz/seed` because we now support running the fuzz container as a user which doesn't have write permission to the `/cifuzz` directory.

### How to use/reproduce

```
make install && make build-container-image && cifuzz -C examples/maven container run com.example.FuzzTestCase::myFuzzTest --interactive=false --no-notifications -v
```


adombeck:
This is blocked by Jazzer creating an unused `.cifuzz-corpus` dir in the current working directory.

adombeck:
> > Will I be able to download the result of this build and try it out?
>
> The summary page of the "Installer Pipeline / installer / build installer + upload as artifact (ubuntu-20.04)" job contains a cifuzz_installer artifact. But maybe wait until I debugged the test failures.

@tkrueger Tests are passing now, so you could now test the cifuzz installer from the pipeline artifacts. Sorry it took so long, it was first blocked by an issue in Jazzer and then by a Jazzer.js update.

tkrueger:
@adombeck Thanks! Will take it for a spin and come back. And I don't think it took long at all, tbh.

PR #1081, opened 2023-09-20
chore: Migrate Jest config to be compatible with Jazzer.js 2.0.0

adombeck:
### Motivation/Context

See https://github.com/CodeIntelligenceTesting/jazzer.js/blob/main/docs/migration-guide.md

### Description

Closes CLI-1297


MarkusZoppelt:
I also added a commit to update the examples to work with the latest jazzer.js release

florianGla (approved):
The package.json in integration-tests/errors/nodejs/testdata/ needs to be updated.
Why are there changes for integration-tests/nodejs/nodejs_test.go and integration-tests/node-typescript/nodets_test.go is not changed?

adombeck:
> The package.json in integration-tests/errors/nodejs/testdata/ needs to be updated.

I did that at first but that broke the tests. It hasn't been updated since Jazzer.js 1.6.1 so I assume it's fine to delay the update a bit more and merge this as is because it fixes the pipelines on all our branches. Could you create a follow-up task?

> Why are there changes for integration-tests/nodejs/nodejs_test.go and integration-tests/node-typescript/nodets_test.go is not changed?

Because the migration broke `integration-tests/nodejs/nodejs_test.go` but not `integration-tests/node-typescript/nodets_test.go`. I don't know why, maybe @bertschneider is interested in investigating why the update to Jazzer.js 2.0.0 changed the branch coverage in our nodejs test but not in the nodets test.

bertschneider:
I investigated the two required changes:

- The function name in the second stack frame is indeed new. In the new integration Jest loads the tests and apparently does that in a way that the called function is now included in the stack frame. If an exception is thrown before invoking `exploreMe` the name is not present in the stack. I would leave it at that for now. 

- I compared the coverage runs of both projects and noticed that 8 branches are found in the JS project, whereas only 4 are found in the TS one. As far as I can see the transpiled code looks identical in both cases. We will investigate this further.

adombeck:
> I investigated the two required changes: [...]

Thanks a lot!

PR #1082, opened 2023-09-21
fix(reporthandler): start the printer after fuzzer is initialized

florianGla:
When running a fuzz test, we show the fuzzer execution stats (paths, exec/s etc.) before the fuzzer actually started fuzzing.

### Motivation/Context

This is confusing because it gives the user the impression that the fuzz test was already started.

### Description

When creating a new updating printer, a new go routine is started, which continuously prints the fuzzing stats. This go routine can be removed. New metrics are already printed by the reporthandler when they are available.

### How to use/reproduce

Start a fuzzing run for an example project.


adombeck:
> When creating a new updating printer, a new go routine is started, which continuously prints the fuzzing stats. This go routine can be removed. New metrics are already printed by the reporthandler when they are available.

The purpose of that go routine is to ensure that the updating printer is updated once per second, to display the correct runtime and duration since last new path. If we delete it, the updating printer is only updated when libFuzzer prints new metrics, which can take many seconds if the fuzzer doesn't find any new paths.

adombeck:
I think the correct solution is to let the libFuzzer output parser figure out when the fuzzing actually started and only then start the updating printer.

adombeck:
> If we delete it, the updating printer is only updated when libFuzzer prints new metrics, which can take many seconds if the fuzzer doesn't find any new paths.

You can reproduce this by running a fuzz test which doesn't find a fatal finding and doesn't find any new paths after a few seconds:

```bash
git apply << EOF
heredoc> diff --git a/examples/cmake/src/explore_me.cpp b/examples/cmake/src/explore_me.cpp
index 00e01a18..45c7a103 100644
--- a/examples/cmake/src/explore_me.cpp
+++ b/examples/cmake/src/explore_me.cpp
@@ -14,9 +14,9 @@ void exploreMe(int a, int b, string c) {

         if (c == "FUZZING") {
           // Trigger a heap buffer overflow
-          char *s = (char *)malloc(1);
-          strcpy(s, "too long");
-          printf("%s\n", s);
+//          char *s = (char *)malloc(1);
+//          strcpy(s, "too long");
+//          printf("%s\n", s);
         }
       }
     }
EOF
cifuzz -C examples/cmake run my_fuzz_test
```

florianGla:
All right. I thought it is ok if the printer is only updated, when libFuzzer prints new metrics. I'll try to activate the go routine after the fuzzer is initialized.

adombeck:
I pushed two more commits. @jochil, @mlsandnerCI could one of you review those?

jochil:
> I pushed two more commits. @jochil, @mlsandnerCI could one of you review those?

I only see two commits from florian 5 days ago?!

adombeck:
> I only see two commits from florian 5 days ago?!

Huh, maybe my push was aborted by the pre-push hook. I pushed my commits now.

jochil (approved):
lgtm, thanks

PR #1083, opened 2023-09-21
test: use mockserver instead of staging environment for e2e tests

MarkusZoppelt:
### Motivation/Context

Testing against `app.staging.code-intelligence.com` needs us to test our code to an external dependency and requires us to store production tokens in our GitHub secrets.
Using the mockserver can test if our API calls are sent and we can do basic checking of the conntents without pinging `staging`.

### Description

Replaces all calls to `app.staging.code-intelligence.com` with a server mock.

### How to use/reproduce

    make test/e2e


jochil (approved):

jochil:
@MarkusZoppelt I think we can update the pipeline `.github/workflows/pipeline_e2e_matrix_nightly.yml` as well. This might not be needed anymore:
```
env:
   E2E_TEST_CIFUZZ_API_TOKEN: ${{ secrets.E2E_TEST_CIFUZZ_API_TOKEN }}
```
As well as the github secret as well

And if you changing the pipeline, can you please add
```
on:
  pull_request:
    branches: [main]
```
to make sure it is executed on PRs (as it is required now)

PR #1087, opened 2023-09-25
Fix --seed-corpus flag behavior with Java fuzz tests

adombeck:
### Motivation/Context

We used to pass the first directory passed via a `--seed-corpus` as the first positional argument to Jazzer, so Jazzer passed that as the first corpus directory to libFuzzer, which had the effect that libFuzzer wrote new seeds back to that directory instead of the generated corpus directory.

To avoid that, this commit always passes the generated corpus directory as the first positional argument to Jazzer. 

Closes CLI-1307

### How to use/reproduce

To reproduce the issue on main, run the maven example fuzz test with the `--seed-corpus` flag (the project directory needs to be cleaned to avoid that existing seeds trigger the finding during initialization, which causes the fuzzer to abort before it puts any new seeds in the corpus directory):

```
git -C examples/maven clean -fdx && \
rm -rf /tmp/seeds && \
mkdir -p /tmp/seeds && \
make install && \
cifuzz run -C examples/maven com.example.FuzzTestCase::myFuzzTest --interactive=false --no-notifications --seed-corpus /tmp/seeds/ -v && \
tree /tmp/seeds && \
tree examples/maven/.cifuzz-corpus
```

You will see that on main, new seeds have been added to `/tmp/seeds` instead of `examples/maven/.cifuzz-corpus`. If you rerun the same command on this branch, the seeds are added to `examples/maven/.cifuzz-corpus` instead.


pkg/parser/libfuzzer/libfuzzer.go
```diff
@@ -163,23 +161,6 @@ func (p *parser) parseLine(ctx context.Context, line string) error {
...
-						GeneratedCorpus: inputPath,
-					})
-				} else if strings.Contains(inputPath, p.ProjectDir) {
-					return p.sendReport(ctx, &report.Report{
-						SeedCorpus: inputPath,
-					})
-				}
-			}
-		}
 		// Try to parse the line as the libFuzzer message which tells us
```
florianGla, line 164: Why shall we not parse jazzer input path messages anymore?

pkg/parser/libfuzzer/libfuzzer.go
```diff
@@ -163,23 +161,6 @@ func (p *parser) parseLine(ctx context.Context, line string) error {
...
 		// how many seeds the corpus contains.
 		// Note: We used to stop parsing lines until we saw the first
```
florianGla, line 166: Why shall we not parse jazzer input path messages anymore?

internal/cmd/run/reporthandler/reporthandler.go
```diff
@@ -89,19 +89,6 @@ func NewReportHandler(fuzzTest string, options *ReportHandlerOptions) (*ReportHa
...
-		}
-		if r.GeneratedCorpus != "" {
-			h.GeneratedCorpusDir = r.GeneratedCorpus
-		}
-		return nil
-	}
-
 	if r.Status == report.RunStatusInitializing && !h.initStarted {
 		h.initStarted = true
 		h.numSeedsAtInit = r.NumSeeds
```
florianGla, line 94: Why shall we not parse jazzer input path messages anymore?

mlsandnerCI: I think Adrians commit message answers this:
```
We know the corpus dirs before running Jazzer, so we don't parse
Jazzer's output anymore to figure out the corpus dirs
```
We don't need to parse the input path from jazzer anymore to get the directories since we always set them now 🤔

florianGla: I'm not sure if this is correct. With this PR we set the generated corpus directory, but I think we should still parse the seed corpus directory. Moreover we parse the input path for jazzer.js with the same code block. This PR does not change the generated corpus dir for jazzer.js and I'm not sure if jazzer.js allows to change that.

mlsandnerCI: Probably the reason why the nodeJS tests are failing...

PR #1091, opened 2023-09-27
feat(execute): add flag to write json output to a file

jochil:
### Motivation/Context
As discussed with the backend team we want to introduce a flag that allows writing the json output of `cifuzz execute` directly to a specified file

### Description
Previously the `reporthandler` directed the json output, depending on the `PrintJSON` variable. This is not working anymore with the new flag as there are now multiple combinations possible with the `--json` flag.
Now it accepts two `io.Writer`  for the json & regular output

### How to use/reproduce
```
make install
cd examples/cmake
cifuzz bundle my_fuzz_test
tar -xcf my_fuzz_test.tar.gz
cifuzz execute my_fuzz_test --json-output-file=test.json
```


internal/cmd/run/reporthandler/reporthandler.go
```diff
@@ -153,23 +152,11 @@ func (h *ReportHandler) Handle(r *report.Report) error {
...
-		}
-		jsonString = string(bytes)
-	} else {
-		jsonString, err = stringutil.ToJSONString(r)
-		if err != nil {
-			return err
-		}
+	jsonString, err := stringutil.ToJSONString(r)
+	if err != nil {
+		return err
```
adombeck: I see that printing with colors was broken before, because the condition was inverted (it should have been `file, ok := p.Writer.(*os.File); ok && term.IsTerminal(int(file.Fd()))`) but we can fix that instead of removing the functionality to print with colors. IMO the JSON output is quite a bit more readable with colors.

adombeck (commented):
The output of the updating printer flickers on this branch. To reproduce:

```
git apply << EOF
diff --git a/examples/cmake/src/explore_me.cpp b/examples/cmake/src/explore_me.cpp
index 00e01a18..45c7a103 100644
--- a/examples/cmake/src/explore_me.cpp
+++ b/examples/cmake/src/explore_me.cpp
@@ -14,9 +14,9 @@ void exploreMe(int a, int b, string c) {

         if (c == "FUZZING") {
           // Trigger a heap buffer overflow
-          char *s = (char *)malloc(1);
-          strcpy(s, "too long");
-          printf("%s\n", s);
+//          char *s = (char *)malloc(1);
+//          strcpy(s, "too long");
+//          printf("%s\n", s);
         }
       }
     }
EOF
make install && cifuzz -C examples/cmake run my_fuzz_test
```

internal/cmd/execute/execute.go
```diff
@@ -91,28 +94,7 @@ It is currently only intended for use with the 'cifuzz container' subcommand.
...
-					if fuzzerName == "" {
-						fuzzerName = fuzzer.Name
-					}
-					fmt.Printf("  %s\n", fuzzerName)
-					fmt.Printf("    using: %s\n", fuzzer.Engine)
-					fmt.Printf("    run fuzz test with: cifuzz execute %s\n", fuzzerName)
-					fmt.Println("")
-				}
-				return nil
+				return printNotice(metadata)
```
adombeck, line 97: I would appreciate if you could add unrelated changes like this in a separate commit.

internal/cmd/execute/execute.go
```diff
@@ -145,6 +129,23 @@ It is currently only intended for use with the 'cifuzz container' subcommand.
...
+		}
+		defer jsonOutputFile.Close()
+	}
+
+	if c.opts.PrintBundleMetadata {
+		err := printMetadata(metadata, jsonOutputFile)
+		if err != nil {
+			return err
+		}
+	}
```
adombeck, line 147: This causes the metadata to be printed twice. I suspect you wanted to remove lines 87-93.

internal/cmd/execute/execute.go
```diff
@@ -264,6 +284,48 @@ func getMetadata() (*archive.Metadata, error) {
...
+		return err
+	}
+	fmt.Println(metadataJSON)
+
+	if jsonOutputFile != nil {
+		_, err := jsonOutputFile.WriteString(metadataJSON)
+		if err != nil {
+			return errors.WithStack(err)
+		}
+	}
```
adombeck, line 299: I don't think we need to print the metadata both to stdout and the JSON file, one should be enough. We could ask the backend team which they prefer (maybe they don't use the metadata in the JSON file).

internal/cmd/execute/execute.go
```diff
@@ -133,6 +116,7 @@ It is currently only intended for use with the 'cifuzz container' subcommand.
 	cmd.Flags().Bool("single-fuzz-test", false, "Run the only fuzz test in the bundle (without specifying the fuzz test name).")
 	cmd.Flags().Bool("print-bundle-metadata", false, "Print the bundle metadata as JSON.")
 	cmd.Flags().String("stop-signal-file", "", "CI Fuzz will create a file 'cifuzz-execution-finished' upon exit")
+	cmd.Flags().String("json-output-file", "", "Path for saving the json output into a file")
```
adombeck, line 119: The JSON in the output file contains colors. This line needs to be adapted: https://github.com/CodeIntelligenceTesting/cifuzz/blob/65f2b4c082ce761f0c684d80d7f08f7034ae70d1/internal/cmd/execute/execute.go#L209-L209

adombeck, line 119:
```suggestion
	cmd.Flags().String("json-output-file", "", "Print output as JSON to the specified file (implies --json)")
```

internal/cmd/execute/execute.go
```diff
@@ -159,17 +160,36 @@ func (c *executeCmd) run(metadata *archive.Metadata) error {
...
+		printerOutput = os.Stderr
+		jsonOutput = io.MultiWriter(os.Stdout, jsonOutputFile)
+	} else if !c.opts.PrintJSON && jsonOutputFile != nil {
+		// keep regular terminal output but write json to a file
+		printerOutput = os.Stdout
+		jsonOutput = jsonOutputFile
+	} else {
+		printerOutput = os.Stdout
+		jsonOutput = io.Discard
+	}
```
adombeck, line 179: I don't think we need to support printing JSON to stdout and a file *at the same time*. That would allow us to set `c.opts.PrintJSON = true` if `jsonOutputFile != nil` which I expect to simplify things a bit, for example the issue mentioned in https://github.com/CodeIntelligenceTesting/cifuzz/pull/1091/files#r1338563224.

internal/cmd/run/reporthandler/reporthandler.go
jochil: Okay, I removed it because I thought it would be a problem with `io.MultiWriter` but I will check again

internal/cmd/execute/execute.go
jochil: good catch, fixed it :see_no_evil:

internal/cmd/execute/execute.go
jochil: thats right. So now it gets printed to stdout only if `--json` is set or `--json-output-file` is missing. I thing this is consistent what someone would expect using the combination of these three flags

internal/cmd/execute/execute.go
jochil: I think here lies a misunderstanding, also relevant for the other comments ( as well. I implemented it in a way that `--json` and `--json-output-path` can be used separately. So if you use `--json-output-file` without `--json` the regular output on stdout and stderr will not be affected...and if I want to see json in my terminal I have to use `--json`

This is how I would expect it to work. This way you can have readable output on your terminal and still write json to a file.

internal/cmd/run/reporthandler/reporthandler.go
jochil: Ok, worked. If you use `--json` without `--json-output-file` the terminal output is colored... when using `--json-output-file` the output is not colored because of `io.Multiwriter` but I think this is good enough

internal/cmd/execute/execute.go
jochil: done

jochil:
> The output of the updating printer flickers on this branch. To reproduce:
> 
> ```
> git apply << EOF
> diff --git a/examples/cmake/src/explore_me.cpp b/examples/cmake/src/explore_me.cpp
> index 00e01a18..45c7a103 100644
> --- a/examples/cmake/src/explore_me.cpp
> +++ b/examples/cmake/src/explore_me.cpp
> @@ -14,9 +14,9 @@ void exploreMe(int a, int b, string c) {
> 
>          if (c == "FUZZING") {
>            // Trigger a heap buffer overflow
> -          char *s = (char *)malloc(1);
> -          strcpy(s, "too long");
> -          printf("%s\n", s);
> +//          char *s = (char *)malloc(1);
> +//          strcpy(s, "too long");
> +//          printf("%s\n", s);
>          }
>        }
>      }
> EOF
> make install && cifuzz -C examples/cmake run my_fuzz_test
> ```

is fixed now

internal/cmd/execute/execute.go
jochil: see https://github.com/CodeIntelligenceTesting/cifuzz/pull/1091#discussion_r1339102194

internal/cmd/execute/execute.go
jochil: My use case was to have the regular readable output on the terminal or in the container logs, while still writing to the json file.
IIRC in k8s you usually see the stdout/stderr combined and having it mixed with json might be not great if you have to go through the logs.

adombeck: But the regular readable output is the non-JSON output, which we print to the terminal anyway. The JSON output is meant for the use case that you want to process the output in some way. So why would I want that both on stdout and a file at the same time?

adombeck: > IIRC in k8s you usually see the stdout/stderr combined and having it mixed with json might be not great if you have to go through the logs.

I don't understand what you're arguing for here

jochil: So we talked about it and this is what we agreed on
```
--json = json -> stdout, plain -> stderr
--json + --json-output-file = json -> file, plain -> stderr/stdout
--json-output-file = json -> file, plain -> stderr/stdout
```

adombeck:
I pushed a refactoring commit which simplifies the code a bit. Please take a look.

jochil:
> I pushed a refactoring commit which simplifies the code a bit. Please take a look.

Nice, looks good...thanks
 I squashed your commit with the initial one and added you to the commit

adombeck (approved):

PR #1092, opened 2023-09-27
refactor(test): optimize cleanup of test helper functions

jochil:
### Motivation/Context
While writing tests for another task I noticed that the `BootstrapExampleProjectForTest` and `ChdirToTempDir` functions were using cleanup function that have to be called manually

### Description
using `t.Cleanup` instead of returning cleanup functions

### How to use/reproduce
- should have no effects


internal/testutil/testutil.go
```diff
@@ -79,17 +79,17 @@ func ChdirToTempDir(prefix string) (tempDir string, cleanup func()) { //nolint:n
...
+	t.Cleanup(func() {
 		err = os.Chdir(oldWd)
 		if err != nil {
 			log.Printf("Failed to change working directory back to %s: %+v", oldWd, err)
 			os.Exit(1)
 		}
 		ChdirMutex.Unlock()
 		fileutil.Cleanup(testTempDir)
-	}
+	})
```
adombeck, line 90: I'm not completely convinced that it's better to always run this cleanup function when the test is complete instead of letting the caller decide when to call it. For example if the same test wants to chdir into another temporary directory, calling this method again will cause a deadlock, because the `ChdirMutex` is never unlocked. Since apparently this is not the case in any our tests, I won't block on this and let you decide if you want to go ahead anyway.

adombeck (approved):

internal/testutil/testutil.go
jochil: I understand. But as this is not an actual use case for  this function I think it is fine to keep it like this... especially as it makes the current uses easier. 
Can you think about a another solution in the case someone calls it twice? Checking the Mutex and calling `t.Fatal` to make it more clear?
Of course I would be happy to change it later, if we can come up with something

PR #1097, opened 2023-10-02
fix: update package-lock.json files for nodejs/nodejs-typescript example projects

florianGla:
Updating the package-lock.json files is required to run the example projects.

### Motivation/Context

Without this change our example projects for Jazzer.js use the wrong dependencies.

### How to use/reproduce

Execute `cifuzz run FuzzTestCase` for our javascript and typescript example projects.


florianGla:
Note: I tried to run the example projects on main again and they seem to work. Can someone try to reproduce the issue on their machine as well? The node_modules need to be removed and `npm install` has to be executed before starting the fuzzing run.

jochil:
I checked `main` and everything works fine with a fresh `npm install`. Maybe the error has happened because the local `node_modules` where not updated?
Also the changes in this PR on the `package-lock.json` are not including any jazzer.js updates. So I am not sure if there is an actual problem with our examples right now

florianGla:
Thank you, I will close the PR.

PR #1099, opened 2023-10-04
Specify Jazzer.js version in instructions

florianGla:
We want to prevent that a Jazzer.js update breaks cifuzz and our test pipelines. Therefore we want to require a specific version in the instructions.


pkg/messaging/instructions/nodejs
```diff
@@ -2,10 +2,10 @@ To use jazzer.js, add a dev-dependency to @jazzer.js/jest-runner to your
 project. To do so, execute the following command:
 
     # if you use npm
-    npm install --save-dev @jazzer.js/jest-runner
+    npm install --save-dev @jazzer.js/jest-runner@2.0.0
 
     # or if you use yarn
-    yarn add --dev @jazzer.js/jest-runner
+    yarn add --dev @jazzer.js/jest-runner@2.0.0
```
adombeck, line 8: I'm not familiar with the npm/yarn update mechanism. If we specify the specific version here instead of `2.x`, how do users who ran these instructions get newer Jazzer.js releases?

florianGla: They do not get newer Jazzer.js releases. I think it makes sense to require/recommend a specific version and update the instructions together with the package.json using our update script. We also require specific versions for jazzer and our gradle plugin. Maybe it makes sense to have a discussion if we should require specific versions or major versions in general?

jochil: > They do not get newer Jazzer.js releases

That's not right. If you use `npm install --save-dev @jazzer.js/jest-runner@2.0.0`, it will create the following entry in the package.json
```
  "devDependencies": {
    "@jazzer.js/jest-runner": "^2.0.0"
  }
```
The `^` is coming from semver and means "compatible version". In our case this would be >= `2.0.0` and < `3.0.0`
see: https://docs.npmjs.com/cli/v8/configuring-npm/package-json#dependencies

IMO this is exactly what we want, or?

adombeck: > I think it makes sense to require/recommend a specific version and update the instructions together with the package.json using our update script. 

How do our users know that they need to re-run the command printed by the `cifuzz init` instructions? As far as I know we have no mechanism for that yet, right? That's also a problem for updates to a new major version.

jochil: I guess it would be the best if we add dependency checks for jazzer and jazzer.js as well (like the ones we are doing with gradle, clang, ...).
In npm it can be something like `npm list --depth=0 | grep "jazzer.js"`. 

Maybe doing this in another task? This one seems ok on its own, or?

jochil (approved):

pkg/messaging/instructions/nodejs
adombeck: Yes, if this part:

> The ^ is coming from semver and means "compatible version". In our case this would be >= 2.0.0 and < 3.0.0

is correct, then I don't want to block the PR. Please create a new task for the update mechanism / dependency check.

PR #1100, opened 2023-10-04
container: Support producing LCOV coverage report for C/C++ fuzz tests

adombeck:
**~~This is based on https://github.com/CodeIntelligenceTesting/cifuzz/pull/1102, please review that first.~~**

### Motivation/Context

We want fuzz containers to support producing LCOV reports to be able to replace `ci-coverage` in the backend.

### Description

Support `--coverage-output-path` option in `cifuzz execute` to produce an LCOV coverage report at the specified path.

### How to use/reproduce

```bash
make install && \
make build-container-image && \
rm -rf /tmp/foo && \
mkdir -p /tmp/foo && \
cifuzz -C examples/cmake container run my_fuzz_test --docker-image cifuzz-test-container-run-other:latest --bind /tmp/foo:/output -v -- -- --coverage-output-path /output/coverage.lcov && \
cat /tmp/foo/coverage.lcov
```


MarkusZoppelt:
The reproduce steps unfortunately don't work for me. I don't have the `cifuzz-test-container-run-other:latest` image. It seems to be generated from the integration tests.

Even if I generate the required image via

```
CI=1 CIFUZZ_ALLOW_UNSUPPORTED_PLATFORMS go test -v ./integration-tests/other/... -v
```
(other is unsupported on macOS)

I still can't reproduce it, because it cannot find the image, even tho it exists according to `docker images`

pkg/runner/libfuzzer/libfuzzer_runner.go
```diff
@@ -101,6 +107,23 @@ func NewRunner(options *RunnerOptions) *Runner {
...
+	// Return the error unless it's a context canceled error (caused by
+	// a terminating signal), because we still want to produce the
+	// coverage report in that case.
+	if err != nil && !errors.Is(err, context.Canceled) {
+		return err
+	}
+
+	if r.CoverageBinary != "" {
+		return r.produceCoverageReport(ctx)
+	}
```
florianGla, line 121: Why do we produce a coverage report in libfuzzer_runner? I thought running libfuzzer has nothing to do with a coverage run. I see that `cifuzz execute` has the option to produce a coverage report after executing a fuzzing run, but it feels weird to me to have that coupling between libfuzzer_runner and coverage. It is very different for our non-containerized approach. What do the other guys think?

florianGla (approved):

pkg/runner/libfuzzer/libfuzzer_runner.go
adombeck: I refactored the code and now call the coverage generator directly from the execute command.

PR #1104, opened 2023-10-05
feat: LCOV parser

mlsandnerCI:
### Motivation/Context

We want to use the `.lcov` format as our general coverage format so we need a way to parse a `jacoco.xml` into an lcov report. 

### Description

New package `pkg/coverage` includes a struct `LCOVReport` and parsers to parse an `.lcov` file into a `LCOVReport`, a `jacoco.xml` into a `LCOVReport` and a `LCOVReport` into our own `Summary` which is used to display the coverage table after a run.

`Summary` now uses the `LCOVReport` instead of reading a `.lcov` file and parsing it on it's own.

All of the files related to parsing (jacoco & summary) have been moved to the parser package.

Unit tests have been added.

### How to use/reproduce
Run coverage on one of our example java projects, e.g. maven to create a `jacoco.xml`.
```
cifuzz coverage com.example.FuzzTestCase::myFuzzTest
```

With the function `ParseJacocoXMLIntoLCOVReport()` you can read in the jacoco coverage report and convert it to a lcov report. You can write this report to a file with `WriteLCOVReportToFile()` to get the `.lcov` file. 

To see if the results are correct you can run 
```
genhtml <file> --output report
```
to create the html coverage report and compare it with the jacoco coverage report.

Note, that the .lcov file has to be in the project root so that `genhtml` can find all source files for mapping and generating the html. The output directory also has to be defined.


pkg/parser/coverage/jacoco.go
```diff
@@ -0,0 +1,211 @@
...
+
+	return lcovReport, nil
+}
+
+// ParseJacocoXMLIntoSummary takes a jacoco xml report and turns it into
+// the `Overview` struct. The parsing is as forgiving
+// as possible. It will output debug/error logs instead of
+// failing, with the goal to gather as much information as
+// possible
+func ParseJacocoXMLIntoSummary(in io.Reader) *Summary {
```
MarkusZoppelt, line 158: I see the explanation in the comment, but it feels odd that this does not return an error but `ParseJacocoXMLIntoLCOVReport` does...

MarkusZoppelt (approved):
Nice work!

pkg/parser/coverage/jacoco.go
mlsandnerCI: Definitely understand that! I discussed this with @jochil before and we decided that because `LCOVReport` will have a more important role than the `Summary` we should make it fail on errors instead, because we need the parsing to succeed to continue working with the report. If the `Summary` parsing fails, the only thing it affects is the output on the console. I didn't want to change the current functionality so that's why there is this mismatch now.

PR #1109, opened 2023-10-06
feat: switch to maven extension

jjohannes:
### Motivation/Context
https://github.com/CodeIntelligenceTesting/cifuzz-maven-extension

### Description

- Use similar means to get information directly from the build tool (Requires 'cifuzz-maven-extension')
- Improve getting the jacoco.xml path from build tool
- Fix issues with multi-projects

With this there is certain duplication of code in 'maven.go' and 'gradle.go'. Ideally, both can be combined into one integration for "Java projects". Where only the command lines to call the tools differ but the processing of information is the same for both.


jochil:
> LGTM right now but I would like to wait with my approval until all the tests pass to see if something else has to change.

@mlsandnerCI beside the flaky nodejs test, everything passed now

mlsandnerCI (approved):

PR #1112, opened 2023-10-09
 container: Support producing JaCoCo XML coverage report for Java fuzz tests

adombeck:
~~**This is based on https://github.com/CodeIntelligenceTesting/cifuzz/pull/1100, please review that first.**~~

### Motivation/Context

We want fuzz containers to support producing coverage reports to be able to replace ci-coverage in the backend. In a follow-up task, the JaCoCo XML report will be converted to LCOV.

### Description

Support `--coverage-output-path` option in `cifuzz execute` to produce a JaCoCo XML report at the specified path.

### How to use/reproduce

```bash
make install && \
make build-container-image && \
rm -rf /tmp/foo && mkdir -p /tmp/foo && \
cifuzz -C examples/maven container run com.example.FuzzTestCase::myFuzzTest --bind /tmp/foo:/output -v -- -- --coverage-output-path /output/jacoco.xml && \
cat /tmp/foo/jacoco.xml
```


adombeck:
@MarkusZoppelt I added some tests in the last commit but they need some more work:

* The Maven test fails because the `cifuzz init` command doesn't yet print the new dependencies which must be added to the pom.xml:
```xml
<dependency>
	<groupId>org.jacoco</groupId>
	<artifactId>org.jacoco.cli</artifactId>
	<version>0.8.8</version>
	<classifier>nodeps</classifier>
</dependency>
<dependency>
	<groupId>org.jacoco</groupId>
	<artifactId>org.jacoco.agent</artifactId>
	<version>0.8.8</version>
	<classifier>runtime</classifier>
	<scope>test</scope>
</dependency>
```

* Producing a coverage report in a Gradle project fails with `== Java Exception: java.nio.file.AccessDeniedException: /cifuzz/.cifuzz-corpus`. You can reproduce this with the example project:
```bash
make install && \
make build-container-image && \
rm -rf /tmp/foo && mkdir -p /tmp/foo && \
cifuzz -C examples/gradle container run com.example.FuzzTestCase::myFuzzTest --bind /tmp/foo:/output -v -- -- --coverage-output-path /output/jacoco.xml && \
cat /tmp/foo/jacoco.xml
```

I strongly suspect that the reason for that failure is that our Gradle example project (and the Gradle testdata project) still use Jazzer v0.20.1 which tries to create `.cifuzz-corpus` in the working directory. They should use v0.21.1 instead.

MarkusZoppelt:
> I strongly suspect that the reason for that failure is that our Gradle example project (and the Gradle testdata project) still use Jazzer v0.20.1 which tries to create `.cifuzz-corpus` in the working directory. They should use v0.21.1 instead.

I opened a pull request in https://github.com/CodeIntelligenceTesting/cifuzz-gradle-plugin/pull/19 to update Jazzer to 0.21.1

Makefile
```diff
@@ -169,7 +169,7 @@ fmt/check: deps/dev
 		echo -e >&2 "Unformatted files:\n$${DIFF}"; \
 		exit 1; \
 	fi;
-	npx prettier --loglevel=warn --check .
+	npx prettier --plugin=@prettier/plugin-xml --print-width=120 --xml-whitespace-sensitivity=preserve --log-level=warn --write .
```
jochil, line 172: I think you accidentally replaced `--check` with `--write` here

pkg/runner/jazzer/jazzer_runner.go
```diff
@@ -153,6 +158,164 @@ func (r *Runner) Run(ctx context.Context) error {
...
+	}
+	if jacocoCLIJar == "" {
+		return errors.New("jacococli JAR not found in class paths")
+	}
+
+	// Produce a JaCoCo XML report from the jacoco.exec file
+	cmd := executil.CommandContext(ctx, "java", "-jar", jacocoCLIJar, "report", jacocoExecFile,
+		"--xml", outputFile, "--classfiles", classFilesDir)
+	cmd.Stderr = os.Stderr
+	cmd.Stdout = os.Stdout
```
jochil, line 207: Can we use `log.Debug` for printing the command? It came handy in the past having the actual commands in the verbose output

pkg/runner/jazzer/jazzer_runner.go
```diff
@@ -349,19 +364,3 @@ func (r *Runner) FuzzerEnvironment() ([]string, error) {
 func (r *Runner) Cleanup(ctx context.Context) {
 	r.Runner.Cleanup(ctx)
 }
-
-func (r *Runner) printDebugVersionInfos(classPath string) error {
```
jochil, line 190: Why pulling this function into the already very long `Run` instead of just printing the java version here?

Makefile
adombeck: Right, thanks for catching that.

pkg/runner/jazzer/jazzer_runner.go
adombeck: Good point. Done.

Makefile
adombeck: done

pkg/runner/jazzer/jazzer_runner.go
adombeck: I extracted it again

jochil (approved):
nice one :+1: 
If we are continuing with the maven extension we should figure out if there a things (like the jacococli dependencies) we can move into the extension. But right now this is fine

PR #1114, opened 2023-10-09
feat(stacktrace): filter java stack frames with source file not in source map

florianGla:
Filter java stack frames with source file not contained in source map, until the first stack frame is found, whose source file is contained in the source map. All further stack frames are used without applying any filters.
This ensures that the first stack frame always contains a source location, which can be resolved to a local user file.

### Motivation/Context

A similar filter was implemented for CISense and this change ensures that the stack traces are filtered in the same way.

### Description

The source map is now passed to the libfuzzer runner directly. This is necessary in order for the execute command to be able to pass the source map to libfuzzer. It is read from the bundle. 

### How to use/reproduce

Start a fuzzing run for the spring boot example (https://github.com/CodeIntelligenceTesting/spring-boot-example). The first stack frame should contain a local source file.


internal/cmd/execute/execute.go
```diff
@@ -416,3 +423,27 @@ func findBinary(nameToFind string, bundleMetadata *archive.Metadata, isCoverageB
 
 	return nil, errors.Errorf("fuzzer '%s' not found in a bundle metadata file", nameToFind)
 }
+
+func getSourceMap() (*sourcemap.SourceMap, error) {
```
jochil, line 427: what do you think about moving this function to the `sourcemap` package?

jochil (approved):

PR #1119, opened 2023-10-12
fix(gradle): Display correct error message if plugin is missing

mlsandnerCI:
### Motivation/Context

If the gradle plugin task `cifuzzPrintPluginVersion` failed for any reason, we printed an error that the plugin was missing, even if the plugin was added but e.g. the project build failed.

### Description

We are now checking the output of the command that calls the `cifuzzPrintPluginVersion` task. If it can't find the task it means the plugin is missing and we return the correct error and a help message. Before this was handled in the build function itself but that caused the output to differ from our usual error messages where we first stop the progress spinner and then print the stderr output and displayed the error. Now this step is handled in the Adapter to ensure that the behavior is identical. 

Before:
Bildschirmfoto 2023-10-12 um 09 29 08

Now:
If plugin is missing:
Bildschirmfoto 2023-10-12 um 09 32 01

If project build is failing (e.g. due to error in the `build.gradle` or in my case, my OS not being able to use the gradle wrapper):
Bildschirmfoto 2023-10-12 um 09 27 45

We now also check in the gradle integration tests that the correct error message is displayed if the plugin is missing.
### How to use/reproduce

Remove the line
```
id "com.code-intelligence.cifuzz" version "1.5.0"
```
from the `build.gradle` in `example/gradle` and run cifuzz 
```
cifuzz -C examples/gradle run com.example.FuzzTestCase::myFuzzTest
```

You can also cause the project build to fail by removing the line  
```
id 'application'
```
in the `build.gradle` to generate a different error when running cifuzz.


MarkusZoppelt (commented):
**Both** described methods to invoke the new error message result in the same error message for me, i.e., _either_ removing `id "com.code-intelligence.cifuzz" version "1.5.0"` _or_ removing `id 'application'` **both** results in:

```
❯ cifuzz -C examples/gradle run com.example.FuzzTestCase::myFuzzTest
cifuzz version dev
✅ You are authenticated.
▀  Build in progress... Error.


FAILURE: Build failed with an exception.

* What went wrong:
A problem occurred configuring root project 'gradle'.
> Could not open cp_proj generic class cache for build file '/Users/mz/code/cifuzz/examples/gradle/build.gradle' (/Users/mz/.gradle/caches/7.6/scripts/dgyx2xwihefm22dsb0jmmkdvo).
   > BUG! exception in phase 'semantic analysis' in source unit '_BuildScript_' Unsupported class file major version 65

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 2s
❌ exit status 1
```

jochil (approved):

MarkusZoppelt (approved):
Works as described with Java < 21

PR #1124, opened 2023-10-16
feat(gradle): update build.gradle and instructions with JUnit deps

MarkusZoppelt:
## Description

The example projects didn't have JUnit 5 set up correctly for Gradle.

This change updates all the `build.gradle` files and the instructions for Gradle with the missing deps. We should decide if this is the preferred way to set up dependencies compared to using the Gradle plugin.


examples/gradle-kotlin/build.gradle.kts
```diff
@@ -12,6 +12,11 @@ repositories {
     mavenCentral()
 }
 
+dependencies {
+    testImplementation("org.junit.jupiter:junit-jupiter:5.10.0")
+    testImplementation("com.code-intelligence:jazzer-junit:0.21.1")
```
fmeum, line 17: The JUnit samples also recommend `useJUnitPlatform()` as well as using a BOM: https://github.com/junit-team/junit5-samples/blob/r5.10.0/junit5-jupiter-starter-gradle-kotlin/build.gradle.kts.

fmeum (commented):
I would recommend updating the Maven instructions and example in the same way.

jochil:
Did you checked with @jjohannes if there is a way to handle the junit5 setup (or at least verification) in the gradle plugin? I would like to keep the instructions as easy as possible for the user

MarkusZoppelt:
> Did you checked with @jjohannes if there is a way to handle the junit5 setup (or at least verification) in the gradle plugin? I would like to keep the instructions as easy as possible for the user

@jochil I wouldn't know how. These changes at least seem to fix CLI-1332. I will ping @jjohannes in our Slack.

jochil (approved):

fmeum:
I find it important to distinguish between fuzzing setup and general testing setup. While the fuzzing setup is cifuzz-specific and should be as short as possible, hiding the regular JUnit 5 from the user may end up making it more and not less difficult to maintain for them.

Having verification in the plugin that ensures that the general testing setup is sufficient for fuzzing sounds great to me though.

jjohannes:
Right now, the plugin automatically adds the required dependencies. You could remove all the 

```
dependencies {
   ...
}
```

blocks from the examples and they still work.

> I find it important to distinguish between fuzzing setup and general testing setup. While the fuzzing setup is cifuzz-specific and should be as short as possible, hiding the regular JUnit 5 from the user may end up making it more and not less difficult to maintain for them.

To still satisfy this, you could reduce the dependencies blocks to one line:

```
dependencies {
  testImplementation("org.junit.jupiter:junit-jupiter:5.10.0")
}
``` 

If you upgrade the examples to the latest Gradle 8 version, you could also use the new [test suites feature](https://docs.gradle.org/current/userguide/jvm_test_suite_plugin.html) which I would say is the **new** best way to setup testing:

```
testing {
  suites.named('test') {
    useJUnitJupiter()
  }
}
```
(remove `dependencies` and `tasks.test { }`, the `useJUnitJupiter()` line does all that for you)

------

Unfortunately, there is not the "one right way" to setup testing in Gradle at the moment. 😞 

So there are other options how to declare the dependencies to JUnit 5. 

Although I understand the argument for "distinguish between fuzzing setup and general testing setup", IMO these examples are not for explaining how to setup testing in Gradle in general. Users will most likely already have done that and chosen one of the many options that fits them best. So if it would be my call, I would limit these examples to the minimum you need to get fuzz testing working. Which would mean:
1. Removing the `dependencies { }` block, **or**
2. Updating the sample to Gradle 8.4 and using the _test suites_ solution

PR #1125, opened 2023-10-16
Support default dictionaries for CMake projects

florianGla:
### Motivation/Context

If a user is using dictionaries to deal with complex input formats, they are not able to configure those for a specific fuzz test. In cifuzz.yaml they can only configure a dictionary for the whole project. This is not desired and wastes a lot of CPU time.

### Description

Users shall be able to easily define dictionaries for each fuzz test. The handling should be similar to seed corpus directories. The user can define a dictionary named <fuzz-test-name>.dict in the same folder where the fuzz test is defined in the CMakeLists.txt file. This dictionary is used automatically, if no other dictionary is specified via a flag or the cifuzz.yaml config file.

### How to use/reproduce

Create a file `my_fuzz_test.dict` in the CMake example project folder and add some keywords (for example: kw1="test"). Note that dictionaries cannot be empty.


jochil:
@florianGla can you add some user facing documentation for this feature? From what I got there is no way the user can figure out this behavior without reading source code

jochil (approved):
can you add some user facing documentation for this feature? From what I got there is no way the user can figure out this behavior without reading source code

PR #1127, opened 2023-10-17
fix(mockserver): correctly bind address for container

MarkusZoppelt:
The mock server was using a wrong address on macOS (and is still not working on Windows).

This change correctly binds the address of the mock server to `host.docker.internal`.

This allows to run container-related tests that require a mock server locally on macOS.


integration-tests/shared/mockserver/server.go
```diff
@@ -45,53 +45,44 @@ func New(t *testing.T) *MockServer {
...
-		require.NoError(t, err)
-	}()
-}
-
-func (server *MockServer) StartForContainer(t *testing.T) {
-	if testing.Short() {
+func (server *MockServer) Start(t *testing.T, forContainer bool) {
+	if forContainer && testing.Short() {
 		t.Skip("skipping test in short mode (requires docker)")
 	}
```
adombeck, line 51: Could we move the `forContainer` parameter to `MockServer.New`? Or are any mock server instances used both inside and outside a container?

integration-tests/shared/mockserver/server.go
```diff
@@ -45,53 +45,44 @@ func New(t *testing.T) *MockServer {
...
+			cli, err := client.NewClientWithOpts(client.FromEnv)
+			require.NoError(t, err)
+
+			networkInspect, err := cli.NetworkInspect(ctx, "bridge", types.NetworkInspectOptions{
+				Verbose: false,
+			})
+			require.NoError(t, err)
+
+			addr = networkInspect.IPAM.Config[0].Gateway
+		}
```
adombeck, line 83: I think with this patch we could use `host.docker.internal` on all platforms:

```diff
diff --git a/e2e/docker.go b/e2e/docker.go
index 6d3adcfa..1021ac3f 100644
--- a/e2e/docker.go
+++ b/e2e/docker.go
@@ -149,7 +149,8 @@ func runTestCaseInContainer(t *testing.T, ctx context.Context, dockerClient *cli
                ctx,
                containerConfig,
                &dockerContainer.HostConfig{
-                       Binds: containerBinds,
+                       Binds:      containerBinds,
+                       ExtraHosts: []string{"host.docker.internal:host-gateway"},
                },
                nil,
```

adombeck: Source: https://docs.docker.com/engine/reference/commandline/dockerd/#configure-host-gateway-ip

MarkusZoppelt: Thanks, I will try this

MarkusZoppelt: Unfortunately, this does not work on Windows: https://github.com/CodeIntelligenceTesting/cifuzz/actions/runs/6626068132/job/17998252796?pr=1127#step:5:2066

adombeck: I'm confused now. **Inside a Docker container**, you should be able to resolve `host.docker.internal` to the host's gateway IP. Now I see that this mock server is being started on the host, **outside the Docker container**: https://github.com/CodeIntelligenceTesting/cifuzz/blob/ea693d20b73c5f417ac0d6fadcbe5db6425e4910/e2e/runner.go#L62-L62

The host doesn't know the special value `host.docker.internal` so it can't bind on the correct address. How was this supposed to work?

adombeck: Our CI doesn't run the e2e tests on macOS. Did you try it locally?

MarkusZoppelt: It does work (see passing Ubuntu e2e test and macOS local test).

> this mock server is being started on the host, outside the Docker container:

Yes, this is what we want. But we need to access the correct address on the host from within the container.

MarkusZoppelt: But your patch at least allows us to use `host.docker.internal` on both Linux and macOS successfully. So I guess I will re-disable the Windows tests and take the partial win.

adombeck: > It does work (see passing Ubuntu e2e test and macOS local test).

How? I don't understand how the host can bind to `host.docker.internal` which is a Docker internal thing

adombeck: Can you try this on your host please?

```bash
cat << EOF > /tmp/main.go
package main

import "net/http"

func main() {
        address := "host.docker.internal:8080"
        err := http.ListenAndServe(address, nil)
        if err != nil {
                panic(err)
        }
}
EOF
go run /tmp/main.go
```

MarkusZoppelt: Thanks to @jochil, I re-understood what the mock server is doing:

The mock server is always bound to `listener` and therefore to the host's address(es):
https://github.com/CodeIntelligenceTesting/cifuzz/blob/630381ffbf46eb94cc277ef52f597c65a6344b10/integration-tests/shared/mockserver/server.go#L54 

It also gets started for this listener:
https://github.com/CodeIntelligenceTesting/cifuzz/blob/630381ffbf46eb94cc277ef52f597c65a6344b10/integration-tests/shared/mockserver/server.go#L67

However, we add the `Address` field to the `server` struct to let callers know the docker address and port for calling it.
https://github.com/CodeIntelligenceTesting/cifuzz/blob/630381ffbf46eb94cc277ef52f597c65a6344b10/integration-tests/shared/mockserver/server.go#L64

adombeck: Ok, then lets please not set this address in `MockServer.Start` since it has nothing to do with starting the server. I propose we remove the `MockServer.Address` field and add new methods `MockServer.AddressOnHost()` and `MockServer.AddressInContainer()` with doc comments which can be called by the caller to figure out the address it needs to connect to. I hope that will avoid confusion in the future.

MarkusZoppelt: Done

integration-tests/shared/mockserver/server.go
```diff
@@ -45,57 +41,39 @@ func New(t *testing.T) *MockServer {
...
-	for path, handler := range server.Handlers {
-		mux.Handle(path, handler)
-	}
-
-	listener, err := net.Listen("tcp4", ":0")
-	require.NoError(t, err)
+// AddressOnHost returns the address of the mock server on the host machine
+// that local clients can use to connect to it.
+func (server *MockServer) AddressOnHost(t *testing.T) string {
+	t.Helper()
```
adombeck, line 47: I don't think calling `t.Helper()` serves any purpose here because this function doesn't use any functions which print file and line information.

MarkusZoppelt: Thx, will remove

integration-tests/shared/mockserver/server.go
```diff
@@ -45,57 +41,36 @@ func New(t *testing.T) *MockServer {
...
-		mux.Handle(path, handler)
-	}
-
-	listener, err := net.Listen("tcp4", ":0")
-	require.NoError(t, err)
-
-	server.Address = fmt.Sprintf("http://127.0.0.1:%d", listener.Addr().(*net.TCPAddr).Port)
+// AddressOnHost returns the address of the mock server on the host machine
+// that local clients can use to connect to it.
+func (server *MockServer) AddressOnHost(t *testing.T) string {
```
adombeck, line 46: You can remove the `t *testing.T` argument now. Same for `AddressInContainer`.

integration-tests/shared/mockserver/server.go
```diff
@@ -45,57 +41,36 @@ func New(t *testing.T) *MockServer {
...
+	server.listener, err = net.Listen("tcp4", ":0")
 	require.NoError(t, err)
 
-	ctx := context.Background()
-
-	networkID := "bridge"
-	// On Windows, the default network is 'nat'.
-	if runtime.GOOS == "windows" {
-		networkID = "nat"
-	}
```
adombeck, line 83: In another thread you said that using `host.docker.internal` doesn't work on Windows but here you remove the Windows specific code.

adombeck: Then lets explain that in a comment

florianGla (approved):

PR #1137, opened 2023-10-24
ci: Cancel in-progress jobs on push

adombeck:
### Motivation/Context

To avoid running tests when new commits were already pushed to the same branch. Note that currently email notifications are sent for all canceled workflows.
This behavior cannot be disabled. I'm not sure if the gain of concurrency groups is worth the spam :)

Closes CLI-1345


MarkusZoppelt (approved):
LGTM

PR #1139, opened 2023-10-25
fix: adjust Jazzer.js parser for v2.1.0

florianGla:
Adjust Jazzer.js parser for v2.1.0

### Motivation/Context

Without this change the fuzzer output of Jazzer.js v2.1.0 cannot be parsed correctly.

### Description

This PR simplifies the parsing and adjusts the error id for unhandled javascript exceptions.

### How to use/reproduce

Start a fuzzing run for the javascript or typescript example projects. You may need to run `npm i` in order to get the latest version of Jazzer.js.


pkg/parser/libfuzzer/libfuzzer.go
```diff
@@ -435,48 +444,50 @@ func (p *parser) parseAsJazzerFinding(line string) *finding.Finding {
...
-func (p *parser) determineJestErrorDetails() {
-	logs := strings.Join(p.pendingFinding.Logs, "\n")
-	_, found := regexutil.FindNamedGroupsMatch(jestPathTraveralPattern, logs)
+	result, found := regexutil.FindNamedGroupsMatch(jazzerJSExceptionErrorPattern, line)
 	if found {
-		p.pendingFinding.Details = "Path Traversal"
-		return
+		return &finding.Finding{
+			Type:    finding.ErrorTypeWarning,
+			Details: result["message"],
```
MarkusZoppelt, line 458: Why do we only add the message details for this error pattern?
Does it make sense to add it to the other finding structs as well, e.g.,

* Command Injection: `in exec(); called with 'jazzer'`
* Prototype Pollution: `Prototype of Object changed. Additional properties in object1: { 'polluted': true }`

Especially the last part in the Prototype Pollution looks like valuable information.

integration-tests/errors/nodejs/testdata/package.json
```diff
@@ -1,7 +1,7 @@
 {
 	"name": "nodejs-errors",
 	"devDependencies": {
-		"@jazzer.js/jest-runner": "^1.6.1",
```
MarkusZoppelt, line 4: This version looks very old.

Is this file not in the `updater.go`?
If not, we should add it.

MarkusZoppelt (approved):

integration-tests/errors/nodejs/testdata/package.json
florianGla: Good catch

pkg/parser/libfuzzer/libfuzzer.go
florianGla: I think adding the message details for the custom bug detectors leads to really long error messages. For java we also only add the error type to the details. The rest of the error message can be looked up in the logs.

PR #1141, opened 2023-10-25
feat(bazel): coverage supports --add-corpus flag

mlsandnerCI:
### Motivation/Context

Bazel users noticed that the `--seed-corpus` (now `--add-corpus`) flag was not usable.

### Description
The `cifuzz coverage` command for bazel now supports the `--add-corpus` flag.
The easiest way to hand the corpora to our `bazel coverage` call was to use the generated corpus directory, which gets included under the hood in bazel thanks to our cifuzz rules. To avoid adjusting these or the projects `BUILD.bazel`, we symlink the files of the user specified corpora into the generated corpus directory and remove them after the coverage report is created. 

Also fixed a missing "=" in the `bazel coverage` call when setting the environment variables.

### How to use/reproduce
Create a directory in the bazel example project with a file inside that is supposed to be used as an input, e.g. "inputs/i-am-a-test" with some text inside.

Run `cifuzz coverage` with the `--add-corpus` flag and for testing purposes add `--test_output=all` as an argument for bazel to get the full log. 
```
cifuzz -C examples/bazel coverage src:explore_me_fuzz_test --add-corpus inputs -v -- --test_output=all 
```
In the log there should be this line, showing that the replayer recognized the input.
```
==================== Test output for //src:explore_me_fuzz_test:
Replaying 'src/explore_me_fuzz_test_corpus/src-.explore_me_fuzz_test_cifuzz_corpus-i-am-a-test'
```
(I have zero idea why the path is so weird, it has nothing to do with the actual directory or the symlink so it seems to be a bazel thing)

If you didn't run `cifuzz run` before, meaning you didn't have a generated corpus, there should only be the one line above. If you did have a generated corpus, all of those inputs should be listed too.

FYI for other macOS users: At least for me the coverage command for bazel fails right now because bazel creates an empty coverage report which then of course can't be parsed into a valid lcov report. But you can still see in the logs if the corpus is added.

CIF-122


integration-tests/bazel/bazel_test.go
```diff
@@ -156,6 +162,42 @@ func testCoverageWithAdditionalArgs(t *testing.T, cifuzz string, dir string) {
...
+	// has been recognized by bazel
+	cmd := executil.Command(cifuzz, "coverage", "//src/parser:parser_fuzz_test", "--add-corpus=inputs", "--", "--test_output=all")
+	cmd.Dir = dir
+
+	// Terminate the cifuzz process when we receive a termination signal
+	// (else the test won't stop).
+	shared.TerminateOnSignal(t, cmd)
+
+	output, err := cmd.CombinedOutput()
+	fmt.Println(string(output))
```
MarkusZoppelt, line 188: We usually don't print the output in the integration tests.
If the command fails, it will show the output either way, iirc.

internal/cmd/coverage/bazel/bazel.go
```diff
@@ -35,7 +37,90 @@ type CoverageGenerator struct {
...
+		}
+		generatedCorpusBasename := "." + filepath.Base(fuzzTestPath) + "_cifuzz_corpus"
+		generatedCorpus := filepath.Join(cov.ProjectDir, filepath.Dir(fuzzTestPath), generatedCorpusBasename)
+
+		// Make sure that the generated corpus directory actually exists.
+		// If the user for any reason calls the coverage command
+		// without a prior fuzzing run, we still want this to work
+		// but also delete the directory again to not clutter up the
+		// project.
+		if _, err = os.Stat(generatedCorpus); os.IsNotExist(err) {
```
MarkusZoppelt, line 78: we have `fileutil.Exists()`

integration-tests/bazel/bazel_test.go
mlsandnerCI: Good catch, this was just a remnant from debugging 👍  
But FYI if the command fails, we do not get the error message because it is stored in `output` and not printed on the console.

internal/cmd/coverage/bazel/bazel.go
```diff
@@ -35,7 +38,94 @@ type CoverageGenerator struct {
...
 func (cov *CoverageGenerator) BuildFuzzTestForCoverage() error {
-	var err error
+	commonFlags, err := cov.getBazelCommandFlags()
+	if err != nil {
+		return err
+	}
+
+	// User defined input directories are added as symlinks to the generated
+	// corpus directory, so they can be included while creating the coverage
+	// report. We remove them again afterward.
```
florianGla, line 48:
```suggestion

	// User defined inputs are added as symlinks to the generated
	// corpus directory, so they can be included while creating the coverage
	// report. We remove them again afterward.
```

internal/cmd/coverage/bazel/bazel.go
```diff
@@ -35,7 +38,94 @@ type CoverageGenerator struct {
...
+		defer func() {
+			for _, t := range temp {
+				err = os.RemoveAll(t)
+				if err != nil {
+					log.Errorf(errors.WithStack(err), "Failed to remove '%s': %v", t, err.Error())
+				}
+			}
+		}()
+
+		// Get path to generated corpus of the fuzz test
```
florianGla, line 66: Maybe it makes sense to extract the new code block, which handles user defined inputs, to a separate function?

internal/cmd/coverage/bazel/bazel.go
```diff
@@ -35,7 +38,94 @@ type CoverageGenerator struct {
...
+				}
+
+				link := filepath.Join(generatedCorpus, filepath.Base(path))
+				err = os.Symlink(path, link)
+				if err != nil && !errors.Is(err, os.ErrExist) {
+					return errors.Wrapf(err, "Failed to symlink'%s' to '%s'", path, generatedCorpus)
+				}
+				if errors.Is(err, os.ErrExist) {
+					// TODO: don't skip input if a file with the same name already exists
+					log.Debugf("File %s already exists in corpus directory, a symlink for %s cannot be created", link, path)
```
florianGla, line 114: Is the `TODO` still relevant?

internal/cmd/coverage/bazel/bazel.go
```diff
@@ -35,7 +38,94 @@ type CoverageGenerator struct {
...
+					return nil
+				}
+
+				temp = append(temp, link)
+				return nil
+			})
+			// nolint: wrapcheck
+			if err != nil {
+				return err
+			}
```
florianGla, line 126: Maybe it makes sense to wrap the error instead of adding a linter exception.
E.g.: `return errors.WithMessagef(err, "Failed to walk through corpus directory %s", dir)`

florianGla (approved):
LGTM

internal/cmd/coverage/bazel/bazel.go
mlsandnerCI: Yes, I don't know if we will run into this use case at some point but I think a reminder could be useful 🤔

PR #1143, opened 2023-10-26
feat(coverage): replace '--seed-corpus' flag with '--add-corpus'

mlsandnerCI:
### Motivation/Context
While working on adding support for this flag for Bazel, I noticed that the term "seed-corpus" is not correct in this context and could be misleading.

### Description

The `--seed-corpus` flag has been removed for the `coverage` command and replaced with a new flag `--add-corpus`. "seed corpus" is a fuzzing term but we do not use fuzzing for coverage and the inputs in the directory specified with this flag are not used as seeds to produce new inputs but are simply used to increase the code coverage.

### How to use/reproduce

Create a directory with a file with some random text inside the cmake example project, e.g. `inputs/test` and add the directory with `--add-corpus` to the command.
```
cifuzz coverage my_fuzz_test --add-corpus inputs -v
```
You can see in the logs that the inputs directory is added to the coverage call (I added some `<<<<` to the log to highlight the line) and merged with the other corpus.
```
...
Running my_fuzz_test on corpus
🔍 Executable: /Users/msandner/IdeaProjects/cifuzz/examples/cmake/.cifuzz-build/libfuzzer/coverage/my_fuzz_test
🔍 Command: LLVM_PROFILE_FILE='/var/folders/jw/bwkg1q1x0bx_74lg_8jjd_5r0000gn/T/llvm-coverage-1055712260/output/%c%m.profraw' NO_CIFUZZ='1' "/Users/msandner/IdeaProjects/cifuzz/examples/cmake/.cifuzz-build/libfuzzer/coverage/my_fuzz_test" "-artifact_prefix=/var/folders/jw/bwkg1q1x0bx_74lg_8jjd_5r0000gn/T/llvm-coverage-1055712260/output/merge-artifacts/" "-runs=0" 
"/var/folders/jw/bwkg1q1x0bx_74lg_8jjd_5r0000gn/T/llvm-coverage-1055712260/output/empty-file-corpus"
INFO: Running with entropic power schedule (0xFF, 100).
INFO: Seed: 2842269560
INFO:        0 files found in /var/folders/jw/bwkg1q1x0bx_74lg_8jjd_5r0000gn/T/llvm-coverage-1055712260/output/empty-file-corpus
INFO: -max_len is not provided; libFuzzer will not generate inputs larger than 4096 bytes
INFO: A corpus is not provided, starting from an empty corpus
#2      INITED exec/s: 0 rss: 26Mb
#2      DONE   lim: 4 exec/s: 0 rss: 26Mb
Done 2 runs in 0 second(s)
🔍 Command: LLVM_PROFILE_FILE='/var/folders/jw/bwkg1q1x0bx_74lg_8jjd_5r0000gn/T/llvm-coverage-1055712260/output/%c%m.profraw' NO_CIFUZZ='1' 
"/Users/msandner/IdeaProjects/cifuzz/examples/cmake/.cifuzz-build/libfuzzer/coverage/my_fuzz_test" "-artifact_prefix=/var/folders/jw/bwkg1q1x0bx_74lg_8jjd_5r0000gn/T/llvm-coverage-1055712260/output/merge-artifacts/" 
"-merge=1" "/var/folders/jw/bwkg1q1x0bx_74lg_8jjd_5r0000gn/T/llvm-coverage-1055712260/output/merge-target" "/Users/msandner/IdeaProjects/cifuzz/examples/cmake/inputs"   <<<<<<<<<<<<
INFO: Running with entropic power schedule (0xFF, 100).
INFO: Seed: 2842276184
MERGE-OUTER: 1 files, 0 in the initial corpus, 0 processed earlier 
MERGE-OUTER: attempt 1
INFO: Running with entropic power schedule (0xFF, 100).
INFO: Seed: 2842284170
INFO: -max_len is not provided; libFuzzer will not generate inputs larger than 1048576 bytes
MERGE-INNER: using the control file '/var/folders/jw/bwkg1q1x0bx_74lg_8jjd_5r0000gn/T//libFuzzerTemp.Merge32931.txt'
MERGE-INNER: 1 total files; 0 processed earlier; will process 1 files now
#1      pulse  exec/s: 0 rss: 26Mb
#1      DONE   exec/s: 0 rss: 26Mb
MERGE-OUTER: successful in 1 attempt(s)
MERGE-OUTER: the control file has 90 bytes
MERGE-OUTER: consumed 0Mb (27Mb rss) to parse the control file
MERGE-OUTER: 0 new files with 0 new features added; 0 new coverage edges
```


internal/cmd/coverage/coverage.go
```diff
@@ -39,14 +39,14 @@ type Generator interface {
...
-	NumBuildJobs   uint     `mapstructure:"build-jobs"`
-	SeedCorpusDirs []string `mapstructure:"seed-corpus-dirs"`
-	UseSandbox     bool     `mapstructure:"use-sandbox"`
+	OutputFormat string   `mapstructure:"format"`
+	OutputPath   string   `mapstructure:"output"`
+	BuildSystem  string   `mapstructure:"build-system"`
+	BuildCommand string   `mapstructure:"build-command"`
+	CleanCommand string   `mapstructure:"clean-command"`
+	NumBuildJobs uint     `mapstructure:"build-jobs"`
+	CorpusDirs   []string `mapstructure:"corpus-dirs"`
```
adombeck, line 48: This has the effect that the `cifuzz coverage` command ignores the `seed-corpus-dirs` setting in the `cifuzz.yaml`. We do want the `cifuzz coverage` command to use the input files from those directories though.

I think the best solution is to keep the name `seed-corpus` for the option and better document what those directories are used for in the usage message of `cifuzz coverage`.

mlsandnerCI: @CodeIntelligenceTesting/cifuzz-devs I need your opinion on this since Adrian is not able to discuss this anymore.

I still think using `seed-corpus` for coverage is confusing, especially with the link we provide in the description of the option in the `cifuzz.yaml` (https://llvm.org/docs/LibFuzzer.html#corpus) which only explains the word in context of fuzzing. IMO it is acceptable for the user to add the directories they want to use as inputs for coverage the same way as for fuzzing and adjusted our `cifuzz.yaml` template to include a section for `corpus-dirs` and tested that the directories are used when calling coverage.

jochil: I agree that `seed-corpus` might be confusing in the context of `coverage`... so maybe we can make it more clear in the `cifuzz.yaml`  that `seed-corpus-dirs` is used for a fuzzing run only? 

An alternative could also be just renaming `seed` for `run` too

mlsandnerCI: I adjusted the documentation to specify that `seed-corpus-dirs` is only used for fuzzing runs 👍 I think CS would like us to keep the word "corpus", they also wanted it for the new flag.

jochil (approved):

PR #1147, opened 2023-10-30
Validate node.js fuzz tests

florianGla:
### Motivation/Context

After starting a fuzzing run for a Node.js project with an invalid fuzz target, either no error message or a generic npx error message was printed. This is confusing for users.

### Description

An error message is printed in case no fuzz test or multiple fuzz tests are found.

### How to use/reproduce

Run the Node.js example project with an invalid test name pattern or an invalid test path pattern.
For example: `cifuzz run BuzzTestCase` or `cifuzz run FuzzTestCase:'My buzz test'`


pkg/options/node.go
```diff
@@ -18,6 +19,13 @@ func JazzerJSTestPathPatternFlag(value string) string {
 	return JazzerJSTestPathPattern + fmt.Sprintf("='%s'", value)
 }
 
+func JazzerJSReportersFlag(value string) string {
+	if value == "" {
+		return JazzerJSReporters
+	}
+	return JazzerJSReporters + fmt.Sprintf("='%s'", value)
+}
+
```
MarkusZoppelt, line 28: Why was this reporters flag added?
Can you explain why it is needed in a comment? Thanks!

pkg/options/node.go
```diff
@@ -6,6 +6,7 @@ import (
 
 const JazzerJSTestNamePattern string = "--testNamePattern"
 const JazzerJSTestPathPattern string = "--testPathPattern"
+const JazzerJSReporters string = "--reporters"
```
MarkusZoppelt, line 9: Why was this reporters flag added?
Can you explain why it is needed in a comment? Thanks!

florianGla: I explained it in cmdutils/node.go, where it is used.
`// disable jest reporters to prevent unnecessary output`
`args = append(args, options.JazzerJSReportersFlag(""))`

Do you think that this is enough? I'm not sure if comments are needed in the node options. These are just helper functions to wrap the flags.

florianGla: You could also use the reporters flag to customize the test result output.

MarkusZoppelt (approved):

PR #1151, opened 2023-11-06
feat: convert jacoco xml reports to `.lcov` format for Maven/Gradle

MarkusZoppelt:
### Motivation/Context

Our common coverage format is `.lcov`.

### Description

Calls the `ParseJacocoXMLIntoLCOVReport` function to generate a `.lcov` file from a jacoco XML report (coming from jacocoCLI).

### How to use/reproduce

    make install && make build-container-image
    cifuzz -C examples/maven container run com.example.FuzzTestCase::myFuzzTest --bind /tmp/foo:/output -v -- -- --coverage-output-path /output/bar.lcov
    cat /tmp/foo/bar.lcov


pkg/runner/jazzer/jazzer_runner.go
```diff
@@ -195,7 +195,12 @@ func (r *Runner) ProduceJacocoReport(ctx context.Context, outputFile string) err
 	cmd.Stderr = os.Stderr
 	cmd.Stdout = os.Stdout
 	log.Debugf("Command: %s", strings.Join(stringutil.QuotedStrings(cmd.Args), " "))
-	return cmd.Run()
+	err = cmd.Run()
+	if err != nil {
+		return "", err
```
mlsandnerCI, line 200:
```suggestion
		return "", errors.WithStack(err)
```

internal/cmd/execute/execute.go
```diff
@@ -315,7 +316,32 @@ func (c *executeCmd) run(metadata *archive.Metadata) error {
...
+		}
+		defer fileReader.Close()
+
+		// ... and convert it to an LCOV report.
+		lcovReport, err := jacocoCoverage.ParseJacocoXMLIntoLCOVReport(fileReader)
+		if err != nil {
+			return err
+		}
+		// we don't need the jacoco.xml file anymore
+		defer os.Remove(jacocoXMLFile)
```
mlsandnerCI, line 337:
```suggestion
defer func() {
    err = os.Remove(jacocoXMLFile)
		if err != nil {
			log.Errorf(err, "Failed to remove intermediate jacoco.xml report: %v", err)
    }
}()
```
Just to make sure we know if this fails. It could also be a `log.Debugf` if you prefer to keep it "hidden".

MarkusZoppelt: Thx, updated!

mlsandnerCI (approved):

PR #1158, opened 2023-11-07
ci: cache maven and gradle dependencies

florianGla:
It can take up to ~15 minutes for Maven and Gradle just to download the project dependencies. We should cache those in order to speed up the tests and maybe to prevent occasional timeouts.

I have used the recommended steps from: https://github.com/actions/cache/blob/main/examples.md#java---gradle


.github/actions/vm_setup/action.yml
```diff
@@ -21,6 +21,24 @@ runs:
           ./llvm
         key: ${{ inputs.os }}-llvm-14.0
 
+    - name: Cache Gradle dependencies
+      uses: actions/cache@v3
+      with:
+        path: |
+          ~/.gradle/caches
+          ~/.gradle/wrapper
+        key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties') }}
```
jochil, line 30: shouldn't this be `${{ inputs.os }}`... I am not sure if `{{ runner.os }}` is available here
Same for the maven step

jochil (approved):

.github/actions/vm_setup/action.yml
florianGla: Yes, you are right. Good catch!

PR #1160, opened 2023-11-07
feat(coverage): enable `--format lcov` for maven/gradle

MarkusZoppelt:
### Motivation/Context

Previously, only `jacocoXML` and `html` was supported as an output format for Maven/Gradle projects.
But because we have a jacocoXML-to-lcov parser now, we can also support `--format lcov` for Maven/Gradle.

### Description

Adds `lcov` as a supported output format to coverage for buildsystems Maven / Gradle.

### How to use/reproduce

    cifuzz coverage -C examples/maven com.example.FuzzTestCase::myFuzzTest --format lcov


internal/cmd/coverage/gradle/gradle.go
```diff
@@ -85,13 +85,32 @@ func (cov *CoverageGenerator) GenerateCoverageReport() (string, error) {
...
+	}
+
 	parser.ParseJacocoXMLIntoSummary(reportFile).PrintTable(cov.Stderr)
 
-	if cov.OutputFormat == coverage.FormatJacocoXML {
+	switch cov.OutputFormat {
+	case coverage.FormatJacocoXML:
 		return filepath.Join(cov.OutputPath, "jacoco.xml"), nil
+	case coverage.FormatLCOV:
+		return filepath.Join(cov.OutputPath, "coverage.lcov"), nil
```
mlsandnerCI, line 110: If the `reportPath` gets returned in line 101, why is this necessary?

mlsandnerCI: Andf don't we want to print a summary in the case of lcov?

MarkusZoppelt: Both fixed 👍🏼

internal/cmd/coverage/maven/maven.go
```diff
@@ -118,13 +118,44 @@ pom.xml can disrupt the JaCoCo execution and have to be prefixed with '@{argline
...
+		defer reportFile.Close()
+
+		summary, err := parser.ParseLCOVReportIntoSummary(reportFile)
+		if err != nil {
+			return "", err
+		}
+		summary.PrintTable(cov.Stderr)
+
+	} else {
+		parser.ParseJacocoXMLIntoSummary(reportFile).PrintTable(cov.Stderr)
```
jochil, line 148: This is the same code as in the `gradle.go` file, or? WDYT about moving it to a reusable function to avoid the duplication?

MarkusZoppelt: Good idea, thx. I moved the relevant code to a new [`shared.HandleLCOV`](https://github.com/CodeIntelligenceTesting/cifuzz/blob/CLI-1356-java-coverage-lcov/internal/cmd/coverage/shared/lcov.go#L14) func.

jochil (approved):

internal/cmd/coverage/maven/maven.go
jochil: thanks. One minor thing: the name :) 
IMO something like `convertToLCOV` would be easier to get

MarkusZoppelt: Done. I even found a better way to generalize the function.
It now returns a summary and the `.PrintTable` gets called on the summary in both cases by the caller.

PR #1170, opened 2023-11-14
feat(dialog): improve project selection

MarkusZoppelt:
### Motivation/Context

Users wanted an indication that the project list in `dialog.Select()` is 1) longer and 2) scrollable.

Another request was to group projects by organisations.
It would be helpful to have the organisation listed in `dialog.Select()` as well (either as a new column or as in-between headlines) (tbd).

The list was sorted case-sensitive (naive ASCII).
Users reported it would make more sense to sort in a case-insensitve manner.

### Description

- feat(dialog): increase max amount of shown entries in Select() from 5 (default) to 15
- feat(dialog): sort entries in Select() case-insensitively
- feat(dialog): improve project selection prompt

### How to use/reproduce

Screenshot of the new prompt UI:
image

`Showing t out of m projects` adapts to the shown/total projects available with a `t <= 13` (15 - minus the entries for `<<new>>` and `<<cancel>>`)

**Update:** _The dialog prompt only shows the enhanced prmpt when the list is indeed scrollable (entries `t > 13`)_


florianGla (approved):
LGTM

PR #1171, opened 2023-11-14
feat(coverage): java coverage generator

mlsandnerCI:
### Motivation/Context

Two bugs where discocered during the development of generating a coverage report in containers:
1) The generated coverage report from the container differed from the locally generated reports if there was more than 
one fuzz test in the project.
2) In a project with two fuzz tests, running the coverage command first on one and then the other changed the results for the first fuzz test.

It seems that the coverage reports generated with maven and gradle locally always produce an aggregated report of all fuzz tests but the reports generated with the jacoco cli only covered the specified fuzz test.
After some discussion it was decided, that the local reports should also be generated with the jacoco cli to get consistency.

### Description

This is a bigger PR, I apologize already. But it makes sense to keep all these changes together and a lot of code was deleted.

1) The cifuzz installations now provides the jacoco JARs `org.jacoco.agent` and `org.jacoco.cli` (version 8.11) and they are also included in the bundle and containers as runtime dependencies (similar to the list-fuzz-test JAR).
2) The code to produce a coverage report with the jacoco CLI was moved (and adjusted) from the `execute` command to the `coverage` command.
3) Maven, Gradle and the container command all use the same code to produce a coverage report. Maven and Gradle now function exactly the same which means that differences like Gradle saving the html report in a directory `html` but Maven just in the directory with all other files, have been eradicated.
4) The original maven and gradle coverage generators were removed.
5) Setting up a maven project now doesn't rely on the jacoco plugin anymore so the instructions and all pom.xml were adjusted. 
6) Since we now call jazzer when generating the coverage report, maven and gradle can now support engine args. This change was necessary to pass the `TestAdditionalJazzerParameters` test because this checks for jazzer calls and if they respect flags set via `--engine-arg`.

### How to use/reproduce
In a maven or gradle project with >=2 fuzz tests, call the `cifuzz coverage` command for each fuzz test and you should get different reports.
Coverage in the container command should work as before, you can test it with:
```
make install
make build-container-image
cifuzz container run -C examples/maven com.example.FuzzTestCase::myFuzzTest --bind <any/path/you/want>:/output -v -- -- --coverage-output-path /output/report 
```

CIF-124
CIF-125


florianGla:
We should maybe mention in the coverage command help that the `--build-jobs` flag is only used for C/C++ projects?
I'm not sure if we should do that in this PR. What do you think?

florianGla:
We currently test in our integration tests for maven and gradle the output formats jacocoxml and html. Do you think it makes sense to add a test case for lcov as well?

florianGla (approved):
LGTM

PR #1174, opened 2023-11-20
feat(api): deprecate usage of project_external_id in v3 calls

MarkusZoppelt:
### Motivation/Context

`project_external_id` bad. `nid` good.
CIF-281

### Description

Deprecates usage of `project_external_id` for `v3` use.

### How to use/reproduce

    go test ./internal/api -run ^TestProjectNameToNID$

Make @zgtm happy.


internal/api/api.go
```diff
@@ -146,17 +147,32 @@ func ConvertProjectNameForUseWithAPIV1V2(projectName string) string {
...
 
-// ConvertProjectNameForUseWithAPIV3 converts a project name to the v3 API
-// format. The API format is <project-name>, where <project-name> is URL
-// unescaped.
-func ConvertProjectNameForUseWithAPIV3(projectName string) (string, error) {
-	projectName = strings.TrimPrefix(projectName, "projects/")
+// ConvertProjectNameToProjectNid converts a project name to the v3 API
+// project NID. The project NID format is 'prj-XXXXXXXXXXXX', where 'X' is
+// any alphanumeric.
+func ConvertProjectNameToProjectNid(projectName string) (string, error) {
```
jochil, line 153: sounds a bit like a name from a java library ;) WDYT about `ProjectNameToNid`?

internal/api/api.go
```diff
@@ -146,17 +147,32 @@ func ConvertProjectNameForUseWithAPIV1V2(projectName string) string {
...
-	projectName, err := url.QueryUnescape(projectName)
-	if err != nil {
-		return "", errors.WithStack(err)
+	if strings.HasPrefix(projectName, "projects/") {
+		projectName = strings.TrimPrefix(projectName, "projects/")
 	}
-	return projectName, nil
+	if strings.HasPrefix(projectName, "projects%2F") {
+		projectName = strings.TrimPrefix(projectName, "projects%2F")
+	}
```
jochil, line 161: maybe this can be avoided with a `url.QueryUnescape` at the beginning of this function? I can also imagine unforeseen problems with the regex later on having an escape URL

jochil (approved):

internal/api/api.go
jochil, line 153: This should be easy to cover with a simple unit test, testing some weird project names

MarkusZoppelt: Fair point. Disgusting!

MarkusZoppelt: Good catch!
There was actually a "bug(?)" in the function!

@zgtm I added 11e106735019e2db986330f746958c22c27b438b that passes the test, but I am not sure if this is the way to go... regexes... maybe you can also have a look. Thx!

zgtm: @jochil @MarkusZoppelt Thanks for catching this! I replaced 11e106735019e2db986330f746958c22c27b438b with 1222d1aea0b4216c8f0e1bf89d1694d2ecba2a22. The NIDs format should not be relied on, but we know for sure that project external-ids have a dash and 8 hex-digits. I just forgot to force the checking of the full length string.

Apparently my commit message does not satisfy the gods. But the tests pass now, as far as I can see, so feel free to fixup my commit into the original change! :)

internal/api/api_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+	// test NID without prefix
+	nid, err = ProjectNameToNID("prj-ow6h1UIwHXTr")
+	require.NoError(t, err)
+	assert.Equal(t, expectedNID, nid)
+
+	// test NID with prefix
+	nid, err = ProjectNameToNID("projects/prj-ow6h1UIwHXTr")
+	require.NoError(t, err)
+	assert.Equal(t, expectedNID, nid)
```
jochil, line 48: This screams for table driven tests: https://dave.cheney.net/2019/05/07/prefer-table-driven-tests

jochil (approved):

PR #1175, opened 2023-11-20
Use local error details

florianGla:
### Motivation/Context

We want to ship the error details with cifuzz in order to offer a self contained solution, which does not rely on a connection to CI Sense.

### Description

The error-details.json file is now located in the share folder of the cifuzz installation directory.

### How to use/reproduce

Start a fuzzing run without being logged in to CI Sense.


internal/cmdutils/auth/auth.go
```diff
@@ -52,7 +52,7 @@ func GetValidToken(server string) (string, error) {
 		return "", err
 	}
 	if !isValid {
-		return "", &NoValidTokenError{errors.New("Please log in with a valid API access token")}
```
MarkusZoppelt, line 55: Why don't we return an error here? The function name suggests otherwise.

MarkusZoppelt (approved):

PR #1181, opened 2023-11-21
feat: remove not authenticated warnings

florianGla:
### Motivation/Context

cifuzz shall be used locally without a connection to CI Sense without being warned that a connection is required for commercial usage.

### Description

All warnings are removed.

### How to use/reproduce

Start a fuzzing run without a valid CI Sense token.


MarkusZoppelt (approved):
Nice cleanup!

PR #1184, opened 2023-11-22
chore: use download portal in install script

HenrichN:
### Motivation/Context

CI Fuzz releases move to Azure blob storage and will only be available via our download portal with a valid token.
This PR adjusts the install.sh script to account for that.


install.sh
```diff
@@ -1,29 +1,36 @@
 #!/bin/sh
 set -e
 
-BASE_URL="https://github.com/CodeIntelligenceTesting/cifuzz/releases/latest/download/"
+# Check
+if [ -z "$1" ]
+  then
+    echo "No token supplied. Head over to downloads.code-intelligence.com to generate one."
```
MarkusZoppelt, line 7:
```suggestion
    echo "No token supplied. Head over to https://downloads.code-intelligence.com to generate one."
```

MarkusZoppelt (commented):
This script now needs an argument (`$1`), so we need to also update the curl command in our README.md.

HenrichN:
Deleting this PR as the file was transferred to the download portal itself. When CI Fuzz goes private, we can delete the install.sh from here.

HenrichN:
Deleting this PR as the file was transferred to the download portal itself. When CI Fuzz goes private, we can delete the install.sh from here.

PR #1196, opened 2023-11-28
feat(nodejs): support setting engine args

mlsandnerCI:
### Motivation/Context

We ignore the `--engine-arg` flag for NodeJS projects right now.

### Description

Because flags for libfuzzer can't be added to the `npx jest` command for Jazzer.js to recognize them, we now set them in the environment variable `JAZZER_FUZZER_OPTIONS`. The flag `-timeout` is filtered out and separately set in `JAZZER_TIMEOUT`. Additional logic to prioritize values in the `.jazzerjsrc` of the project has been added as well as unit tests.

### How to use/reproduce

Run the nodejs project with engine arguments in verbose mode to see that the environment variables are set for the fuzzing command and that libfuzzer receives the flags.

```
cifuzz run -C examples/nodejs FuzzTestCase --engine-arg=-help=1 --engine-arg=-seed=1 --engine-arg=-timeout=2000 -v
```


florianGla (approved):
LGTM

PR #1197, opened 2023-11-28
build: use private `dlth` mod

MarkusZoppelt:
### Motivation/Context

We want to use the new features (e.g. candidates) from our own private `dlth` module.

### Description

This PR adds the `dlth` mod to the project.
It requires `GOPRIVATE=github.com/CodeIntelligenceTesting/dlth` to be set, which relevant `make` targets will now automatically set.

### How to use/reproduce

If you have set up to clone / pull from GitHub via ssh (recommended), you need to set this in your `.gitignore`:

```git
[url "ssh://git@github.com/"]
	insteadOf = https://github.com/
```


MarkusZoppelt:
Won't do this. We'll copy `dlth` into cifuzz later.

PR #1201, opened 2023-12-01
feat: fetch registry auth from CI Sense

MarkusZoppelt:
**_Endpoint `v2/docker_registry/authentication` is currently only available on `staging`._**

### Motivation/Context

Users should get the credentials for our azure container registry via CI Sense.

### Description

Queries and uses credentials for our default container registry.
Removed the `--registry` flag from the `remote-run` cmd.

### How to use/reproduce

    cifuzz container remote-run \
      -C examples/maven com.example.FuzzTestCase::myFuzzTest \
      --server app.staging.code-intelligence.com -v


internal/container/auth.go
```diff
@@ -12,6 +12,8 @@ import (
 	"github.com/pkg/errors"
 )
 
+// RegistryAuth returns the base64 encoded auth config for the given registry from
+// the docker config file, e.g. ~/.docker/config.json.
 func RegistryAuth(registry string) (string, error) {
```
MarkusZoppelt, line 17: This function will now be unused - shall I delete it entirely? We might need it in the future...

jochil: I would delete it... why keeping dead code ;)

jochil (approved):

PR #1208, opened 2023-12-06
ci: using the new install script from the download portal

jochil:
### Motivation/Context
Some pipelines are using the install script to receive the current release of cifuzz. We are switching to the download portal, so we need to get the latest version with the new install script

### Description
switching from `sh -c "$(curl -fsSL https://raw.githubusercontent.com/CodeIntelligenceTesting/cifuzz/main/install.sh)"` to 
`sh -c "$(curl -fsSL https://downloads.code-intelligence.com/assets/install.sh)" ${{ token }} latest`
### How to use/reproduce
-


.github/actions/compatibility_check/action.yml
```diff
@@ -4,6 +4,10 @@ inputs:
   os:
     description: "current vm operating system"
     required: true
+  download_token:
+    description: "token for the download portal"
+    required: true
+
```
MarkusZoppelt, line 10: why do this instead of using `secrets.DOWNLOAD_TOKEN` like everywhere else? 🤔

jochil: This is an action and an action does not have access to (repository) secrets... see also https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions#accessing-your-secrets
> To make a secret available to an action, you must set the secret as an input or environment variable in the workflow file

jochil:
@bertschneider I experienced strange problems with handing over secrets from a reusable workflow to an action... the secret was only available after I added `secrets: inherit` when including the workflow. You did the same thing in the same workflow before with
```
      - name: Setup VM for cifuzz
        uses: ./.github/actions/vm_setup
        with:
          repository_user: ${{ secrets.REPOSITORY_USER }}
          repository_token: ${{ secrets.REPOSITORY_TOKEN }}
```
And I wonder how this worked. I debugged the inputs of the vm_setup action a bit and it looked like they are empty (mistakes on my site not ruled out). Any ideas?

bertschneider:
Hmm, didn't had that problem. Like you mentioned, the initial workflow had access to the secrets and passed them on as input arguments to our actions. That being said, I'm currently investigating a problem in the `main push` pipeline that looks similar.

jochil:
> The old install script is still used in `internal/container/ensure-cifuzz.sh` and `README.md`.

I would consider this out-of-scope as the task is about the pipelines only... but I created other tasks to handle these points:

- https://code-intelligence.atlassian.net/browse/CIF-269
- https://code-intelligence.atlassian.net/browse/CIF-270

bertschneider (approved):

PR #1212, opened 2023-12-07
feat(container): monitor container remote runs

MarkusZoppelt:
### Motivation/Context

In order to replace our internal `cictl` tool, cifuzz needs to be able to monitor container remote runs.

### Description

This change now starts the monitoring after a `cifuzz container remote-run ...`
It will keep running until the remote run finishes, times out, `monitor-duration` passes, or a remote finding is reported.

The *monitor duration* is configurable via `--monitor-duration`.

### How to use/reproduce

    cifuzz container remote-run \
    -C examples/maven com.example.FuzzTestCase::myFuzzTest \
    --registry cifuzzcontainers.azurecr.io/cifuzz-dev/maven \
    --monitor-duration 69s


internal/api/container_run.go
```diff
@@ -102,3 +102,45 @@ func (client *APIClient) PostContainerRemoteRun(image string, project string, fu
...
+
+type ContainerRemoteRun struct {
+	Nid    string `json:"nid"`
+	Status string `json:"status"`
+}
+
+// GetContainerRemoteRunStatus gets the status of a container run from the CI
+// Sense API at /v3/runs/{run_nid}.
+func (client *APIClient) GetContainerRemoteRunStatus(runNID string, token string) (*ContainerRemoteRunStatus, error) {
+	var response ContainerRemoteRunStatus
```
jochil, line 118: a bit nitpick , but I would declare a variable close to their first usage (L140). Otherwise you have to scroll up to figure out what kind of variable that is or where it is coming from when you have to read/understand the code

internal/api/api.go
```diff
@@ -345,14 +345,23 @@ func (client *APIClient) sendRequestWithTimeout(method string, endpoint string,
 	req.Header.Add("Accept", "application/json")
 	req.Header.Add("Content-Type", "application/json")
 
-	log.Debugf("Sending HTTP request: %s %s\n%s", method, endpoint, body)
+	// Don't log requests to get the status of a run because they are sent every
+	// 5 seconds
+	if !(method == "GET" && strings.Contains(endpoint, "v3/runs/")) {
+		log.Debugf("Sending HTTP request: %s %s\n%s", method, endpoint, body)
+	}
```
jochil, line 352: that might be confusing... if I am using `--verbose` I want to see all request, otherwise I might be tricked into thinking the request are not happening at all (eg. when debugging stuff)

internal/api/finding.go
```diff
@@ -104,6 +126,44 @@ func (client *APIClient) DownloadRemoteFindings(project string, token string) (F
...
 }
 
+// RemoteFindingsForRun uses the v3 API to download all findings for a given
+// (container remote-)run.
+func (client *APIClient) RemoteFindingsForRun(runNID string, token string) (Findings, error) {
+	remoteFindings := Findings{}
+
+	uri, err := url.JoinPath("v3", "runs", runNID, "findings")
+	if err != nil {
+		return remoteFindings, errors.WithStack(err)
```
jochil, line 136: in this case I would explicitly return `nil` instead a variable that you think is `nil` at the point (which might change for whatever reason). Also it is common in go to just do `return nil, err` in the case of an error

internal/api/finding.go
```diff
@@ -104,6 +126,44 @@ func (client *APIClient) DownloadRemoteFindings(project string, token string) (F
 	return remoteFindings, nil
 }
 
+// RemoteFindingsForRun uses the v3 API to download all findings for a given
+// (container remote-)run.
+func (client *APIClient) RemoteFindingsForRun(runNID string, token string) (Findings, error) {
```
jochil, line 131: From a higher view this looks similar to `GetContainerRemoteRunStatus` and maybe other functions in the `api` package. 
Basically it is always: here are some params, build a path, make a GET|POST request and unmarshall it into a variable
Maybe we can refactor this into a reusable function using generics, wdyt?

internal/cmdutils/pflag.go
```diff
@@ -181,6 +181,16 @@ func AddInteractiveFlag(cmd *cobra.Command) func() {
 	}
 }
 
+func AddMonitorDurationFlag(cmd *cobra.Command) func() {
+	cmd.Flags().Duration("monitor-duration", 0,
+		"Duration of the monitoring of container remote runs, e.g. \"30m\", \"1h\".\n"+
+			"The default is to run indefinitely.\n"+
+			"cifuzz will only start monitoring if this flag is provided.")
```
jochil, line 197: this is a bit confusing for me... so there is a default (indefinitely) but you still have to provide the flag to start monitoring?

jochil (changes requested):
I would appreciate adding a few tests... this is a complete new feature and there is not a single test in this PR

internal/api/api.go
MarkusZoppelt: I know... I removed the logging for this specific case, because the messages were just too much for a run, that takes, e.g., multiple minutes. It's then basically a full terminal screen exclusively filled with these messages and you lose context of the "regular" end of the command

internal/cmdutils/pflag.go
MarkusZoppelt: This was a specific request by our CS team. They want these three scenarios covered:

1. no `--monitor-duration` provided: no monitoring
2. `--monitor-duration` provided (with or without `0` value) => monitor forever
3. `--monitor-duration duration` => monitor run for `duration`

MarkusZoppelt:
> I would appreciate adding a few tests... this is a complete new feature and there is not a single test in this PR

We have a separate ticket for tests and a special testcase in CIF-19

jochil:
> > I would appreciate adding a few tests... this is a complete new feature and there is not a single test in this PR
> 
> @jochil We have a separate ticket for tests and a special testcase in [CIF-10](https://code-intelligence.atlassian.net/browse/CIF-10)

that ticket reads like a special test case, but why shipping this feature without any basic tests? IMO we should do this directly in the feature PR if there are no very good reasons to not do it.

The monitoring itself looks fine :+1:

internal/api/finding.go
MarkusZoppelt: We can think of a way to refactor this but I would not like to introduce generics for this... from what I've seen, they are pretty weird in Go.

internal/api/api.go
MarkusZoppelt: Especially in CI/CD, this will be incredibly chatty and @HenrichN was eager to make it non-verbose in that case.

internal/cmdutils/pflag.go
jochil: Especially the 2. case feels strange... setting a duration flag without setting a value

WDYT about having a `--monitor` flag that turns monitoring off/on and an additional flag for setting the time out?
So for example
```
cifuzz run ... --monitor # forever
cifuzz run ... --monitor --monitor-timeout=1h
```

This would also leave room for something like
`cifuzz run --monitor --monitor-interval=30s`

internal/api/finding.go
jochil: I agree that they are not easy to get in the first place, but this is more for writing the function... using the function is easy and would enables us to have a nice, slick interface for the `APIClient`
Imagine having 10+ API endpoints... do you want to have 10+ functions doing basically the same?

I can help with that, of course

internal/api/api.go
jochil: Changing the interval might help a bit with this.
`--verbose` is a technical log... we should not intentionally hide information there. I understand that showing all of the requests in the log is distracting, but showing none at all can not be the solution... at least from my point of view.
If I (or someone else) had to debug something for this feature and I will not see the request in the output it will lead them to a different route and waste time

MarkusZoppelt: Then let's print all of them and wait for feedback.
I don't like this if either...

internal/api/finding.go
MarkusZoppelt: Added a refactor commit with a new generic function `GenericAPIRequest`

internal/cmdutils/pflag.go
MarkusZoppelt: Done

jochil (approved):

PR #1215, opened 2023-12-08
feat: add reproduce command for other build systems

florianGla:
### Motivation/Context

Currently the only way to test if an issue does not occur after a fix is to manually configure the debugger with a crashing input or let the fuzzer run and cancel the fuzzer after some time and hope that the crashing input was generated and tested against the program. This PR solves this problem for other build systems with a new command which allows to reproduce a local or remote finding.

### Description

The reproduce command first tries to find the finding locally. If it does not exist, it tries to find the finding remotely and adds the finding data to .cifuzz-findings. The finding input has to be written to a new crashing-input file and has to be set as the finding input file. This is necessary, because we can only pass input files to the fuzz test binary. After that the corresponding fuzz test is build and run with the reproducing input file.

### How to use/reproduce

Run `cifuzz reproduce` in our other example project with a local or a remote finding.


internal/api/finding.go
```diff
@@ -68,43 +70,43 @@ type Severity struct {
 }
 
 // DownloadRemoteFindings downloads all remote findings for a given project from CI Sense.
-func (client *APIClient) DownloadRemoteFindings(project string, token string) (Findings, error) {
+func (client *APIClient) DownloadRemoteFindings(project string, token string) (*Findings, error) {
```
MarkusZoppelt, line 73: If you change the return Type to `*Findings`, you can return nil instead of an empty list.
I do the same change in #1212 (4b3d6f732820ccae21868645e77c06c179b6fda5)

internal/cmd/reproduce/reproduce.go
```diff
@@ -0,0 +1,316 @@
...
+
+	cBuildResult, err := builder.Build(fuzzTest)
+	if err != nil {
+		return nil, err
+	}
+
+	return cBuildResult, nil
+}
+
+func (c *reproduceCmd) getFinding() (*findingPkg.Finding, error) {
```
MarkusZoppelt, line 233: maybe `storeFinding` or `persistFinding` is a better name?

internal/cmd/reproduce/reproduce.go
```diff
@@ -0,0 +1,316 @@
...
+		err = finding.Save(c.opts.ProjectDir)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return finding, nil
+}
+
+func (c *reproduceCmd) getRemoteFinding() (*findingPkg.Finding, error) {
```
MarkusZoppelt, line 254: `selectRemoteFinding` maybe is a better name?
`get` is always so generic and non-explicit.

integration-tests/other/other_test.go
```diff
@@ -324,6 +329,30 @@ func testContainerRun(t *testing.T, cifuzzRunner *shared.CIFuzzRunner) {
...
+	env := cifuzzEnv(dir)
+	cmd := executil.Command(cifuzz, "reproduce", "-v",
+		"--build-command", buildCommand(), findings[0].Name)
+	cmd.Dir = dir
+	cmd.Env, err = envutil.Copy(os.Environ(), env)
+	require.NoError(t, err)
+
+	t.Logf("Command: %s", envutil.QuotedCommandWithEnv(cmd.Args, env))
+	output, err := cmd.CombinedOutput()
+	log.Printf("xxxxxxxxxxxxxxxxxxxxxx %s", string(output))
```
MarkusZoppelt, line 347: This can be deleted I assume

MarkusZoppelt (approved):
Nice job!

internal/cmd/reproduce/reproduce_test.go
```diff
@@ -0,0 +1,175 @@
...
+
+	// install cifuzz
+	testutil.RegisterTestDepOnCIFuzz()
+	installDir := shared.InstallCIFuzzInTemp(t)
+	cifuzz := builder.CIFuzzExecutablePath(filepath.Join(installDir, "bin"))
+
+	testutil.BootstrapExampleProjectForTest(t, "reproduce-cmd-test-", config.BuildSystemOther)
+
+	// setup mock server
+	t.Setenv("CIFUZZ_API_TOKEN", "token")
```
MarkusZoppelt, line 158: I think you can use 
```suggestion
	t.Setenv("CIFUZZ_API_TOKEN", mockserver.ValidToken)
```

PR #1217, opened 2023-12-08
feat(maven): print overridden jazzer version

florianGla:
### Motivation/Context

We want to print a warning if the user manually sets the jazzer version and doesn't use the one provided by our maven extension. This can lead to unexpected issues.

### Description

Use new maven extension to detect jazzer versions.
Print warning in cifuzz, e.g. "Overriding default Jazzer version with version x.y.z".

### How to use/reproduce
Add the following snipped to the pom.xml of our maven example project and start a fuzzing run.
```xml
<dependencies>
	<dependency>
		<groupId>com.code-intelligence</groupId>
		<artifactId>jazzer-junit</artifactId>
		<version>0.21.1</version>
		<scope>test</scope>
	</dependency>
	<dependency>
		<groupId>org.junit.jupiter</groupId>
		<artifactId>junit-jupiter-engine</artifactId>
		<version>5.9.2</version>
		<scope>test</scope>
	</dependency>
</dependencies>
```


jochil:
@florianGla what do you think about changing the wording from `manually set` to `overwriting default`? IMO this is what is actually happening: The user overwrites the default jazzer version shipped by the plugin.

internal/build/java/maven/maven.go
```diff
@@ -219,3 +220,23 @@ func GetSourceDir(projectDir string) (string, error) {
 
 	return "", nil
 }
+
+func GetOverwrittenJazzerVersion(projectDir string) string {
```
MarkusZoppelt, line 224: The "Effective Go" documentation discourages the "Get" prefix in Getter functions:
https://go.dev/doc/effective_go#Getters

internal/build/java/java.go
```diff
@@ -49,3 +49,11 @@ func RootDirectory(projectDir string, buildSystem string) (string, error) {
...
 	return projectDir, nil
 }
+
+func CheckOverwrittenJazzerVersion(projectDir string, buildSystem string) {
+	if buildSystem == config.BuildSystemGradle {
+		// not yet implemented
+	} else if buildSystem == config.BuildSystemMaven {
+		maven.GetOverwrittenJazzerVersion(projectDir)
+	}
+}
```
MarkusZoppelt, line 59: The technically correct term would be to **_"override"_** instead of **_"overwrite"_**.

To "overwrite" something is to put something else in its place, destroying the thing overwritten. To "override" something is to cause something else to operate instead of it without harming or changing the thing overridden

MarkusZoppelt (approved):
I left two nits, the rest LGTM.

internal/build/java/maven/maven.go
florianGla: You are right. Most function in build/java/maven and build/java/gradle use the "Get" prefix. Do you think it makes sense to rename them all?

jochil: > The "Effective Go" documentation discourages the "Get" prefix in Getter functions:
https://go.dev/doc/effective_go#Getters

I felt into this trap before, too... this refers to Getter, so methods on a struct. It does not cover regular functions.
IMO we can keep the `Get` prefix in this case. Skipping the `Get` in a struct makes more sense as the context is clear, you are getting something from the struct... in a regular function it is nice to make it more explicit

florianGla:
> I am not sure about the implementation details... Usually we have something in the `dependency` package extracting the version for a specific dependency. In this case we do it in another way
> 
> Was there a reasoning to do it like this?

I think it is not really a dependency check. We just check if the user has added a jazzer dependency to his/her pom.xml file overriding the one provided by the maven plugin and print a warning. We don't have an explicit dependency check for jazzer or other dependencies provided by our maven/gradle extensions. We usually put all functions, that communicate with our maven/gradle extensions, in build/java/maven or build/java/gradle. @bertschneider what do you think?

jochil:
And does this change with this task coming up? https://code-intelligence.atlassian.net/browse/CIF-266

florianGla:
> And does this change with this task coming up? https://code-intelligence.atlassian.net/browse/CIF-266



> And does this change with this task coming up? https://code-intelligence.atlassian.net/browse/CIF-266



> And does this change with this task coming up? https://code-intelligence.atlassian.net/browse/CIF-266

I don't think so. Shall me maybe have a short meeting? @bertschneider was the main author driving this change.

florianGla:
> And does this change with this task coming up? https://code-intelligence.atlassian.net/browse/CIF-266

I don't think so. Shall me maybe have a short meeting? @bertschneider was the main author driving this change.

jochil (approved):

PR #1221, opened 2023-12-12
feat(gradle): adjust multi project check

mlsandnerCI:
### Motivation/Context

`cifuzz init` checks if a gradle project is a multi project build to print out a warning that cifuzz should be setup in the subprojects. Our original workflow checked for the existence of a settings.gradle but this is not correct because even single project builds can have this file. 

### Description

~~`IsGradleMultiProject()` now checks in all of the found `settings.gradle` if they have the line "include(...)" which defines sub projects (https://docs.gradle.org/current/userguide/writing_settings_files.html). The tests have been adjusted.~~

`IsGradleMultiProject()` was renamed to `WarnIfGradleModuleProject()` which now checks all the `settings.gradle` files in the project if they contain the line "include(...)" which defines sub projects and prints a warning in that case. The message was adjusted to handle false positives.

### How to use/reproduce

Run `cifuzz init` in a gradle project with a `settings.gradle` file and the warning should only be printed if sub projects are defined in it.


internal/config/config.go
```diff
@@ -228,7 +230,19 @@ func IsGradleMultiProject(projectDir string) (bool, error) {
...
 	}
-	return true, nil
+
+	for _, match := range matches {
+		bs, err := os.ReadFile(match)
+		if err != nil {
+			return false, errors.WithStack(err)
+		}
+
+		if strings.Contains(string(bs), "include(") {
```
fmeum, line 240: I don't think it's possible to make a check on the level of string inclusion reliable. 

What do we need the information for? The Gradle plugin or, in the worst case, the Gradle CLI would probably be a better place to get this information from.

bertschneider: Good point. `include` could also be defined in other contexts, e.g. plugin configurations, haven't thought about that. IMHO it's quite unlikely, but we should use a more robust approach if possible.

Gradle provides a `projects` task listing all subprojects, but that only works on the root level. Is cifuzz aware of the project root level?

@mlsandnerCI proposed to rephrase the message to fit all use-cases. Looking at the implications here, I tend towards that as well :smile:

florianGla: We have a gradle task, which prints the project root level. Why don't we add a task to our gradle plugin to decide if the project is a multi module project?

fmeum: I this method really only used to decide whether to show this message?
https://github.com/CodeIntelligenceTesting/cifuzz/blob/b23a2c34832d992f49b679c14cbd59c2b018b278/internal/cmd/init/init.go#L24

If so, I support the idea of rephrasing that message to make it less confusing in case of a "false positive" and keeping the `strings.Contains` logic. In that case, how about we rename it to `WarnIfGradleModuleProject` and move the log statement into it? That makes it less likely that this function will be relied upon for any future "production" logic, which would definitely be problematic with a string-based approach.

mlsandnerCI: Sounds like a good approach to me, @fmeum  👍 The only use case is the warning message.

bertschneider (approved):
LGTM

fmeum (approved):

PR #1222, opened 2023-12-12
ci: sign macOS installers with rcodesign

MarkusZoppelt:
### Motivation/Context

macOS installers were unsigned and were flagged by macOS Gatekeeper.

### Description

Uses [`rcodesign`](https://github.com/indygreg/apple-platform-rs) to sign macOS executables.

### How to use/reproduce

Download a signed installer binary and verify its signature:

    rcodesign verify cifuzz_installer_macOS_arm64


.github/workflows/workflow_installer.yml
```diff
@@ -42,6 +42,24 @@ jobs:
           download_token: ${{ secrets.DOWNLOAD_TOKEN }}
           os: ${{ matrix.os }}
 
+      - name: Sign macOS installers
+        if: contains(matrix.os, 'mac')
+        run: |
+          curl \
+            -L https://github.com/indygreg/apple-platform-rs/releases/download/apple-codesign%2F0.26.0/apple-codesign-0.26.0-macos-universal.tar.gz \
```
fmeum, line 49: The content at this URL can change at any time and start stealing our certificate. I would recommend pinning the hash at least for downloads from GitHub releases.

MarkusZoppelt: Wonderful suggestion, thank you!
Added a shasum check

.github/workflows/workflow_installer.yml
```diff
@@ -42,6 +42,32 @@ jobs:
...
+            --output rcodesign.tar.gz
+
+          SHA256SUM=$( shasum -a 256 filename | grep -Eo '^[^ ]+')
```
fmeum, line 52: `shasum` has a `--check` flag that we could use here, see https://security.stackexchange.com/a/219325 for an example (for Linux, but should be adaptable to macOS).

MarkusZoppelt: Done

fmeum (approved):
Looks good to me.

We may want to audit the other actions used by the installer workflow for their download hygiene.

PR #1238, opened 2024-01-02
fix: using default image with all needed dependencies for C/C++ projects

jochil:
### Motivation/Context
Currently we are using `ubuntu:rolling` as our default image for bundling and containers. Unfortunately this will cause problems as `llvm-symbolizer` is not present in these images and therefore even the examples are not working when using `cifuzz container run` without defining an image

see also: https://codeintelligence.slack.com/archives/C064KJQUJ10/p1703087675155069

### Description
- using a custom image as default with the required dependencies: https://hub.docker.com/r/cifuzz/cifuzz-ubuntu
- removing some unnecessary Dockerfiles in favor of using the new image

### How to use/reproduce
Before this change `cifuzz container run my_fuzz_test -C examples/cmake` will fail, afterwards it should run without problems


bertschneider (approved):
LGTM

Should the image only be updated once in a while or is there an update process in place?

jochil:
> Should the image only be updated once in a while or is there an update process in place?

Good point... I added a commit with a new pipeline. You can see it in action here: https://github.com/CodeIntelligenceTesting/cifuzz/actions/runs/7395896939/job/20119973977

I set it up to run every monday morning, WDYT?

.github/workflows/pipeline_default_image.yml
```diff
@@ -0,0 +1,32 @@
...
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: Build and push ubuntu based image
+        uses: docker/build-push-action@v5
+        with:
+          context: docker/cifuzz-ubuntu
+          push: true
+          tags: cifuzz/cifuzz-ubuntu:latest
```
bertschneider, line 32: Should we add a tag based on the timestamp or the like as well, so that users could pin the base image?

bertschneider (approved):
That should ensure up to date images, thx!

.github/workflows/pipeline_default_image.yml
jochil: I thought about it, but it is based on `ubuntu:rolling` so I am not sure about the value of sticking to an old version of this.
And one could also argue that if the user comes to the point they need a specific version of something to build their application/system, they should/will not use our default image at all.

PR #1242, opened 2024-01-04
fix(deps): incorrect version detection for gradle

jochil:
### Motivation/Context
The version detection for gradle was not working correctly, for example identifying 6.3 as 6.0

### Description
Fixing the used regex and added some test cases

### How to use/reproduce
`go test ./go test ./pkg/dependencies/... -v` with the old regex or installing gradle 6.3 and running the broken cifuzz version


bertschneider (approved):
LGTM

PR #1245, opened 2024-01-05
feat: update Gradle plugin to autoconfigure JUnit

bertschneider:
### Motivation/Context

The CI Fuzz Gradle extension already automatically added required JUnit 5 dependencies to the project, but did not configure it as test engine. 

### Description

With this change the extension also adds the required JUnit 5 configuration, so that fuzz tests can be executed only by adding the extension to the project.

### How to use/reproduce

Add the extension to a Gradle project without explicit JUnit 5 configuration and execute a fuzz test via the IDE.


examples/gradle-kotlin/build.gradle.kts
```diff
@@ -17,18 +17,6 @@ repositories {
 	mavenCentral()
 }
 
-dependencies {
```
fmeum, line 20: I fear that we will lose test coverage for projects with existing JUnit 5 setups this way. Could we keep one project around in which we set up JUnit 5 according to the official sample repositories?

jochil: Can we have actual test cases for this?

jochil:
@bertschneider I think the following file is missing the gradle plugin update `tools/dependency-bundler/bundle-dependencies.sh`
Maybe adding it to the `updater.go` as well?

examples/gradle-kotlin/build.gradle.kts
bertschneider: Good point. I reverted the Kotlin example back to include the JUnit 5 setup based on the official example project from https://github.com/junit-team/junit5-samples/blob/r5.10.1/junit5-jupiter-starter-gradle-kotlin/build.gradle.kts.

Coverage works with and without explicit JUnit 5 setup.

@jochil The integration tests already contain coverage tests. Should I add different ones?

bertschneider:
> @bertschneider I think the following file is missing the gradle plugin update tools/dependency-bundler/bundle-dependencies.sh
Maybe adding it to the updater.go as well?

The file is already included in `updater.go`, but I missed to update the corresponding regex to support `dev` as well. Good catch, fixed, thx!

examples/gradle-kotlin/build.gradle.kts
jochil: I guess it makes sense to have an explicit test case to cover the concern that @fmeum raised, as the examples are not meant to be used inside our tests (besides some E2E pipeline tests (but on a very high level)) and covering different project setups seems reasonable in our domain.
maybe the tests under `e2e-tests/` provide an easy way to tackle this... I would be happy to help

jochil:
@bertschneider after this PR https://github.com/CodeIntelligenceTesting/cifuzz/pull/1246 was merged it is necessary to run the updater again to make sure `pkg/dependencies/definitions.go` is updated too

jochil (approved):

examples/gradle-kotlin/build.gradle.kts
bertschneider: Added e2e coverage tests.

PR #1246, opened 2024-01-05
feat: check maven extension/gradle plugin as dependency

mlsandnerCI:
### Motivation/Context

To make sure that the user uses the correct maven extension/gradle plugin version we wanted to add them as a dependency for the commands.

### Description

Major: 
- Installation and version of the gradle plugin and maven extension are now checked like all other dependencies during validation. The checks are using the `printPluginVersion` tasks from the respective plugin/extension. 
- The gradle integration tests have been adjusted to check for the new error message if the plugin is missing and the maven integration test has this check now too

Minor: 
- If a dependency is not installed, we now link to our official documentation instead of the github docs
- Error message for dependency check (`Unable to get current version for ... message: ...`) has the `message:` part removed because it is unneccesary
- Removed unused code in the maven/gradle source code
- Ordered the dependencies alphabetically because I think it looks neater and it is easier to search for a dep/function (this is just a personal pet peeve from me, if somebody doesn't like this approach, I can revert it)

### How to use/reproduce

Run the example projects to see the log message that the plugin/extension was checked. You can also remove/change the version of the plugin/extension from the `build.gradle`/`pom.xml` to check the error messages.

```
cifuzz run -C test/projects/gradle/app com.example.TestCases::remoteCodeFuzzTest -v
```
```
cifuzz run -C test/projects/maven/util util.UtilTest::joinFuzzTest -v
```


internal/build/java/gradle/gradle.go
```diff
@@ -293,3 +255,28 @@ func GetOverriddenJazzerVersion(projectDir string) string {
 	}
 	return jazzerVersion
 }
+
+func RunGradlePrintPluginVersionTask(projectDir string) (bool, string, error) {
```
jochil, line 259: This function name is a bit different from the other functions (`GetDependencies`, `GetOverriddenJazzerVersion`)... As a caller I don't want to know how it achieves getting the version number (calling gradle task), I just want the version number :)

internal/build/java/gradle/gradle.go
```diff
@@ -293,3 +255,28 @@ func GetOverriddenJazzerVersion(projectDir string) string {
...
+
+		log.Debugf("Command: %s", cmd.String())
+		_, writeErr := stderr.Write(stderr.Bytes())
+		if writeErr != nil {
+			log.Errorf(errors.WithStack(writeErr), "Failed to write command output to stderr: %v", writeErr.Error())
+		}
+		return false, "", errors.WithStack(err)
+	}
+
+	return true, string(output), nil
```
jochil, line 281: I am not sure if we need the bool indicating that the plugin is there... every time it is `false`, the version string is set to `""`. 
And one could argue that the plugin missing is an error as well
Especially if we are ignoring it in the `gradlePluginVersion` function below (`version.go`)
`_, output, err := gradle.RunGradlePrintPluginVersionTask(projectDir)`

pkg/dependencies/definitions.go
```diff
@@ -198,6 +198,18 @@ var deps = Dependencies{
 			return dep.checkFinder(dep.finder.GradlePath)
 		},
 	},
+	GradlePlugin: {
+		Key:        GradlePlugin,
+		MinVersion: *semver.MustParse("1.11.0"),
```
jochil: I think we have to update this to `1.12.0` already :)
Maybe we can add this to the `updater.go` as well

internal/build/java/maven/maven.go
```diff
@@ -222,3 +223,26 @@ func GetOverriddenJazzerVersion(projectDir string) string {
 
 	return jazzerVersion
 }
+
+func RunMavenPrintPluginVersionTask(projectDir string) (bool, string, error) {
```
jochil, line 227: Same as with the naming of the gradle function

internal/build/java/maven/maven.go
```diff
@@ -222,3 +223,26 @@ func GetOverriddenJazzerVersion(projectDir string) string {
...
+		if writeErr != nil {
+			log.Errorf(errors.WithStack(writeErr), "Failed to write command output to stderr: %v", writeErr.Error())
+		}
+		return false, "", errors.WithStack(err)
+	}
+
+	if len(output) == 0 {
+		return false, "", nil
+	}
+	return true, string(output), nil
```
jochil, line 247: See the comment on the bool for the gradle equivalent

internal/build/java/gradle/gradle.go
mlsandnerCI: Makes sense 👍

pkg/dependencies/definitions.go
mlsandnerCI: > Maybe we can add this to the updater.go as well

What do you mean with this exactly?

jochil: We have this tool to update all our internal dependencies: `tools/updater/updater.go`. Everytime there is a new version of jazzer, gradle-plugin, ... you can just run `go run ./tools/updater/updater.go --dependency gradle-plugin --version 1.12.0`
And I guess it makes sense to include this min version requirement into the updater as well.

jochil (approved):

PR #1247, opened 2024-01-05
fix(coverage): include Java source file contents in coverage report

fmeum:
### Motivation/Context

Java coverage reports without source file contents do not allow users to see which lines are covered.

### Description

Java coverage reports now include source file contents, which is achieved by passing `--sourcefiles` to JaCoCo.

### How to use/reproduce

`cifuzz coverage` on the Maven example project


internal/cmd/coverage/java/java.go
```diff
@@ -78,8 +78,11 @@ func (cov *CoverageGenerator) GenerateCoverageReport() (string, error) {
 		classFilesDir = filepath.Join(cov.ProjectDir, "build", "classes")
 	}
 
+	// TODO: Get the correct path from the build system.
+	sourceFilesDir := filepath.Join(cov.ProjectDir, "src", "main", "java")
```
jochil, line 82: You should be able to do this with the `SourceDirs` function from this package: `internal/build/java/java.go`
But this might make the coverage generation depending on the build system again... which is something we wanted to avoid for the fuzz container :thinking:  
Maybe giving the source directories into this function as a parameter and allowing it to be empty/optional. The `coverage` command itself can then find out the source directories if the build system is available, and if not... we just skip adding the source path (bundle/container case).
@mlsandnerCI WDYT?

fmeum: I mostly did it this way because we already have another hard-coded reference here:
https://github.com/CodeIntelligenceTesting/cifuzz/blob/5c63442bb6741166fb1705fd658eddac78a3255d/pkg/parser/coverage/jacoco.go#L79

As source files aren't available in fuzz containers, this would only have to work locally. I can look into getting the right path from the build system.

jochil: Ok, would be great if we can get rid of both hard-coded references... we already have a customer that renamed his source directory. Thanks :pray: 
IIRC the existing functions (and the respective tasks in the grade/maven plugin) can return multiple source directories... just to have that in mind

@mlsandnerCI WDYT about adding this case to our gradle example in `test/projects`: Having a project with a renamed source directory?

fmeum: I replaced the hard-coded references.

internal/cmd/coverage/java/java.go
```diff
@@ -78,8 +79,16 @@ func (cov *CoverageGenerator) GenerateCoverageReport() (string, error) {
...
 	}
 
-	// TODO: Get the correct path from the build system.
-	sourceFilesDir := filepath.Join(cov.ProjectDir, "src", "main", "java")
+	sourceFilesDirs, err := java.SourceDirs(cov.ProjectDir, cov.BuildSystem)
+	if err != nil {
+		return "", err
+	}
+	if len(sourceFilesDirs) == 0 {
+		return "", errors.Errorf("Failed to find source file directory in %s", cov.ProjectDir)
```
jochil, line 87: I don't think we should leave here with an error. The coverage command is used inside fuzz container remote runs, where no build system is available. In this case the report without source code information is ok and expected, we just want the plain lcov report for the backend

internal/cmd/coverage/java/java.go
mlsandnerCI: > @mlsandnerCI WDYT about adding this case to our gradle example in test/projects: Having a project with a renamed source directory?

Good idea, since we already have this for our maven project 👍 I will create a ticket.

internal/cmd/coverage/java/java.go
fmeum: Is there a way to tell whether the current run is a fuzz container coverage run? In that case, we probably shouldn't even attempt to ask the build system in the first place. Not failing if the output isn't useful in the case of a local run could easily end up hiding bugs.

jochil: I guess you can see this happing in the cifuzz execute command:
```
cd examples/gradle
cifuzz bundle com.example.FuzzTestCase::myFuzzTest
mkdir bundle-test
tar -xvf com.example.FuzzTestCase_myFuzzTest.tar.gz -C bundle-test 
cd bundle-test
cifuzz execute com.example.FuzzTestCase::myFuzzTest -v --coverage-output-path .
```

fmeum: I confirmed this with your test case, the new error check appears to be only in the path taken with `cifuzz coverage`, not `cifuzz execute`.

jochil: Ok, great :+1:

jochil (approved):

PR #1248, opened 2024-01-05
feat: handle missing network connections

florianGla:
### Description
If an access token is configured but no connection to CI Sense is available, the `run` and `findings` commands used to crash. Now we show a warning instead.
If an invalid token is configured we used to show a warning for the `run` and `findings` commands. Now we exit with an error message.
In case of a missing connection to CI Sense, new error messages are printed for `login`, `reproduce`, `remote-run` and `container remote-run`.

### How to use/reproduce

Run `cifuzz -C examples/cmake run my_fuzz_test` with an invalid token with/without an internet connection.
Run `cifuzz -C examples/cmake remote-run my_fuzz_test`  without an internet connection.


internal/api/api.go
```diff
@@ -405,13 +405,6 @@ func (client *APIClient) IsTokenValid(token string) (bool, error) {
 	// TOOD: Change this to use another check without querying projects
 	_, err := client.ListProjects(token)
 	if err != nil {
-		var apiErr *APIError
-		if errors.As(err, &apiErr) {
-			if apiErr.StatusCode == 401 {
-				log.Warnf("Invalid token: Received 401 Unauthorized from server %s", client.Server)
-				return false, nil
-			}
-		}
```
MarkusZoppelt, line 411: Why should we not handle this explicitly? Doesn't it help us debug issues?

internal/cmdutils/auth/auth.go
```diff
@@ -52,12 +52,17 @@ func GetValidToken(server string) (string, error) {
...
-	if err != nil {
+	if err != nil || !isValid {
+		var apiError *api.APIError
+		if errors.As(err, &apiError) {
+			log.Warnf(
+				"API access token for server %s is not valid. Use cifuzz login to get a valid token.",
+				server,
+			)
+			return "", &NoValidTokenError{errors.New("Invalid token")}
+		}
```
MarkusZoppelt, line 63: Does this handle all cases where the 401 check was doing so? Because it is removed in `api/api.go`

internal/cmd/login/login.go
```diff
@@ -88,6 +88,10 @@ func (c *loginCmd) run() error {
 
 		isValid, err := auth.IsValidToken(c.opts.Server, token)
 		if err != nil {
+			var connectionError *api.ConnectionError
+			if errors.As(err, &connectionError) {
+				return errors.New("No connection to CI Sense")
```
MarkusZoppelt, line 93: This message is misleading, imho. `No connection to CI Sense` reads like users are not connected to CI Sense, not that the network might be unreachable (which, in this case, it might, because it is the login cmd)

Maybe:

```go
    return errors.New("No connection to CI Sense - Network is unreachable")
```

MarkusZoppelt (commented):
I think we need to change the warning strings, see comments

internal/cmd/login/login.go
florianGla: Good point. I will change that.

MarkusZoppelt (commented):
Do you think it makes sense to distinguish between the cases

1) Network is reachable, but CI Sense is not connected
2) Network is unreachable

both in tests and the command logic?

Only if that makes sense to distinguish.

florianGla:
> Do you think it makes sense to distinguish between the cases
> 
> 1. Network is reachable, but CI Sense is not connected
> 2. Network is unreachable
> 
> both in tests and the command logic?
> 
> Only if that makes sense to distinguish.

Currently it makes no difference for our behavior. Shall we maybe start an internal discussion and create a separate ticket?

MarkusZoppelt (approved):

PR #1253, opened 2024-01-08
feat(ci): update internal deps

jochil:
### Motivation/Context
We would like to have automatic updates of our internal dependencies

### Description
- updater is able to figure out the latest version now
- Adding a pipeline to to regular checks for new dependencies (this should be changed to a push-approach in the future)

### How to use/reproduce
You can test the updater locally via
```
export GITLAB_UPDATER_TOKEN=<updater token>
go run ./tools/updater --dependency jazzer 
```


tools/updater/updater.go
```diff
@@ -112,28 +113,25 @@ func updateJazzerNpm(path string, version string) {
 	cmd.Dir = path
 	cmd.Stderr = os.Stderr
 	cmd.Stdout = os.Stdout
-	err := cmd.Run()
-	if err != nil {
-		log.Error(err)
-		os.Exit(1)
-	}
+	handleErr(cmd.Run())
```
MarkusZoppelt, line 116: Personal taste, but I'd prefer:

```suggestion
    err := cmd.Run()
    handleErr(err)
```

tools/updater/updater.go
```diff
@@ -112,28 +113,25 @@ func updateJazzerNpm(path string, version string) {
...
 
 	s := re.ReplaceAllString(buildFile, fmt.Sprintf(`${1}%s${3}`, version))
 
 	err = os.WriteFile(path, []byte(s), 0x644)
+	handleErr(err)
+
+	fmt.Printf("updated %s to %s\n", path, version)
+}
+
+func handleErr(err error) {
```
MarkusZoppelt, line 132: I like the idea of an error handler function. Maybe we can adopt this for more errors? I've already seen this a bunch of times in other codebases.

Not in scope for this PR of course, but I had the thought for a while now...

tools/updater/gitlab.go
```diff
@@ -0,0 +1,81 @@
...
+	err = json.Unmarshal(body, &packages)
+	handleErr(err)
+
+	// find latest version
+	var versions semver.Collection
+	for _, pkg := range packages {
+		versions = append(versions, semver.MustParse(pkg.Version))
+	}
+	sort.Sort(versions)
+	return versions[len(versions)-1].String()
```
MarkusZoppelt, line 80: I am not sure what's happening here... why do we return the last element?

Can you confirm if my assumption is correct:

`https://gitlab.code-intelligence.com/api/v4/projects/89/packages` expects a parameter for the dependency kind and returns a list(?) of semver versions —which, for whatever reason, is unsorted.

We then return the last element because it is the most recent semver?

tools/updater/updater.go
jochil: IMO it is okay for small stuff... but the problem is that you delegate the error handling to another function and you are not able to react to it in the actual code anymore, especially if you want a conditional return in the caller function. But maybe we can come up with something for `cifuzz`

tools/updater/gitlab.go
jochil: btw the API returns more than just the version: https://docs.gitlab.com/ee/api/packages.html. And of course gitlab is not aware that these are semver versions

MarkusZoppelt (approved):

PR #1259, opened 2024-01-08
chore(deps): updating jazzer dependency in list-fuzz-tests tool

jochil:
### Motivation/Context
We want to stay up-to-date with our internal dependencies

### Description
updating the `list-fuzz-tests` tool to use the private repository as well

### How to use/reproduce
you can just build cifuzz with the changed `tools/list-fuzz-tests/pom.xml`: `make install`


bertschneider (approved):
LGTM :tada:

PR #1261, opened 2024-01-08
ci: upload dependencies bundle in release process to Azure

bertschneider:
### Motivation/Context

We shall provide a way for CS to download all bundled dependencies via the download portal.

### Description

In the cifuzz release process, also bundle all dependencies, using the existing script, and upload the resulting artifact alongside the installers.


bertschneider:
Is there an easy way to test this workflow without merging it first?

jochil:
> Is there an easy way to test this workflow without merging it first?

You can create a small test pipeline triggering just the new workflow on `push`?

bertschneider:
> You can create a small test pipeline triggering just the new workflow on push?

Ah, nice. I wasn't aware that a push trigger works without merging it to `main` first. Thx for the tip!

Bundling the dependencies seems to work, see https://github.com/CodeIntelligenceTesting/cifuzz/actions/runs/7458062259, overall the release integration can only be tested in an actual release.

MarkusZoppelt (commented):
LGTM, but just to be sure: Can you point me to a link to a successful pipeline run or how to reproduce this? Easier to verify if it's actually working. Thx!

HenrichN (changes requested):
It would be great if we could rename the file to "dependency_bundle.tar"

bertschneider:
@HenrichN done

@MarkusZoppelt I created a debug pipeline, but removed it already from the PR. An old run can be found at https://github.com/CodeIntelligenceTesting/cifuzz/actions/runs/7458062259/job/20291314317
Should I create another one?

HenrichN (approved):
Go go go!

PR #1267, opened 2024-01-10
fix(container): create valid image names for fuzz containers

MarkusZoppelt:
### Motivation/Context

cifuzz creates image names based on the project name during cifuzz container remote-run (e.g. `fuzzcstaging.azurecr.io/cifuzz/prj-tSg6cwlCgIil:bbe7c...f0132`). This causes issues when running the fuzz container since only lower case letters are supported in the docker repository name and fails in CI Sense with the error message `couldn't parse image name <image_name> invalid reference format: repository name must be lowercase`.
### Description

Creates valid image names for fuzz containers by using the last 12 characters of a `crypto/sha256` hash of the CI Sense project name.

### How to use/reproduce

Run `cifuzz container remote-run -v` with an arbitrary CI Sense project and see that the image name is lowercase only.


simonresch (approved):
Looks good. Thanks!

PR #1271, opened 2024-01-11
fix: check for nil pointer of severity

MarkusZoppelt:
@simonresch knows what's up


simonresch:
Looks good. Thanks for the fix!

simonresch (approved):

PR #1274, opened 2024-01-12
feat(remoterun): allow to bundle multiple fuzz tests for container remote run

florianGla:
### Motivation/Context

A container remote-run shall behave the same as a non-container remote-run.

### Description

First I tried to do a larger refactoring in order to be able to pass a list of resolved fuzz tests to the bundler. Unfortunately this change is rather complicated and I think it ok to let the bundler return the fuzz tests.

### How to use/reproduce

Start a container remote run for a project with multiple fuzz tests. It may be necessary to use `app.staging.code-intelligence.com` as the remote server.


MarkusZoppelt (commented):
Very nice and concise PR, great job!

I think we need to add an e2e test to check for the multiple fuzz tests, tho

MarkusZoppelt (approved):
Nice!

PR #1278, opened 2024-01-15
feat(execute): add dependency check

florianGla:
### Motivation/Context

Dependency checks are missing for the `execute` command. This can lead to confusing error messages.

### Description

This PR adds dependency checks for `cifuzz execute`

### How to use/reproduce

The easiest way is to look at the added test cases.


internal/cmd/execute/execute.go
```diff
@@ -481,3 +486,26 @@ func findBinary(nameToFind string, bundleMetadata *archive.Metadata, isCoverageB
...
 	return nil, errors.Errorf("fuzzer '%s' not found in a bundle metadata file", nameToFind)
 }
+
+func checkDependencies(fuzzer *archive.Fuzzer) error {
+	var deps []dependencies.Key
+	switch fuzzer.Engine {
+	case "JAVA_LIBFUZZER":
+		deps = []dependencies.Key{
+			dependencies.Java,
+		}
```
MarkusZoppelt, line 496: Can / should we also check for Maven (+plugin) / Gradle (+plugin)?


I saw that we have them defined here:
https://github.com/CodeIntelligenceTesting/cifuzz/blob/7811a7f3e1b5e6992408649982adff094c6fc810/pkg/dependencies/deps.go#L31-L35

florianGla: The execute command runs a bundle, which contains only compiled jar files. Maven and Gradle is not needed.

MarkusZoppelt (approved):

PR #1279, opened 2024-01-15
feat: improve error messages for invalid dependencies

mlsandnerCI:
### Motivation/Context

I was unhappy with the output in case of missing/invalid dependencies, especially if the install check for gradle failed, and the mix between warnings/errors and that we print a stack trace where it is not needed .

### Description
- Fixed output for gradle if checking for the plugin failed
- Improved the error output for missing/invalid dependencies
- Not printing link to install docs if version is wrong, since we already say what version is needed in the error message and the docs don't tell the user more than that
- No stack trace for the dependency check error
- Stop the dependency check if one install or version check failed

**Invalid version:**
Before:
Bildschirmfoto 2024-01-15 um 15 29 39
Bildschirmfoto 2024-01-15 um 14 45 55

After:
Bildschirmfoto 2024-01-15 um 15 22 47
Bildschirmfoto 2024-01-15 um 15 23 00

**Failed installation check due to error:**
Before:
Bildschirmfoto 2024-01-15 um 15 27 54
Bildschirmfoto 2024-01-15 um 15 28 14

After:
Bildschirmfoto 2024-01-15 um 15 23 51
Bildschirmfoto 2024-01-15 um 15 26 01

### How to use/reproduce
Adjust the gradle example to use a lower version than minimum requirement (e.g. 1.11.0) or make the build fail to cause an error while running the plugin (e.g. remove plugin managament block from settings.gradle).


florianGla (approved):
Looks good to me

florianGla:
It looks like the spacing between the `X` and the first letter for errors messages got smaller. But maybe it has something to do with you terminal settings.

PR #1280, opened 2024-01-15
feat: set local or remote mode

MarkusZoppelt:
### Motivation/Context

We want to have an explicit way to tell if CI Fuzz is used in a local mode only for the project.

For each project we want to determine in which mode CI Fuzz is operated while giving the user both options: local or remote.

### Description

Determine the mode (local/remote) is done during `init`.

We add a hint on how the user can change this later.

There is no default server anymore.

The way to determine if we are in a local or remote setting is to check the `server` config value.

If no config value is given for `server`, the project is operated in a **local-only** mode.

If a config value is given for `server` the project is operated in a **remote** mode.

### How to use/reproduce

Interactive mode:

    cifuzz init -C examples/gradle

Non-interactive mode (local-only mode):

    cifuzz init -C examples/gradle --interactive=false

Non-interactive mode (remote mode):

    cifuzz init -C examples/gradle --server app.staging.code-intelligence.com --project <prj-nid>

#### e2e tests:

    make test/e2e


MarkusZoppelt:
> I tested out the command and I think the first line to just be "Enter a CI Sense server URL" very confusing. Why do I want that? What will it do? I think a bit of explanation would be good, like, "we determined you are running in remote mode" and maybe even explain what that means. Not sure if we have to assume that the first time user knows pretty much nothing but some explanation could help.
> 
> Tbh I don't even know why I am flagged as remote, I can't remember that I set a flag or something for it 😅 I would have liked to have the choice anyway but I'm not sure if this is suitable for the general use case. Is it because I am logged in?

That was a bug :) It's now fixed

e2e/tests/finding/finding_test.go
```diff
@@ -88,6 +88,31 @@ var findingTests = &[]e2e.TestCase{
 	},
 }
 
+var findingModeTests = &[]e2e.TestCase{
+	{
+		Description:  "findings in local mode should not invoke CI Sense communication",
```
bertschneider, line 93: change to "finding command"

e2e/tests/finding/finding_test.go
```diff
@@ -88,6 +88,31 @@ var findingTests = &[]e2e.TestCase{
...
+		Command:      "finding",
+		Args:         []string{"-v"},
+		SampleFolder: []string{"project-with-findings-and-error-details"},
+		Assert: func(t *testing.T, output e2e.CommandOutput) {
+			assert.NotContains(t, output.Stderr, "Sending HTTP request:")
+			assert.EqualValues(t, 0, output.ExitCode)
+		},
+	},
+	{
+		Description:  "findings in remote mode should invoke CI Sense communication",
```
bertschneider, line 104: change to "finding command"

e2e/tests/init/init_test.go
```diff
@@ -31,6 +32,56 @@ var initTests = &[]e2e.TestCase{
...
+		// note: we are always --interactive=false in CI
+		Args:         []string{"maven --interactive=false"},
+		SampleFolder: []string{"empty"},
+		Assert: func(t *testing.T, output e2e.CommandOutput) {
+			assert.EqualValues(t, 0, output.ExitCode)
+			assert.NotContains(t, output.Stdall, "Do you want to initialize cifuzz for this project in remote (recommended) or local-only mode?")
+			assert.Contains(t, output.Stdall, "Running in local-only mode.")
+			assert.Contains(t, output.Stdall, "Configuration saved in cifuzz.yaml")
+			assert.NotContains(t, output.Stdall, "Failed to create config")
+			output.FileExists("cifuzz.yaml")
```
bertschneider, line 48: Line can be omitted.

e2e/tests/run/run_test.go
```diff
@@ -38,6 +40,35 @@ var runWithInvalidTokenTests = &[]e2e.TestCase{
 	},
 }
 
+var runModeTests = &[]e2e.TestCase{
```
bertschneider, line 43: Description of the commands above is wrong, also they should only use projects from the sample folder.

internal/cmd/finding/finding.go
```diff
@@ -155,14 +159,17 @@ Skipping remote findings because running in non-interactive mode.`)
...
 	// them individually later. These won't be stored on disk.
 	var remoteFindings []*finding.Finding
-	if remoteAPIFindings != nil {
-		for i := range remoteAPIFindings.Findings {
-			// we access the element via index to avoid copying the struct
-			rf, err := api.RemoteToLocalFinding(&remoteAPIFindings.Findings[i])
-			if err != nil {
-				return err
+
+	if isRemoteMode {
```
bertschneider, line 163: Move this up into the block that loads the remote findings?

internal/cmd/finding/finding.go
```diff
@@ -230,10 +237,12 @@ Skipping remote findings because running in non-interactive mode.`)
 	// prints the information available for the specified finding
 	findingName := args[0]
 
-	// check if the finding is a remote finding...
-	for _, remoteFinding := range remoteFindings {
-		if strings.TrimPrefix(remoteFinding.Name, fmt.Sprintf("projects/%s/findings/", cmd.opts.Project)) == findingName {
-			return cmd.printFinding(remoteFinding)
+	if isRemoteMode {
```
bertschneider, line 240: `remoteFindings` should be empty if not in remote mode, hence the outer `if` could be removed, right?

internal/cmd/finding/finding.go
bertschneider, line 163: Nowhere else to add this, so I do it here :smile: The `localFindings` variable is only used in case `len(args) == 0`, so the whole finding lookup code could be moved in that `if` block. Perhaps also extracting the two `if`-`else`-blocks into dedicated functions.

internal/cmd/init/init.go
```diff
@@ -21,6 +22,7 @@ import (
 )
 
 type options struct {
+	apiClient   *api.APIClient
```
bertschneider, line 25: `apiClient` is only used in the interactive mode handling. Just create it there?

internal/cmd/init/init.go
```diff
@@ -114,22 +118,76 @@ func New() *cobra.Command {
...
+				return err
+			}
+
+			if opts.Project == "<<cancel>>" {
+				log.Info("Canceled")
+				return cmdutils.ErrSilent
+			}
+		}
+	} else { // non-interactive mode
+		if !viper.IsSet("server") {
```
bertschneider, line 161: Add a comment why `opts.Server` is set to `""`?

internal/cmd/init/init.go
```diff
@@ -114,22 +118,76 @@ func New() *cobra.Command {
...
+			opts.Server = ""
+		}
+
+		// whoever reads this code: I'm sorry
+		if (viper.IsSet("server") && opts.Project == "") || (!viper.IsSet("server") && opts.Project != "") {
```
bertschneider, line 166: Extract the check out into a reasonably named variable, similar to 'isLocalMode' below?

bertschneider (approved):
Nice. Only some nit-picks :smile:

internal/cmd/init/init.go
MarkusZoppelt: No, it's also used in non-interactive if both `--server` and `--project` flags are set. I think it's better to keep it in the opts here.

PR #1286, opened 2024-01-16
chore: replace links to github docs with docs.code-intelligence.com

mlsandnerCI:
### Motivation/Context

Another step on the road to making CI Fuzz private.

### Description

Replaced links to our public Github docs with links to docs.code-intelligence.com, except in the case of "Sandboxing" because we don't have an equivalent for that and it seems like we are removing that feature soon anyway.


bertschneider (approved):
LGTM. Thx!

PR #1288, opened 2024-01-16
**MEGA PR** feat: remove minijail and all related code

MarkusZoppelt:
### Motivation/Context

Remove minijail related code - it is no longer required by CI Sense.

### Description

Remove minijail related code

### How to use/reproduce

Let's see what the tests say...


MarkusZoppelt:
Looks like only the compatibility check in the installer pipeline is complaining (because the `--use-sandbox` flag is now gone).

Rest seems fine.

PR #1291, opened 2024-01-17
feat: use gradle and maven tasks for fuzz test completion

mlsandnerCI:
### Motivation/Context

In custom setups where fuzz tests are not in the directory "test", we couldn't find them for auto completion because we used a static approach to only check the standard directory layout.

### Description

Fuzz test completion for Maven and Gradle are now using the designated task `cifuzzPrintTestSourceFolders` in the extension/plugin to get all directories with tests to search in. 
### How to use/reproduce

Try out auto completion in the `test/projects/gradle/testsuite` directory.


florianGla:
I'm not sure if we want to run external maven/gradle tasks for autocompletion. In the worst case the whole project gets build by executing those tasks. I think we should discuss this with our team first.

PR #1293, opened 2024-01-18
fix(finding): omit empty api.Finding fields

MarkusZoppelt:
### Motivation/Context

finding upload via /v1 API was broken, because new /v3 api.Finding fields were not omitted when empty.

### Description

omit empty api.Finding fields.

### How to use/reproduce

tests passing


oetr (approved):
Thanks for this PR!

